PR #1, opened 2015-05-21
Draft dreamcode in README

leplatrem:
First version imported from collaborative pad.

* Everything could be rewritten without pity;
* Synchronization strategy is definitely naive so far (our use cases are too);

Please review:
* [ ] Shape and wording
* [ ] Weird ideas like URL placeholder etc.
* [ ] Sync approach and strategy;
* [ ] Markdown formatting for API docs, snippets and notes.

Example of code that implemented another strategy to sync records with *Cliquet*:
* https://github.com/mozilla-services/readinglist/wiki/Client-phases
* https://hg.mozilla.org/releases/mozilla-aurora/file/18286de4be87/browser/components/readinglist/Sync.jsm

Synchronization using ETags:
* https://unhosted.org/practice/28/Synchronizing-browser-storage-with-server-storage.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,214 @@ Build
 
     $ npm run dist
 
+Usage
+-----
+
+**Caution: everything in this section is still pure fiction.**
+
+* Every operation is performed locally;
+* Synchronization with server shall be ran explicitly.
+
+### Collection
+
+```js
+
+  import Cliquetis from "cliquet";
+
+  var db = new Cliquetis();
+  var articles = db.collection('articles');
+
+```
+
+The collection object has the following attributes:
+
+* **last_modified**: last synchronization timestamp, ``undefined`` if never sync'ed.
+
+> Synchronization timestamps are persisted in the device local storage.
+
+
+### List records
+
+```js
+
+  articles.all().then(result => {
+    {
+      data: [
+        {
+          id: "705b17be-e957-4c14-8f4c-86f8eaac29c0",
+          title: "foo"
+        },
+        {
+          id: "68e63131-3859-40cc-a4f7-b237ca179329",
+          last_modified: 1432222889336,
+          title: "Web page"
+        },
+      ]
+    }
+  });
+
+```
+
+> Records with ``last_modified`` attribute were sync'ed on a server.
+
+
+### Create records
+
+```js
+
+  articles.save({title: "foo"}).then(result => {
+    data: {
+      id: "2dcd0e65-468c-4655-8015-30c8b3a1c8f8",
+      title: "foo",
+    }
+  });
+
+```
+
+> Records identifiers are generated locally using UUID4.
+
+
+### Update records
+
+```js
+
+  var existing = {
+    id: "2dcd0e65-468c-4655-8015-30c8b3a1c8f8",
+    title: "bar"
+  };
+
+  articles.save(existing).then(result => {
+    data: {
+      id: "2dcd0e65-468c-4655-8015-30c8b3a1c8f8",
+      title: "bar",
+    }
+  });
+
+```
+
+### Delete records
+
+```js
+
+articles.delete({
+    selector: {
+      age: {'$min': 42}
+    }
+  }).
+  then(result => {
+    data: [
+      {
+        id: "2dcd0e65-468c-4655-8015-30c8b3a1c8f8",
+        deleted: true,
+      }
+    ]
+  });
+
+```
+
+A shortcut for ``articles.delete({id: {'$eq': id}})`` is available at ``articles.delete(id)``.
+
+
+
+### Filter records
+
+```js
+
+articles.all({
+    selector: {
+      title: {'$eq': 'foo'}
+    }
+  }).
+  then(result => {
+    data: [
+      …
+    ]
+  });
+
+```
+
+A shortcut for ``articles.all({id: {'$eq': id}})`` is available at ``articles.get(id)``.
+
+
+**Sorting**
+
+```js
+
+articles.all({
+    sort: [{added_on: 'desc'}]
+  }).
+  then(result => {
+    data: [
+      …
+    ]
+  });
+
+```
+
+
+### Fetch and publish changes
+
+```js
+
+var options = {
+  remote: "https://server/v1/{}/records",
+  headers: {
+    "Authorization": "Basic bWF0Og=="
+  },
+  mode: Cliquet.SAFE,
+};
+
+articles.sync(options).then(result => {
+    {
+      ok: true,
+      created: [],  // missing locally.
+      updated: [],  // changed since last sync.
+      deleted: [],  // deleted since last sync.
+      conflicts: []  // changed both sides.
+    }
+});
+
+```
+
+> If ``ok`` is ``false``, the list of conflicting records will be provided in
+> ``conflicts`` attribute.
+>
+> Conflicts must be resolved locally and ``sync()`` called again.
+
+
+**Options**
+
+* **remote**: Server URI with placeholder for collection name
+* **headers**: HTTP headers (*authentication*)
+* **mode**: ``Cliquet.SAFE`` (*server wins*), ``Cliquet.FORCE`` (*client wins*)
+
+**Synchronization strategy**
+
+# Fetch changes since last synchronization using ``?_since``;
+# Detect conflicts and apply changes if not any;
+# Publish deletions of records;
+# Publish creations records.
+
+**Notes**
+
+> During synchronization, records created locally are published on the server
+> using ``PUT`` and the ``If-None-Match: *`` request header to prevent overwriting.
+
+> Since fetching changes is paginated, it should be performed using ``If-None-Match``
+> header to prevent race-conditions.
+> If a ``412 Precondition failed`` is received, synchronization is stopped,
+> and nothing is performed.
+> Another request with ``?_since`` will have to sent in order to fetch information about
+> changes that occured since pagination was initiated.
+> ­­→ Not acceptable: whole sync is cancelled if another device creates a record
+> meanwhile. Find something better (Think of a blamk device that should sync the
+> whole collection).
+
+> ``404 Not Found`` errors are ignored on ``DELETE``.
+
+> After a synchronization, the collection timestamp is updated.
+
+
 Tests
 -----
 
--- a/README.md
+++ b/README.md
@@ -1,224 +1,258 @@
-Cliquetis
-=========
+# Cliquetis
 
 [![Build Status](https://travis-ci.org/mozilla-services/cliquetis.svg?branch=master)](https://travis-ci.org/mozilla-services/cliquetis)
 
 A JavaScript client for [Cliquet](https://github.com/mozilla-services/cliquet/).
 
-Installation
-------------
+This is work in progress, there's no code just yet.
+
+## Installation
 
     $ npm install
 
-Build
------
+## Build
 
     $ npm run dist
 
-Usage
------
+## Usage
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,6 +15,214 @@ Build
...
+          id: "705b17be-e957-4c14-8f4c-86f8eaac29c0",
+          title: "foo"
+        },
+        {
+          id: "68e63131-3859-40cc-a4f7-b237ca179329",
+          last_modified: 1432222889336,
+          title: "Web page"
+        },
+      ]
+    }
```
n1k0: Maybe we should invent another way to describe that resulting data structure; here the code is invalid, doesn't make much sense and could confuse newcomers.

Also, we should probably attach a `last_modified` property to that result object, as described above.

README.md
```diff
@@ -15,6 +15,214 @@ Build
...
+### Create records
+
+```js
+
+  articles.save({title: "foo"}).then(result => {
+    data: {
+      id: "2dcd0e65-468c-4655-8015-30c8b3a1c8f8",
+      title: "foo",
+    }
+  });
```
n1k0: Note that if we want to keep this formalism, we should at least make the code valid, wrapping the returning object into parenthesis:

```
articles.save({title: "foo"}).then(result => ({
  data: {
    id: "2dcd0e65-468c-4655-8015-30c8b3a1c8f8",
    title: "foo",
  }
}));
```

README.md
```diff
@@ -15,6 +15,214 @@ Build
...
+### Update records
+
+```js
+
+  var existing = {
+    id: "2dcd0e65-468c-4655-8015-30c8b3a1c8f8",
+    title: "bar"
+  };
+
+  articles.save(existing).then(result => {
```
n1k0: A minor I see with using a create/update agnostic `save` method is that it's not explicit anymore. What's the reasoning behind this move?

README.md
```diff
@@ -15,6 +15,214 @@ Build
...
+
+### Delete records
+
+```js
+
+articles.delete({
+    selector: {
+      age: {'$min': 42}
+    }
+  }).
```
n1k0: Maybe we should add an example for deleting a single record by its id?

README.md
```diff
@@ -15,6 +15,214 @@ Build
...
+      {
+        id: "2dcd0e65-468c-4655-8015-30c8b3a1c8f8",
+        deleted: true,
+      }
+    ]
+  });
+
+```
+
+A shortcut for ``articles.delete({id: {'$eq': id}})`` is available at ``articles.delete(id)``.
```
n1k0: Haha scratch my comment above.

README.md
```diff
@@ -15,6 +15,214 @@ Build
...
+
+
+
+### Filter records
+
+```js
+
+articles.all({
+    selector: {
+      title: {'$eq': 'foo'}
```
n1k0: Nit: property names doesn't have to be quoted

README.md
```diff
@@ -15,6 +15,214 @@ Build
...
+  }).
+  then(result => {
+    data: [
+      …
+    ]
+  });
+
+```
+
+A shortcut for ``articles.all({id: {'$eq': id}})`` is available at ``articles.get(id)``.
```
n1k0: Now I think about, we should put this before the more complicated query example, as the flow would be more natural to me.

README.md
leplatrem: OH I'm surprised that it works, even with $ :)

README.md
```diff
@@ -15,6 +15,214 @@ Build
...
+
+```js
+
+var options = {
+  remote: "https://server/v1/{}/records",
+  headers: {
+    "Authorization": "Basic bWF0Og=="
+  },
+  mode: Cliquet.SAFE,
+};
```
n1k0: How about separating generic options (`remote`, `headers`) from occasional/specific operations (`mode`)?

We could pass generic options to the `Cliquet` constructor, and keep specific options as a `sync()` argument.

README.md
```diff
@@ -15,6 +15,214 @@ Build
...
+
+articles.sync(options).then(result => {
+    {
+      ok: true,
```
n1k0: In case of `ok: false` I suspect this would be caught in a `.catch` clause, so I suggest we remove this prop here.

README.md
```diff
@@ -15,6 +15,214 @@ Build
...
+      updated: [],  // changed since last sync.
+      deleted: [],  // deleted since last sync.
+      conflicts: []  // changed both sides.
+    }
+});
+
+```
+
+> If ``ok`` is ``false``, the list of conflicting records will be provided in
+> ``conflicts`` attribute.
```
n1k0: Bah, I should just quit commenting in real-time; scratch my comment above.

n1k0: Note that developers could simply check that the conflicts array length is > 0…

README.md
```diff
@@ -15,6 +15,214 @@ Build
...
+>
+> Conflicts must be resolved locally and ``sync()`` called again.
+
+
+**Options**
+
+* **remote**: Server URI with placeholder for collection name
+* **headers**: HTTP headers (*authentication*)
+* **mode**: ``Cliquet.SAFE`` (*server wins*), ``Cliquet.FORCE`` (*client wins*)
```
n1k0: /me likes this a lot

README.md
leplatrem: I was thinking that in most cases we don't care, we just want to save the record.
Otherwise the ``if`` will have to be done by users ?
```
if (record.id && record.last_modified) {
    record = collection.save(record);
}
else {
    record = collection.create(record);
}
```
I would prefer to keep this logics hidden :) But I'm fully open to discussion!

n1k0:
That's a pretty good start. I'll take over the patch to address some of the comments I made, then let's sync up (haha) on this later this morning :)

README.md
leplatrem: +1 for using another way to describe promises results :)

Regarding ``last_modified``, it is a collection attribute that is updated from server during ``sync()``. Should it really be here when querying locally with ``all()``?

README.md
leplatrem: +1 good idea!

leplatrem:
Thanks for your feedback! I let you take over, don't hesitate to tear down some parts!

README.md
```diff
@@ -15,6 +15,214 @@ Build
 
     $ npm run dist
 
+Usage
```
Natim: This should probably be ``## Usage``

Natim: same above for Build and Installation, also Cliquetis should be ``# Cliquetis``

README.md
```diff
@@ -15,6 +15,214 @@ Build
...
+
+**Options**
+
+* **remote**: Server URI with placeholder for collection name
+* **headers**: HTTP headers (*authentication*)
+* **mode**: ``Cliquet.SAFE`` (*server wins*), ``Cliquet.FORCE`` (*client wins*)
+
+**Synchronization strategy**
+
+# Fetch changes since last synchronization using ``?_since``;
```
Natim: This is not the right syntax for numeric list.
See http://stackoverflow.com/editing-help#advanced-lists

README.md
```diff
@@ -15,6 +15,214 @@ Build
...
+# Publish creations records.
+
+**Notes**
+
+> During synchronization, records created locally are published on the server
+> using ``PUT`` and the ``If-None-Match: *`` request header to prevent overwriting.
+
+> Since fetching changes is paginated, it should be performed using ``If-None-Match``
+> header to prevent race-conditions.
+> If a ``412 Precondition failed`` is received, synchronization is stopped,
```
Natim: We should be specific that this is not the case with batch.

README.md
```diff
@@ -15,6 +15,214 @@ Build
...
+> and nothing is performed.
+> Another request with ``?_since`` will have to sent in order to fetch information about
```
Natim: to be sent

Natim:
Quite nice :+1:

README.md
```diff
@@ -1,20 +1,262 @@
...
+### The `Cliquetis` constructor
+
+```js
+var db = new Cliquetis(options);
+```
+
+`options` is an object defining the following option values:
+
+- `remote`: The remote Cliquet server endpoint root URL (eg. `"https://server/v1"`);
+- `headers`: The default headers to pass for every HTTP request performed to the Cliquet server (eg. `{"Authorization": "Basic bWF0Og=="}`);
```
ametaireau: I'm not sure this example is right, since I think we will have a way to do authorization baked in cliquetis at some point?

README.md
```diff
@@ -1,20 +1,262 @@
...
+- `mode`: The conflict default resolution strategy (`Cliquet.SAFE` (*server wins*), `Cliquet.FORCE` (*client wins*)).
```
ametaireau: Should we rename in `SERVER_WINS` and `CLIENT_WINS`.

README.md
```diff
@@ -1,20 +1,262 @@
...
+// result
+{
+  data: {
+    id: "2dcd0e65-468c-4655-8015-30c8b3a1c8f8",
+    title: "foo",
+  }
+}
+```
+
+> Records identifiers are generated locally using UUID4.
```
ametaireau: Why is this a quote?

README.md
```diff
@@ -1,20 +1,262 @@
...
+
+```js
+var articles = db.collection("articles");
+```
+
+This collection object has the following attributes:
+
+* **lastModified**: last synchronization timestamp, ``null`` if never sync'ed.
+
+> Synchronization timestamps are persisted in the device local storage.
```
ametaireau: Why is this a quote?

README.md
```diff
@@ -1,20 +1,262 @@
...
+  }
+}
+```
+
+### Deleting records
+
+#### Single unique record passing its `id`:
+
+```js
+articles.delete("2dcd0e65-468c-4655-8015-30c8b3a1c8f8")
```
ametaireau: Would it be possible to delete an record by passing the object directly (as well as passing the id)?

README.md
n1k0: So that would be exposing a global, I'd rather attach these to the main exported symbol.

README.md
n1k0: In md we don't have macros, so that's a kind of a "note". Is this a pb?

README.md
n1k0: That's definitely an option, though if you have an object to pass, you got its id. So I'd vote for keeping the api offering a single way to do things whenever it's possible (eg. here).

README.md
ametaireau: I mean `Cliquet.SERVER_WINS` and `Cliquet.CLIENT_WINS`

README.md
ametaireau: Let's convert this doc in restructured text? :-)

README.md
ametaireau: It seems counter intuitive: for all other operations, we're passing an object and for this one we're passing an id

README.md
leplatrem: I find this pretty flexible though

README.md
ametaireau: I would not use a quote for this then, because that's not what it is. If it's a note, maybe we could prefix with `note: ` ?

README.md
leplatrem: Using ``Cliquetis.SAFE`` records that changed both locally and remotely will be marked as conflicts. Thus, it is not truely *server wins*.

But I know what you mean:  safe and force are not explicit enough. However is the *wins* word adequate? Overwrite?

ametaireau: At least the description should be changed to talk about conflicts, then.

PR #3, opened 2015-05-26
First implementation of local CRUD operations.

n1k0:
r=? @leplatrem (code style, quality of implementation, docs review, any missing bits?)

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -45,7 +45,8 @@ export default class Collection {
       };
       transaction.onerror = function(event) {
         // XXX reject with proper error
-        let err = new Error();
+        let err = new Error('Transaction error: ' + event.target.error);
+        console.log(event)
         reject(err);
       };
 
@@ -54,11 +55,14 @@ export default class Collection {
         return reject(err);
       }
 
-      // XXX handle update case
+      let store = transaction.objectStore(this._collName);
       if (!record.id) {
         record = Object.assign({}, record, {id: uuid4()});
+        store.add(record);
       }
-      transaction.objectStore(this._collName).add(record);
+      else {
+        store.put(record);
+      }
     });
   }
 }
--- a/src/collection.js
+++ b/src/collection.js
@@ -11,20 +11,20 @@ attachFakeIDBSymbolsTo(typeof global === "object" ? global : window);
 // may still fail after the success event fires.
 
 export default class Collection {
-  constructor(collName) {
-    this._collName = collName;
+  constructor(name) {
+    this._name = name;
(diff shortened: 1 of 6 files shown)


n1k0:
This is now ready for review @leplatrem

src/collection.js
```diff
@@ -0,0 +1,138 @@
...
+        this._db = event.target.result;
+        resolve(this);
+      };
+    });
+  }
+
+  transaction(mode) {
+    return this._db
+      .transaction([this._collName], mode)
+      .objectStore(this._collName);
```
leplatrem: This method is called transaction but actually starts a new transaction and returns an object store.

src/collection.js
```diff
@@ -0,0 +1,138 @@
...
+
+  get name() {
+    return this._collName;
+  }
+
+  init() {
+    return new Promise((resolve, reject) => {
+      var request = indexedDB.open(this._collName, 1);
+      request.onupgradeneeded = event => {
+        var store = event.target.result.createObjectStore(this._collName, {
```
leplatrem: we could use ``this.name`` everywhere no ?

src/collection.js
```diff
@@ -0,0 +1,138 @@
...
+      }
+      var request;
+      if (!record.id) {
+        request = transaction.add(Object.assign({}, record, {id: uuid4()}));
+      } else {
+        request = transaction.put(record);
+      }
+      request.onsuccess = function(event) {
+        resolve({
+          data: Object.assign({}, record, {id: event.target.result}),
```
leplatrem: I don't understand this :( What are we providing in data exactly ?

src/index.js
```diff
@@ -1,15 +1,19 @@
...
+    this._collections = {};
   }
 
-  collection(name) {
-    return new Collection(name);
+  collection(collName) {
+    return new Promise((resolve, reject) => {
+      // if collection collName is missing, reject
+      if (!collName)
+        return reject(new Error("missing collection name"));
```
leplatrem: I wonder if this should faild in a promise, and if it should fail during call...

src/utils.js
```diff
@@ -0,0 +1,18 @@
+export function attachFakeIDBSymbolsTo(obj) {
```
leplatrem:
```
/**
 In FakeIndexedDB, symbols are exposed using ``FDB`` prefixes in names.
 This piece of code will register them with the same names as native API,
 only if indexedDB is not already available.
 */
```

leplatrem:
r+ \o/

PR #4, opened 2015-05-27
Simplified Collection API.

n1k0:
Our current `Cliquetis#collection` returns a Promise, which makes users writing a bunch of boilerplate for no obvious reason:

```js
var db = new Cliquetis();
db.collection("articles").then(function(articles) {
  return articles.save({title: "foo"});
}).then(…);
```

This patch updates this method so it directly returns a `Collection` instance, which will ensure for every of its method calls that the db is opened, resulting in a much more lighweight usage:

```js
var db = new Cliquetis();
db.collection("articles").save({title: "foo"}).then(…);
```

r=? @leplatrem

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 A JavaScript client for [Cliquet](https://github.com/mozilla-services/cliquet/).
 
-This is work in progress, there's no code just yet.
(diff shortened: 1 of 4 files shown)


n1k0:
Woops, lemme update docs accordingly.

leplatrem:
r+

It is simpler indeed. I like the «lazy» approach very much :)

PR #7, opened 2015-06-01
[Ready for Review] First sync() implementation.

n1k0:
Early feedback is welcome.

TODO:
- [x] Avoid rejecting with conflicting sync results, rather resolve with a boolean flag set, eg. `ok: false`.
- [x] Update docs

**Edit:** This is now in a state I can ask for review. Please concentrate on the sync flow and the public API rather than nitpicking/CS for now (thanks!).

r=? @leplatrem @ametaireau @Natim


test/index_test.js
```diff
@@ -1,96 +1,128 @@
...
-    it("should reject on missing collection name", function() {
-      expect(function() {
-        new Cliquetis().collection();
-      }).to.Throw(Error, /missing collection name/);
+    it("should cache collection instance", () => {
+      var db = new Cliquetis();
+      expect(db.collection("a") == db.collection("a")).eql(true);
+    });
+
+    it("should reject on missing collection name", () => {
```
leplatrem: should throw :)

leplatrem:
Some insights for **polling for changes**:

Notes:
* When using ``?_since=``, deleted items are returned in collection with attribute ``delete: true``
* Collection is paginated by default (fetch until ``Next-Page`` header is present)

Simple paginated fetch:

1. Cold sync → no last_modified
1. GET /collection/?_since=<last_modified> (with header ``If-Unmodified-Since = <last_modified>``)
1. If ``304``, done → up to date, nothing to do.
1. If no next page header, done → store ``Last-Modified`` header value for next calls and exploit the obtained records.
1. If ``Next-Page``, go on to it (it's an url)

Simple concurrency control (no pagination):

1. Cold sync → no last_modified
1. GET /collection/?_since=<last_modified> (with header ``If-Unmodified-Since = <last_modified>`` **and** with header ``If-Not-Modified-Since = <last_modified>``)
1. If ``304``, done → up to date, nothing to do.
1. If ``200``, done → store ``Last-Modified`` header value for next calls and exploit the obtained records.
1. If ``412``
    1. store ``Last-Modified`` header value into ``until`` variable
    1. obtain changes in the interim with GET ``/collection?_since<last_modified>&_to=<until>``
    1. Store ``Last-Modified`` into ``last_modified``
    1. Goto step 2.

* Idea*: count retries, and add a longer delay at each retry.


Paginated fetch with concurrency control:

* Combine the two above :)

demo/index.html
```diff
@@ -0,0 +1,157 @@
...
+              ', _status=' + task._status + '</small>',
+            '<input type="button" value="Delete">'
+          ].join("");
+          li.querySelector("input[type=checkbox]")
+            .addEventListener("click", toggleDone(task));
+          li.querySelector("input[type=button]")
+            .addEventListener("click", deleteTask(task));
+          ul.appendChild(li);
+        });
+      });
```
bnjbvr: CAN HAZ a catch() method in case of issues with the promise?

demo/index.html
```diff
@@ -0,0 +1,157 @@
...
+          li.innerHTML = [
+            '<label>',
+            '  <input type="checkbox" ' + (task.done ? 'checked' : '') + '> ',
+            '  ' + (task.done ? '<del>' + task.label + '</del>' : task.label),
+            '</label>',
+            '<br><small>id: ' + task.id +
+              ', last_modified: ' + task.last_modified +
+              ', _status=' + task._status + '</small>',
+            '<input type="button" value="Delete">'
+          ].join("");
```
bnjbvr: template strings FTW

n1k0: Yeah, though this is just a plain page demo, I don't really want to throw babel transpilation for this. The lib code is using template strings though.

demo/index.html
n1k0: GOOD CATCH, HAHAHAHAHAHA

/me needs some rest, though not HTTP one HAHAHAHAHA (sry)

src/api.js
```diff
@@ -0,0 +1,110 @@
+"use strict";
+
+const RECORD_FIELDS_TO_CLEAN = ["_status", "last_modified"];
+
+export function cleanRecord(record, exludeFields=RECORD_FIELDS_TO_CLEAN) {
```
bnjbvr: nit: exlude => exclude

src/api.js
```diff
@@ -0,0 +1,110 @@
...
+  return Object.keys(record).reduce((acc, key) => {
+    if (exludeFields.indexOf(key) === -1)
```
bnjbvr: ditto

src/api.js
```diff
@@ -0,0 +1,110 @@
...
+  "Accept":       "application/json",
+  "Content-Type": "application/json",
+};
+
+export default class Api {
+  constructor(remote, options={}) {
+    this.remote = remote;
+    this._options = options;
+    try {
+      this.version = "v" + remote.match(/\/v(\d+)$/)[1];
```
bnjbvr: How about allowing an optional slash at the end of the URL?

src/api.js
```diff
@@ -0,0 +1,110 @@
...
+    } catch (err) {
+      throw new Error("The remote URL must contain the version: " + remote);
```
bnjbvr: Might be worth to check somewhere that remote is a string, otherwise this error message could be misleading

src/api.js
```diff
@@ -0,0 +1,110 @@
...
+    };
+    var reject = false;
+    if (!records.length)
+      return Promise.resolve(results);
+    return fetch(this.endpoints().batch(), {
+      method: "POST",
+      headers: DEFAULT_REQUEST_HEADERS,
+      body: JSON.stringify({
+        defaults: {
+          headers: headers,
```
bnjbvr: You can use the shorthand here:
`defaults: {
  headers
}`

src/api.js
```diff
@@ -0,0 +1,110 @@
...
+        } else {
+          results.errors.push({
+            path: response.path, // this is the only way to have the id…
+            error: response.body
+          });
+          reject = true;
+        }
+      });
+      if (reject)
+        throw results;
```
bnjbvr: throwing in a then()able needs a catch() method as well, or the error will get silently ignored iirc

n1k0: This actually works as expected, though that's something I want to change; feels odd rejecting with smthg else than an error object.

demo/index.html
n1k0: Holy crap, they're available natively in FFx and Chrome. Will update.

README.md
```diff
@@ -225,46 +253,90 @@ Result:
...
+  published: [], // Successfully published
+  conflicts: []  // Export conflicts
 }
 ```
 
-> If conflicts occured, they're listed in the `conflicts` array property; they must be resolved locally and `sync()` called again.
+If conflicts occured, they're listed in the `conflicts` property; they must be resolved locally and `sync()` called again.
+
+The `conflicts` array is in this form:
+
```
leplatrem: Difference with errors ?

n1k0: Errors = uncaught exceptions (eg. indexedDB errors), conflicts = conflicts.

src/api.js
```diff
@@ -0,0 +1,139 @@
...
+    if (excludeFields.indexOf(key) === -1)
+      acc[key] = record[key];
+    return acc;
+  }, {});
+};
+
+const DEFAULT_REQUEST_HEADERS = {
+  "Accept":       "application/json",
+  "Content-Type": "application/json",
+};
```
leplatrem: Note for later:
``# XXX: This could probably be an attribute of the Api class, so that developers can get a hand on it to add their own headers.``

src/api.js
```diff
@@ -0,0 +1,139 @@
...
+          const body = isDeletion ? undefined : cleanRecord(record);
+          const headers = options.safe && record.last_modified ?
+                          {"If-Unmodified-Since": String(record.last_modified)} : {};
+          return {method, headers, path, body};
+        })
+      })
+    })
+      .then(res => res.json())
+      .then(res => {
+        if (res.error)
```
leplatrem: Is ``res.error`` set by fetch() ?

src/api.js
```diff
@@ -0,0 +1,139 @@
...
+          } else if (response.status === 404) {
+            results.skipped.push(response.body);
+          } else if (response.status === 412) {
+            results.conflicts.push({
+              type: "outgoing",
+              data: response.body
+            });
+          } else {
+            results.errors.push({
+              path: response.path, // this is the only way to have the id…
```
leplatrem: Note for later:
``// XXX: since responses come in the same order, there should be a way to get original record id``

src/api.js
```diff
@@ -0,0 +1,139 @@
...
+   */
+  batch(collName, records, headers={}, options={safe: true}) {
+    const results = {
+      errors:    [],
+      published: [],
+      conflicts: [],
+      skipped:   []
+    };
+    if (!records.length)
+      return Promise.resolve(results);
```
leplatrem: Note for later: if more than X results (default is 25 on server), split in several calls.

src/collection.js
```diff
@@ -44,143 +109,422 @@ export default class Collection {
...
-    const store = transaction.objectStore(this.name);
+  prepare(mode=undefined, name=null) {
+    const storeName = name || this.name;
+    const transaction = this._db.transaction([storeName], mode);
+    const store = transaction.objectStore(storeName);
     return {transaction, store};
   }
 
+  /**
+   * Clears current collection.
```
leplatrem: Clears → Deletes every records in the current collection.

src/collection.js
```diff
@@ -44,143 +109,422 @@ export default class Collection {
...
 
-  _update(record) {
+  /**
+   * Updates a record from the local database.
+   *
+   * @param  {Object} record
+   * @param  {Object} options
+   * @return {Promise}
+   */
+  update(record, options={synced: false}) {
```
leplatrem: Nit: the option name ``synced`` can be confusing (i.e. operation performed asynchronously). I would propose something like ``fresh``, ``raw``, ``unprocessed``...

src/collection.js
```diff
@@ -44,143 +109,422 @@ export default class Collection {
...
 
-  list() {
+  /**
+   * Lists records from the local database.
+   *
+   * @param  {Object} params
+   * @param  {Object} options
+   * @return {Promise}
+   */
+  list(params={}, options={includeDeleted: false}) {
```
leplatrem: Looks like params is unused ?

src/collection.js
```diff
@@ -44,143 +109,422 @@ export default class Collection {
...
+    }).catch(this._handleError("list"));
+  }
+
+  /**
+   * Import a single change into the local database.
+   *
+   * @param  {Object} change
+   * @return {Promise}
+   */
+  importChange(change) {
```
leplatrem: This could be private API ? since it is only used by ``ìmportChanges()``
``_importChange()`` ?

src/collection.js
```diff
@@ -44,143 +109,422 @@ export default class Collection {
...
+            // Remote update was successful, refect it locally
+            return this.update(record, {synced: true});
+          }
+        })).then(published => {
+          syncResultObject.add("published", published.map(res => res.data))
+          return syncResultObject;
+        });
+      });
+  }
+
```
leplatrem: <3

test/api_test.js
```diff
@@ -0,0 +1,386 @@
...
+      it("should provide root endpoint", () => {
+        expect(endpoints.root()).eql(FAKE_SERVER_URL);
+      });
+
+      it("should provide batch endpoint", () => {
+        expect(endpoints.batch())
+          .eql("http://fake-server/v0/batch");
+      });
+
+      it("should provide root endpoint", () => {
```
leplatrem: records endpoint

test/api_test.js
```diff
@@ -0,0 +1,386 @@
...
+        expect(endpoints.collection("toto"))
+          .eql("http://fake-server/v0/collections/toto/records");
+      });
+
+      it("should provide root endpoint", () => {
```
leplatrem: record enpoint

test/api_test.js
```diff
@@ -0,0 +1,386 @@
...
+        expect(endpoints.batch())
+          .eql("/v0/batch");
+      });
+
+      it("should provide root endpoint", () => {
+        expect(endpoints.collection("toto"))
+          .eql("/v0/collections/toto/records");
+      });
+
+      it("should provide root endpoint", () => {
```
leplatrem: copy-paste (see above)

test/api_test.js
```diff
@@ -0,0 +1,386 @@
...
+    it("should pass provided headers", () => {
+      sandbox.stub(root, "fetch").returns(Promise.resolve());
+
+      api.fetchChangesSince("articles", 42, {headers: {Foo: "bar"}});
+
+      sinon.assert.calledOnce(fetch);
+      sinon.assert.calledWithMatch(fetch, /\?_since=42/, {
+        headers: {Foo: "bar"}
+      });
+    });
```
leplatrem: I would get rid of this test, since the next one covers it up pretty well.

src/api.js
```diff
@@ -0,0 +1,139 @@
...
+          throw Object.keys(res).reduce((err, key) => {
+            if (key !== "message")
+              err[key] = res[key];
+            return err;
+          }, new Error("BATCH request failed: " + res.message));
+        res.responses.forEach(response => {
+          if (response.status && response.status >= 200 && response.status < 400) {
+            results.published.push(response.body);
+          } else if (response.status === 404) {
+            results.skipped.push(response.body);
```
leplatrem: Note for later: handle  ``409`` when unicity rule is violated (ex. POST with existing id, unique field like RL article URL, etc.)

test/collection_test.js
```diff
@@ -0,0 +1,653 @@
...
+
+    it("should save a lastModified value", () => {
+      return articles.saveLastModified(42)
+        .then(_ => articles.getLastModified())
+        .should.eventually.become(42);
+    });
+
+    it("should update instance lastModified property value", () => {
+      return articles.saveLastModified(42)
+        .then(val => expect(articles.lastModified).eql(val));
```
leplatrem: difference with ``articles.getLastModified()`` ?

src/api.js
```diff
@@ -0,0 +1,139 @@
...
+      headers: DEFAULT_REQUEST_HEADERS,
+      body: JSON.stringify({
+        defaults: { headers },
+        requests: records.map(record => {
+          const isDeletion = record._status === "deleted";
+          const path = this.endpoints({full: false}).record(collName, record.id);
+          const method = isDeletion ? "DELETE" : "PUT";
+          const body = isDeletion ? undefined : cleanRecord(record);
+          const headers = options.safe && record.last_modified ?
+                          {"If-Unmodified-Since": String(record.last_modified)} : {};
```
leplatrem: If record.last_modified is not set, then it's a creation.
In order to prevent PUT to override a potential existing record in this case, it would be safer to send ``If-Unmodified-Since: 0``. This way the PUT will fail to create with 412 if there is a record with this id already.

test/collection_test.js
```diff
@@ -0,0 +1,653 @@
...
+    it("should reject on virtually deleted record", () => {
+      return articles.delete(uuid)
+        .then(res => articles.get(uuid))
+        .should.be.rejectedWith(Error, /not found/);
+    });
+
+    it("should prefix error encountered", () => {
+      sandbox.stub(articles, "open").returns(Promise.reject("error"));
+      return articles.get().should.be.rejectedWith(Error, /^get/);
+    });
```
leplatrem: Add "should support the includeDeleted option" :)

test/collection_test.js
```diff
@@ -0,0 +1,653 @@
...
+
+  describe("#delete", () => {
+    var articles, uuid;
+
+    beforeEach(() => {
+      articles = testCollection();
+      return articles.create(article)
+        .then(result => uuid = result.data.id);
+    });
+
```
leplatrem: Add a test to assert the default value of virtual when not passed :)

test/collection_test.js
```diff
@@ -0,0 +1,653 @@
...
+      beforeEach(() => {
+        return articles.create({title: "art2"})
+          .then(res => {
+            createdId = res.data.id;
+            sandbox.stub(Api.prototype, "fetchChangesSince").returns(
+              Promise.resolve({
+                lastModified: 42,
+                changes: [
+                  {id: createdId, title: "art2"}, // resolvable conflict
```
leplatrem: Could you please explicit the situation you're trying to reproduce here ?

test/collection_test.js
```diff
@@ -0,0 +1,653 @@
...
+    });
+
+    it("should resolve early on pull failure", () => {
+      const result = new SyncResultObject();
+      result.add("conflicts", [1]);
+      sandbox.stub(articles, "pullChanges").returns(Promise.resolve(result));
+      return articles.sync()
+        .should.eventually.become(result);
+    });
+  });
```
leplatrem: I don't know how we could test that, but that would be nice to have a test that specifies the order of the operations (pull / push / pull)

test/index_test.js
```diff
@@ -1,151 +1,53 @@
...
 chai.use(chaiAsPromised);
 chai.should();
+chai.config.includeStack = true;
 
 const TEST_COLLECTION_NAME = "cliquetis-test";
+const FAKE_SERVER_URL = "http://fake-server"
 
-describe("Cliquetis", function() {
+describe("Cliquetis", () => {
+  var sandbox;
```
leplatrem: Add a test : should be instantiated with localhost 8888 if no server provided

README.md
```diff
@@ -123,6 +134,12 @@ Result is:
 
 ### Deleting records
 
+**Status:** Implemented.
+
+By default, local deletion is performed *virtually*, until the collection is actually synced to the remote server.
```
ametaireau: earlier in the text you were using "sync'ed".

README.md
```diff
@@ -178,10 +198,12 @@ Result is:
 }
 ```
 
-> Records with ``last_modified`` attribute were sync'ed on a server.
+> Records with `last_modified` attribute were sync'ed on a server.
```
ametaireau: "with the"

README.md
```diff
@@ -225,46 +253,90 @@ Result:
 
 ### Fetching and publishing changes
 
-Synchronizing local data with remote ones is performed by calling the `.sync()` method:
+**Status:** Partially implemented.
+
+Synchronizing local data with remote ones is performed by calling the `.sync()` method.
```
ametaireau: Earlier in the doc you where using `#method()`.

README.md
```diff
@@ -225,46 +253,90 @@ Result:
...
+
+Synopsis:
+
+1. Fetch remote changes since last synchronization;
+2. Fail on any conflict detected;
+  * The developer has to handle them manually, and call `sync ()` again when done;
```
ametaireau: `#sync()` ?

README.md
```diff
@@ -31,9 +29,15 @@ const db = new Cliquetis(options);
 
 - `remote`: The remote Cliquet server endpoint root URL (eg. `"https://server/v1"`);
 - `headers`: The default headers to pass for every HTTP request performed to the Cliquet server (eg. `{"Authorization": "Basic bWF0Og=="}`);
-- `mode`: The conflict default resolution strategy (`Cliquet.SAFE` (*server wins*), `Cliquet.FORCE` (*client wins*)).
+- `mode`: The conflict default resolution strategy (`Collection.strategy.SERVER_WINS`, `Collection.strategy.CLIENT_WINS` or `Collection.strategy.MANUAL` (default)
```
ametaireau: should it be `mode` or `strategy` ?

README.md
```diff
@@ -225,46 +253,90 @@ Result:
...
+  ok: true,
+  lastModified: 1434270764485,
+  errors:    [], // Errors encountered, if any
+  created:   [], // Created locally
+  updated:   [], // Updated locally
+  deleted:   [], // Deleted locally
+  conflicts: [], // Import conflicts
+  skipped:   [], // Skipped imports
+  published: [], // Successfully published
+  conflicts: []  // Export conflicts
```
ametaireau: there are two `conflicts` keys.

demo/index.html
```diff
@@ -0,0 +1,161 @@
...
+    <input type="reset" id="clearAllBtn" value="Clear collection">
+    <input type="button" id="syncBtn" value="Synchronize collection">
+  </p>
+  <p>
+    <textarea id="log" class="log"></textarea>
+  </p>
+  <script src="../dist/cliquetis.js"></script>
+  <script>
+  function main() {
+    const db = new Cliquetis({remote: "http://0.0.0.0:8888/v0"});
```
ametaireau: I would expect cliquetis to decide the version number to use, since each cliquetis is tied to a version of the server. As such, I would think the version shouldn't be present.

src/api.js
```diff
@@ -0,0 +1,139 @@
...
+      throw new Error("Invalid remote URL: " + remote);
+    this.remote = remote;
+    try {
+      this.version = "v" + remote.match(/\/v(\d+)\/?$/)[1];
+    } catch (err) {
+      throw new Error("The remote URL must contain the version: " + remote);
+    }
+  }
+
+  endpoints(options={full: true}) {
```
ametaireau: full what?

src/api.js
ametaireau: Makes me think we need to have a mechanism to know what's the configuration of the server for pagination. We probably should expose that in either an extra endpoint or /.

ametaireau: See https://github.com/mozilla-services/cliquet/issues/318

src/collection.js
```diff
@@ -44,143 +109,422 @@ export default class Collection {
...
+        }, {toDelete: [], toSync: []});
+      });
+  }
+
+  /**
+   * Import remote changes to the local database. Will reject on encountered
+   * conflicts.
+   *
+   * @param  {SyncResultObject} syncResultObject
+   * @param  {Options} options
```
ametaireau: {Object}

src/collection.js
```diff
@@ -44,143 +109,422 @@ export default class Collection {
...
+    return this.api.fetchChangesSince(this.name, this.lastModified, options)
+      // Reflect these changes locally
+      .then(changes => this.importChanges(syncResultObject, changes));
+  }
+
+  /**
+   * Publish local changes to the remote server.
+   *
+   * @param  {SyncResultObject} syncResultObject
+   * @param  {Options} options
```
ametaireau: {Object}

src/collection.js
```diff
@@ -44,143 +109,422 @@ export default class Collection {
...
+
+
+  /**
+   * Synchronize remote and local data. The promise will resolve with two lists:
+   * - local imports
+   * - remote exports
+   * The promise will reject if conflicts have been encountered, with the same
+   * result.
+   *
+   * @param  {Object} options options
```
ametaireau: options options ?

src/collection.js
```diff
@@ -44,143 +109,422 @@ export default class Collection {
...
+   * @return {Promise}
+   */
+  sync(options={mode: Collection.strategy.MANUAL, headers: {}}) {
+    // TODO rename options.mode to options.strategy
```
ametaireau: ah!

README.md
n1k0: Totally missed to update that, will update.

README.md
n1k0: Erm, this is being totally revamped in a next docs PR. I'll note that though.

README.md
n1k0: Good catch.

demo/index.html
n1k0: How the client could possibly decide what is the right version to use? Also, I'm finding specifying the version that way very explicit.

src/api.js
n1k0: aka fully qualified. `fullyQualified`? yuk…

demo/index.html
leplatrem: You are right. 
First I would have thought that the number comes from the version of the protocol (for example this PR is only compatible with ``kinto/v0``). But since the version number is the one of the Cliquet project, we cannot set it here as constant.

The only thing we have to have to get it from server is to read the location header of the ``/`` response

ametaireau: One cliquetis client should be tied to one major version of cliquet at one particular point in time. As such, it's not possible to have one version of cliquet specifying a different version than what it's able to handle, otherwise unexpected behaviours will happen.

For instance, with the upcoming changes in Cliquet, changing the version in cliquetis will not arrange anything, the client code needs to be updated, and then it will not be compatible with the old version of the APIs.

src/api.js
n1k0: Went with `fullUrl` and some jsdoc.

src/api.js
n1k0: Just to be sure I properly understand, would this be okay?

```js
var headers;
if (!options.safe)
  headers = {};
else if (record.last_modified)
  headers = {"If-Unmodified-Since": String(record.last_modified)};
else
  headers = {"If-Unmodified-Since": "0"};
```

src/api.js
n1k0: Nope, it's the error format sent by the server (along with `errno`, `message` and so on)

src/api.js
leplatrem: Yes or this:

```javascript
var headers = {};
if (options.safe)
  headers = {"If-Unmodified-Since": String(record.last_modified || 0)};
```

src/api.js
n1k0: I'll add a TODO and create an issue.

src/collection.js
n1k0: Well, the `synced` name is the name of the status it reflects — I'm finding it pretty much explicit. I'll update jsdoc to better reflect its purpose if it's not clear.

src/collection.js
n1k0: For now; will carry the filtering/sorting parameters later.

src/collection.js
n1k0: Good point.

src/api.js
n1k0: Sure.

test/collection_test.js
n1k0: How would you do that?

test/collection_test.js
n1k0: Good catch!

test/collection_test.js
n1k0: Automatic reconciliation of perfectly matching records (https://github.com/mozilla-services/cliquetis/blob/first-sync-impl/src/collection.js#L321)

test/index_test.js
n1k0: Yes!

demo/index.html
n1k0: I'm confused. What should I do?

PR #17, opened 2015-06-17
Versionned dist file to ease sharing.

n1k0:
Versioning the output library file will help developers to toy around the API without having to build the lib setting up a full node environment.


Natim:
Should we also provide an non uglified version?

ametaireau:
Distributing these files on the github GH pages could also help!

n1k0:
> Should we also provide an non uglified version?

Good idea. Like many projects do, I can come with at least two versions, minified with source maps for dev, and uglified + no source maps for production.

n1k0:
I'm merging this. r=me

PR #18, opened 2015-06-17
Add a ``make serve-demo`` command.

Natim:
For now I am using python SimpleHTTPServer but maybe there is a way to use ``npm run demo`` instead?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,2 +1,3 @@
 serve-demo:
+	cp -fr dist	demo/
 	cd demo/; python -m SimpleHTTPServer 9000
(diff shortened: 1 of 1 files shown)


n1k0:
I really wanted to avoid adding a Makefile here. Could we rely on `package.json` scripts, as for all the other commands? We could use [http-server](https://github.com/indexzero/http-server).

Natim:
I understand that, what I like with Makefile is the self-documentation feature. It is more readable than with package.json but I guess it is a matter of taste. (can't we have both? Makefile calling the package script ?)

leplatrem:
+1 to follow conventions, whatever that means :)

PR #19, opened 2015-06-17
Port Cliquetis for Kinto 1.0

Natim:
* [x] Add notion of bucket (``bucket`` name in Cliquetis constructor)
* [x] ``data`` attribute posted in request body
* [x] ``data`` attribute retrieved in response body
* [x] ``If-Modified-Since: X`` -->  ``If-Match: "X"`` (mind quotes)
* [x] ``If-Unmodified-Since: X`` -->  ``If-None-Match: "X"`` (mind quotes)
* [x] ``If-Unmodified-Since: 0`` -->  ``If-None-Match: *`` (no quotes)


ametaireau:
There is also the change with the `data` attribute and changes with `Etags` rather than `IMS`.

src/api.js
```diff
@@ -40,10 +42,11 @@ export default class Api {
    */
   endpoints(options={fullUrl: true}) {
     var root = options.fullUrl ? this.remote : `/${this.version}`;
+    var bucket = this.bucket || '~';
```
ametaireau: Do we support this just yet?

src/api.js
```diff
@@ -40,10 +42,11 @@ export default class Api {
...
     return {
       root:           () => root,
       batch:          () => `${root}/batch`,
-      collection: (coll) => `${root}/collections/${coll}/records`,
+      collection: (coll) => `${root}/buckets/${bucket}/collections/${coll}/records`,
```
n1k0: 1. Shouldn't we add a `bucket` endpoint?
2. We need this to be tested :)

Natim:
This is not ready just yet, please wait for my review calls :)

src/api.js
```diff
@@ -40,47 +42,56 @@ export default class Api {
...
-      batch:          () => `${root}/batch`,
-      collection: (coll) => `${root}/collections/${coll}/records`,
-      record: (coll, id) => `${this.endpoints(options).collection(coll)}/${id}`,
+    var urls = {
+      root: ()                   => root,
+      batch: ()                  => `${root}/batch`,
+      bucket: (bucket)           => `${root}/buckets/${bucket}`,
+      collection: (bucket, coll) => `${urls.bucket(bucket)}/collections/${coll}`,
+      records: (bucket, coll)    => `${urls.collection(bucket, coll)}/records`,
+      record: (bucket, coll, id) => `${urls.records(bucket, coll)}/${id}`,
```
n1k0: Nit: how about keeping the function aligned on the right as previously?

test/api_test.js
```diff
@@ -69,21 +70,31 @@ describe("Api", () => {
...
+        expect(endpoints.record("foo", "bar", 42))
+          .eql("http://fake-server/v0/buckets/foo/collections/bar/records/42");
       });
     });
 
     describe("absolute URL", () => {
       var endpoints;
 
-      beforeEach(() => endpoints = api.endpoints({fullUrl: false}))
+      beforeEach(() => endpoints = api.endpoints({ fullUrl: false }))
```
n1k0: Nit: we should keep consistent regarding inline object definitions; here we're using spaces around, later we're not anymore.

n1k0:
r=@n1k0, provided the demo actually works with the deployed dev version of Kinto 2.1.

n1k0:
This is now ready to land.

PR #22, opened 2015-06-18
Switch to ETags instead of IMS-IUMS

leplatrem:
In Cliquet 2.0, If-Modified-Since and If-Unmodified-Since
headers were replaced by ETags.

See https://github.com/mozilla-services/cliquet/issues/251

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -130,8 +130,7 @@ export default class Api {
             if (record.last_modified) {
               // Safe replace.
               headers["If-Match"] = quote(record.last_modified);
-            }
-            else if (!isDeletion) {
+            } else if (!isDeletion) {
               // Safe creation.
               headers["If-None-Match"] = "*";
             }
--- a/src/utils.js
+++ b/src/utils.js
@@ -26,6 +26,8 @@ export function attachFakeIDBSymbolsTo(obj) {
 
 /**
  * Returns the specified string with double quotes.
+ * @param  {String} str  A string to quote.
+ * @return {String}
  */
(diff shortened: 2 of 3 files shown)


src/api.js
```diff
@@ -73,7 +78,8 @@ export default class Api {
           // TODO: attach better error reporting
           throw new Error("Fetching changes failed: HTTP " + res.status);
         } else {
-          newLastModified = res.headers.get("Last-Modified");
+          var etag = res.headers.get("ETag");  // e.g. '"42"'
+          newLastModified = parseInt(etag.substring(1, etag.length - 1));
```
Natim: Didn't we say that we weren't doing any assumption about what is in the ETag?

src/api.js
```diff
@@ -116,9 +122,17 @@ export default class Api {
...
-          } : {};
+          const headers = {};
+          if (options.safe) {
+            if (record.last_modified) {
+              // Safe replace.
+              headers["If-Match"] = `"${record.last_modified}"`;
+            }
+            else if (!isDeletion) {
+              // Safe creation.
+              headers["If-None-Match"] = "*";
```
Natim: We should add a test for this.

src/api.js
n1k0: Pass the radix to `parseInt()`

src/api.js
leplatrem: Done

leplatrem:
r? = @n1k0 @Natim

PR #24, opened 2015-06-18
Fixes #23 - Don't bump local collection last_modified on conflicts.

### LINKED ISSUE #23: Don't store collection last_modified on sync() when conflicts have been encountered
This prevents calling sync again using an erroneous last_modified value.

n1k0:
refs #23 - r=? @ametaireau @leplatrem @natim

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -383,12 +383,13 @@ export default class Collection {
         return syncResultObject;
       })
       .then(syncResultObject => {
-        // Import is done, save lastModified value received from the server
-        return this.saveLastModified(changeObject.lastModified)
-          .then(lastModified => {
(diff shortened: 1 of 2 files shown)


Natim:
So if I understand correctly:

1. In case of update locally when no conflict, you update the record lastModified value
2. In case of update locally with a conflict, you don't update the record lastModified.

How do you update the record once the conflict has been solved?

n1k0:
>In case of update locally when no conflict, you update the record lastModified value
>In case of update locally with a conflict, you don't update the record lastModified.

Not record's one, collection's one

>How do you update the record once the conflict has been solved?

It's developers responsibility; they needs to solve the conflict locally and call sync() again.

Natim:
r+

PR #21, opened 2015-06-17
Better docs, including tutorial

n1k0:
Refs #10 

- [x] Migrate docs to mkdocs
- [x] Draft tutorial
- [x] Update current demo code to use the tutorial result one
- [ ] Set up RTD integration

PR #27, opened 2015-06-18
Fixes #25 - Conflict resolution helper.

### LINKED ISSUE #25: Implement conflict resolution helper
Suggested API:

```js
collection.sync().then(res => {
  if (res.conflicts.length === 0)
    return;
  return Promise.all(res.conflicts.map(conflict => {
    const resolution = …
    return collection.resolve(conflict, resolution);
  })).then(res => collection.sync());
});
```

`Collection#resolve` should return a Promise being the result of a sucesful local update.

Note: Maybe we should add a new `resolved` entry to `SyncResultObject` to list these?

n1k0:
Refs #25. I took the bare minimum approach here, though seems to serve the purpose.

Docs will be updated along #21.

r=? @leplatrem @ametaireau @Natim

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -186,6 +186,8 @@ export default class Collection {
     return this.open().then(() => {
       if (typeof(record) !== "object")
         return Promise.reject(new Error("Record is not an object."));
+      if (!record.id)
+        return Promise.reject(new Error("Cannot update a record missing id."));
       return this.get(record.id).then(_ => {
         return new Promise((resolve, reject) => {
           var newStatus = "updated";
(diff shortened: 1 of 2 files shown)


src/collection.js
```diff
@@ -186,6 +186,8 @@ export default class Collection {
     return this.open().then(() => {
       if (typeof(record) !== "object")
         return Promise.reject(new Error("Record is not an object."));
+      if (!record.id)
+        return Promise.reject(new Error("Cannot update a record missing id."));
```
Natim: ``missing an id``

Natim:
r+

n1k0:
Thanks! I'll fix the typo along #21

PR #39, opened 2015-06-30
Fixes #11 - Fetch and validate server protocol version.

### LINKED ISSUE #11: Versions should be handled in the client
Currently, Cliquetis handles the version in the host url, with a prefix (e.g. `http://localhost:8888/v1`).

It's important to have a way to tell which version of the protocol is supported by Cliquetis, in order to detect potential problems due to version mismatches.

To do so, I believe a request to the / endpoint of the server should be made, to get the protocol version, and then check that the protocol is the same as the one of Cliquetis.

n1k0:
Refs #11.

- [x] Fetch the protocol version from the server out of the `url` field from the JSON response on `/`
- [x] Validate that the fetched protocol version matches client supported one 
- [x] Update `Api#fetchChangesSince` to ensure protocol version validation is performed
- [x] Update `Api#batch` to ensure protocol version validation is performed
- [x] Cache the fetched protocol version information locally to avoid performing unnecessary requests
- [x] Update docs
- [x] Update demo code

This issue is currently being discussed, hold on merging. Read below.


src/api.js
```diff
@@ -55,6 +58,37 @@ export default class Api {
...
+        return Promise.resolve(check(this.serverVersion));
+      } catch(err) {
+        return Promise.reject(err);
+      }
+    }
+    return fetch(this.remote, {headers: DEFAULT_REQUEST_HEADERS})
+      .then(res => res.json())
+      .then(res => {
+        try {
+          this.serverVersion = "v" + res.url.match(/\/v(\d+)\/?$/)[1];
```
Natim: Why not ``this.serverVersion = res.url.match(/\/(v\d+)\/?$/)[1];``

Natim:
- Does it works if the host contains the version?
- Just to confirm it is calling the home page once and then caching the result?

Why can't we just enforce the SPV and check for the server version in case of a 404 only?

n1k0:
> - Does it works if the host contains the version?

Nope.

> - Just to confirm it is calling the home page once and then caching the result?

Yup.

> Why can't we just enforce the SPV and check for the server version in case of a 404 only?

We need to guard against users using a previous version of the protocol, which might be not compatible with the one Cliquetis currently supports.

Natim:
> We need to guard against users using a previous version of the protocol, which might be not compatible with the one Cliquetis currently supports.

Yes it is what my proposal is doing as well.

If you enforce http://localhost/v1/ then you will get a 404 if the server doesn't supports this version.

n1k0:
After discussing the idea over IRC, it appears a good compromise would be to raise an HTTP 403 *(Authorization will not help and the request SHOULD NOT be repeated)* with a proper error message and errno in the JSON response body, so we:

- avoid the unecessary supplementary hit on `/` the first time a request is sent to the server;
- provide a useful error to the developer.

Also, this solution would quite lighten this patch, which is always good :)

n1k0:
So after having updated demo code, I realize this obviously doesn't work since kinto redirects with a 301 when accessing `/`. Idea: how about exposing the list of versions supported by this server, eg. 

```json
{
  "versions“: [
    "v0": "http://localhost:8888/v0",
    "v1": "http://localhost:8888/v1"
  ]
}
```

That would help picking the right one from the client regarding its expectation, if it's available (if it's not, client can raise early to notify the developer). Also, providing the urls to use will definitely help in case we ever need to change them (I know, if they're cool, they don't; but heh).

Natim:
Actually this is the reason why I want to go for the duck-typing approach.
 - We should assume the developer know what it does and fail if it occurs it doesn't.
 - We should assume the version of cliquetis matches the configured server and fail if it appears it doesn't.

Also listing supported version is a good idea but it adds complexity with the approach taken at the moment. Because we are running a specific server per version and therefore the server doesn't know about the set of supported version.

Natim:
My take is to ask for ``rootPath`` (``scheme://hostname/supported_version/``) and see if we've got a 404 unsupported response.

ametaireau:
+1 for testing the URL and failing if we get a 404 unsupported.

n1k0:
Throwing away this patch, as we've just decided to do things differently.

PR #45, opened 2015-07-02
Await async

n1k0:
Using ES7 async/await allows easing [writing & reading async flows](http://pouchdb.com/2015/03/05/taming-the-async-beast-with-es7.html), and we should probably use it.

What do you think? r=? @leplatrem @Natim @ametaireau


n1k0:
This is now ready for early feedback, possibly full review if you feel like it. r=? @leplatrem @Natim @ametaireau

src/api.js
```diff
@@ -164,37 +159,38 @@ export default class Api {
...
+      } else if (response.status === 412) {
+        results.conflicts.push({
+          type: "outgoing",
+          data: response.body
+        });
+      } else {
+        results.errors.push({
+          // TODO: since responses come in the same order, there should be a
+          // way to get original record id
+          path: response.path, // this is the only way to have the id…
```
Natim: I think we should give another way in Kinto to provide the id.

test/collection_test.js
```diff
@@ -319,7 +317,7 @@ describe("Collection", () => {
 
     it("should prefix error encountered", () => {
       sandbox.stub(articles, "open").returns(Promise.reject("error"));
-      return articles.get().should.be.rejectedWith(Error, /^get/);
+      return articles.get().should.be.rejectedWith(Error, /get\(\)/);
```
leplatrem: why do we still want to return here ?

leplatrem:
I like it ! 

... and feel like we're taking part in the future :)

n1k0:
These ES7 features seem way too fragile to rely on this for Kinto.js. Closing for now, we shall reconsider using async/await when they're stable enough though.

PR #46, opened 2015-07-02
"ReferenceError: render is not defined" and a few typos in tutorial

ferjm:


n1k0:
High five.

Natim:
Should we also update https://github.com/mozilla-services/cliquetis/blob/master/demo/demo.js#L19?

ferjm:
No, that's the final demo code. I removed that line cause at that point of the tutorial we don't have the `render` function defined.

Natim:
Oh ok really good point. Thank you :+1:

PR #37, opened 2015-07-02
Rename project to kinto.js

Natim:
There's been some confusion around the scope of Cliquetis, which in its current form is much more of a Kinto client than a Cliquet one. We should rename the project to reflect the situation and avoid creating more confusion.


Natim:
I don't have necessary rights to create kinto.js.readthedocs.org and I don't even know if it is possible to do so or if we should use kinto-js.readthedocs.org instead.

n1k0:
r+, as I *think* the coverage badge will show up once a report using the new project repo url is generated.

Natim:
The badge actually works: https://coveralls.io/repos/mozilla-services/kinto.js/badge.svg?branch=master
But the image cached by github is not the right one yet.

PR #50, opened 2015-07-06
Added full integration test suite.

n1k0:
This patch adds integration tests by setting up a local kinto server instance and run a test suite against it.

r=? @ametaireau @Natim @leplatrem

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -163,7 +163,8 @@ export default class Collection {
    * Adds a record to the local database.
    *
    * Options:
-   * - {Boolean} synced: Sets record status to "synced" (default: false)
+   * - {Boolean} synced: Sets record status to "synced" (default: false);
+   * - {Boolean} forceUUID: Enforces record creation using any provided UUID.
    *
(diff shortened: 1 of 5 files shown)


n1k0:
This is now ready for review. Take your time. r=? @ametaireau @Natim @leplatrem

ametaireau:
Very cool to see how the actual code is impacted. Proves @Natim point about unit tests :x

ametaireau:
r+

src/collection.js
```diff
@@ -169,14 +169,14 @@ export default class Collection {
    * @param  {Object} options
    * @return {Promise}
    */
-  create(record, options={synced: false}) {
+  create(record, options={forceUUID: false, synced: false}) {
```
ametaireau: Should we expose this in the documentation somehow?

n1k0: I debated this, but I couldn't see any use case for end users to use this option, which I've added for test context creation only. Your call.

PR #51, opened 2015-07-06
Fixes #12 - Improved Api#fetchChangesSince error messages.

### LINKED ISSUE #12: Improve error reporting when fetching changes failed.
```diff
   71            return {items: []};
   72          } else if (res.status >= 400) {
   73:           // TODO: attach better error reporting
   74            throw new Error("Fetching changes failed: HTTP " + res.status);
   75          } else {
```
We'd need to distinguish between possible 4xx and 5xx status codes so the end user knows more precisely what happened.

n1k0:
Refs #12.

r=? @Natim @leplatrem @ametaireau


src/api.js
```diff
@@ -79,21 +89,27 @@ export default class Api {
 
     return fetch(recordsUrl + queryString, {headers})
       .then(res => {
+        response = res;
```
Natim: can't we use response directly instead of res?

n1k0: You mean using res directly? We're assigning the result to the response var available from the outter scope, so next Promise steps can still access the response object.

Natim: Oh ok yes you are right.

n1k0:
Let me land that baby.

PR #52, opened 2015-07-07
Closes #6 - Local collection data ordering & filtering.

### LINKED ISSUE #6: Implement local list filtering
Target API:

```js
// Single filter
articles.list({
  filter: { unread: { $eq: true } }
}).then(console.log.bind(console));

// Combined field filtering
articles.list({
  filter: {
    added_on: { 
      $gt: 1432800794000, 
      $lt: 1432800795000 
    }
  }
}).then(console.log.bind(console));

// Multiple fields filtering (AND only)
articles.list({
  filter: { 
    unread: { $eq: true }, 
    added_on: { $gt: 1432800794000 } 
  }
}).then(console.log.bind(console));
```

First minimal batch of operators to support:

- `$eq`: Strict equality check (strings, numbers);
- `$contains`: Partial equality check (strings only);
- `$gt`, `$gte`: Greater than, greater than equal (strings, numbers);
- `$lt`, `$lte`: Lesser than, lesser than equal (strings, numbers).

#### Notes

- As we're using schemaless stores, we can't rely on indexed fields here (because we can't predict them);
- All queries are performed using the AND condition logic.

n1k0:
WiP: do not merge yet.

- [x] Ordering
- [x] Filtering
- [x] Combined
- [x] Tests
- [ ] Docs

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -48,13 +48,38 @@ export function unquote(str) {
  * @param  {Array}  list     The collection to order.
  * @return {Array}
  */
-export function sortObjects(ordering, list) {
-  const hasDash = ordering[0] === "-";
-  const field = hasDash ? ordering.slice(1) : ordering;
+export function sortObjects(order, list) {
+  const hasDash = order[0] === "-";
+  const field = hasDash ? order.slice(1) : order;
   const direction = hasDash ? -1 : 1;
   return list.slice().sort((a, b) => {
     if (!a[field] || !b[field])
       return 0;
     return a[field] > b[field] ? direction : -direction;
   });
+}
+
+/**
+ * Filters records in a list matching all given filters.
+ * @param  {String} filters  The filters object.
+ * @param  {Array}  list     The collection to order.
+ * @return {Array}
+ */
+export function filterObjects(filters, list) {
+  return list.filter(entry => {
+    return Object.keys(filters).every(filter => {
+      return entry[filter] === filters[filter];
+    });
+  });
+}
+
+/**
+ * Filter and sort list against provided filters and order.
+ * @param  {Object} filters  The filters to apply.
+ * @param  {String} order    The order to apply.
+ * @param  {Array}  list     The list to reduce.
+ * @return {Array}
+ */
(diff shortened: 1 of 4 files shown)


src/utils.js
```diff
@@ -41,3 +41,45 @@ export function quote(str) {
 export function unquote(str) {
   return str.replace(/^"/, "").replace(/"$/, "");
 }
+
+/**
+ * Sorts records in a list according a given ordering.
```
Natim: according to a given ordering

Natim:
LGTM

ametaireau:
Looks good to me as well. I was a bit hesitant with the `reduceRecord` function name since it doesn't really tells what it does, but find `filterAndOrderRecords` way too long to be a function name.

test/collection_test.js
```diff
@@ -391,39 +391,125 @@ describe("Collection", () => {
...
+        return articles.list({order: "title"})
+          .then(res => res.data.map(r => r.title))
+          .should.eventually.become(["art1", "art2", "art3"]);
+      });
+
+      it("should filter records on custom field DESC", () => {
+        return articles.list({order: "-title"})
+          .then(res => res.data.map(r => r.title))
+          .should.eventually.become(["art3", "art2", "art1"]);
+      });
```
leplatrem: Could you please explicit the boolean order in a test ? (false first ?)

src/utils.js
```diff
@@ -41,3 +41,45 @@ export function quote(str) {
...
+ * @param  {Array}  list     The collection to order.
+ * @return {Array}
+ */
+export function sortObjects(order, list) {
+  const hasDash = order[0] === "-";
+  const field = hasDash ? order.slice(1) : order;
+  const direction = hasDash ? -1 : 1;
+  return list.slice().sort((a, b) => {
+    if (!a[field] || !b[field])
+      return 0;
```
leplatrem: I would say that if a value is defined for this field in one of the two records, it comes first (undefined comes after)

leplatrem:
r+ great stuff :)

PR #60, opened 2015-07-13
Fixes #13 - Batch requests chunking.

### LINKED ISSUE #13: Handle BATCH request limit

n1k0:
r=? @leplatrem @Natim @ametaireau


src/api.js
```diff
@@ -163,6 +199,20 @@ export default class Api {
       return Promise.resolve(results);
     return this.fetchServerSettings()
       .then(serverSettings => {
+        const maxRequests = serverSettings["cliquet.batch_max_requests"];
+        if (records.length > maxRequests) {
+          return Promise.all(partition(records, maxRequests).map(chunk => {
+            return this.batch(bucketName, collName, chunk, options);
+          }))
+            .then(batchResults => {
+              return batchResults.reduce((acc, batchResult) => {
```
leplatrem: This would deserve a small comment IMO (e.g. assemble responses of chunked batch into one single result array)

src/api.js
```diff
@@ -135,6 +136,45 @@ export default class Api {
...
+      // TODO: handle 409 when unicity rule is violated (ex. POST with
+      // existing id, unique field, etc.)
+      if (response.status && response.status >= 200 && response.status < 400) {
+        results.published.push(response.body.data);
+      } else if (response.status === 404) {
+        results.skipped.push(response.body);
+      } else if (response.status === 412) {
+        results.conflicts.push({
+          type: "outgoing",
+          local: records[index],
```
leplatrem: I think this relies on two uncertain factors, that would deserve a test: ``Promise.all`` and the reduce function should respect the order of their input

n1k0: That's a very good point, as `Promise.all` runs concurrently, not sequentially. Will update code.

n1k0: PS: for this very method, the indexes of passed `records` can't change really; though the chunks being processed concurrently is a very valid issue.

n1k0: So it appears that while `Promise.all` spawns promises concurrently, order is guaranteed:

```js
```

leplatrem:
r+ :+1:

n1k0:
And one more test to be 100% sure :)

PR #61, opened 2015-07-13
Fixes #28 - Validate uuids

### LINKED ISSUE #28: Validate passed uuids
This would apply to `#get()`, `#update()`, `#delete()` and `#resolve()`. We should throw when an id doesn't match the UUID format. Tests would have to be updated accordingly.

For `#resolve()`, we should also validate that the resolution object uuid matches the one from the conflict, and reject accordingly.

n1k0:
Ready for review. r=? @Natim @leplatrem @ametaireau


src/utils.js
```diff
@@ -1,5 +1,7 @@
 "use strict";
 
+const RE_UUID = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
```
leplatrem: This one is a bit more permissive than the server one (https://github.com/mozilla-services/cliquet/blob/master/cliquet/storage/generators.py#L54-L55)

src/utils.js
```diff
@@ -115,3 +117,13 @@ export function partition(array, n) {
...
   }, []);
 }
+
+/**
+ * Checks if a string is an UUID, according to RFC4122.
+ *
+ * @param  {String} uuid The uuid to validate.
+ * @return {Boolean}
+ */
+export function isUUID(uuid) {
```
leplatrem: version 4: isUUID4 ?

test/utils_test.js
```diff
@@ -175,17 +176,32 @@ describe("Utils", () => {
...
     });
   });
+
+  describe("#isUUID", () => {
+    it("should check that a string uses a valid UUID format", () => {
+      expect(isUUID("110ec58a-a0f2-4ac4-8393-c866d813b8d1")).eql(true);
+    });
+
+    it("should check that a string does not use a valid UUID format", () => {
+      expect(isUUID("110ex58a-a0f2-4ac4-8393-c866d813b8d1")).eql(false);
```
leplatrem: In case you want some more : https://github.com/mozilla-services/cliquet/blob/master/cliquet/tests/test_storage.py#L35-L38

src/utils.js
Natim: version4 is not the format it is the way they are generated. (4 means every part is random)

leplatrem: That's not what wikipedia says https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_.28random.29

Natim: We shouldn't force uuid4 then.

leplatrem: Ah, maybe then :) please open a Cliquet issue, since the original regexp comes from there

Issue #33, opened 2015-06-23
Remote URL should have a trailing slash.


ametaireau:
It allows to copy and paste the URL directly from the browser.

n1k0:
Fixed along #63

PR #63, opened 2015-07-14
Automatically drop the trailing slash from passed remote url.

n1k0:
r=? @leplatrem @Natim


n1k0:
This would also fix #33 btw.

Natim:
r+

PR #65, opened 2015-07-15
Add safety check if batch limit setting is falsy (fixes #64)

### LINKED ISSUE #64: Support unlimited batch requests
In Kinto, if the setting ``cliquet.batch_max_requests`` is set to ``None``, then the server has no limit for batch requests.

Currently, Kinto.js expects a value for this setting.

leplatrem:
I didn't think of this during the review...

@n1k0 r?

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -197,7 +197,7 @@ export default class Api {
     return this.fetchServerSettings()
       .then(serverSettings => {
         const maxRequests = serverSettings["cliquet.batch_max_requests"];
-        if (records.length > maxRequests) {
+        if (maxRequests && records.length > maxRequests) {
           return Promise.all(partition(records, maxRequests).map(chunk => {
             return this.batch(bucketName, collName, chunk, options);
           }))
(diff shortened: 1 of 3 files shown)


n1k0:
Why would it be falsy? I think we decided not supporting previous versions of the protocol version…

n1k0:
Ok, I've read the rationale in #64 and it makes sense.

Coveralls is down atm hence the red build, though code looks good, so r+ once builds are green.

leplatrem:
I can't figure out why my changes would have an impact on Kinto server flush and break the builds :) The sandbox is supposed to be restored after each test, isn't it?

n1k0:
I don't know; FWIW, tests pass on your branch locally.

n1k0:
Also, the failing test is an integration one:

```
  1) Integration tests "before each" hook:
     Error: request to http://0.0.0.0:8888/v1/__flush__ failed, reason: connect ECONNREFUSED
      at Socket.socketErrorListener (_http_client.js:271:9)
      at net.js:459:14
```

I can't see how connection to the test server could suddenly become refused due to your patch…

n1k0:
Grmbl race condition. I'll update the sleep time a little so we're safe, and rebase to remove that commit message.

leplatrem:
GG! Thanks! :)

PR #72, opened 2015-07-15
Fixes #71 - Simplified empty HTTP response handling.

### LINKED ISSUE #71: Error on synchronization with a HTTP 304 on the collection from the server
```HTTP304 - SyntaxError: JSON.parse: unexpected end of data at line 1 column 1 of the JSON data```

### PR DESCRIPTION
**Pull request by n1k0**
r=? @Natim @leplatrem 

Note: will deploy that patch to the demo instance for live validation.

### CODE REVIEW

**Comment by n1k0** [comment]
> Note: will deploy that patch to the demo instance for live validation.

Deployed.

**Comment by n1k0** [comment]
> Deployed.

This patch breaks things, don't merge ^^'

File: src/http.js [hunk 1]
```diff
@@ -29,7 +29,8 @@ export default function request(url, options={headers:{}}) {
       headers = res.headers;
       status = res.status;
       statusText = res.statusText;
-      if (headers.get("Content-Length") == 0) // 0 or "0"
+      const contentLength = headers.get("Content-Length");
+      if (!contentLength || contentLength == 0)
```
leplatrem: ``===`` ?

n1k0: Nope, because that way both `0` and `"0"` are handled :)

n1k0:
Pending on https://github.com/mozilla-services/cliquet/pull/390 being merged.

n1k0:
Confirmed that fixes the issue, merging.

Issue #74, opened 2015-07-15
Builds are marked green even if failures occured.


n1k0:
Sample build: https://travis-ci.org/mozilla-services/kinto.js/builds/71136831

```
…
  177 passing (13s)
  1 failing
  1) Integration tests "before each" hook:
     Error: timeout of 12500ms exceeded. Ensure the done() callback is being called in this test.
…
The command "npm test" exited with 0.
```

n1k0:
Fixed by #76.

PR #73, opened 2015-07-15
Upgrade to Kinto 1.3.1.

n1k0:

PR #76, opened 2015-07-16
Refs #74 - Removed coverage from continuous integration.

n1k0:
After speding a little while trying to come with a patch for [isparta](https://github.com/douglasduteil/isparta) to stop marking command results are okay [while they were actually failing](https://github.com/douglasduteil/isparta/issues/48), I decided we should simply drop coverage support from our continous integration process.

We shall abviously consider readding support for this if the issue is fixed someday.


n1k0:
Note: These failing builds are actually what's expected here; now we've droppped coverage, the failing tests are actually marking the builds are failing, providing this vital information (yeah, current master is broken because of kinto test server setup issues).

n1k0:
I'm gonna land this with a failing build, so that will hopefully put some pressure (read: self-pressure) for #73 to be fixed quickly.

PR #79, opened 2015-07-16
Fixes #69 - Handle deprecation header.

### LINKED ISSUE #69: Handle deprecation headers
As specified here http://cliquet.readthedocs.org/en/latest/api/deprecation.html, the client should be aware of ``Alert`` headers, and use them to effectively alert the users/developers

n1k0:
r=? @leplatrem @Natim

### DIFF
--- a/test/http_test.js
+++ b/test/http_test.js
@@ -101,7 +101,8 @@ describe("request()", () => {
         status: 200,
         headers: {
           get(name) {
-            return "fake";
+            if (name !== "Alert")
+              return "fake";
           }
         },
         json() {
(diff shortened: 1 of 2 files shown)


Natim:
LGTM, also should we add a test for the hard-eol?

test/http_test.js
```diff
@@ -101,7 +101,8 @@ describe("request()", () => {
         status: 200,
         headers: {
           get(name) {
-            return "fake";
+            if (name !== "Alert")
+              return "fake";
```
leplatrem: You could also always return an empty string (since the scope of this test is about response anyway)

leplatrem:
I would suggest that kinto.js let the developer provide a hook for this. By default it can be console.warn, but it can be useful for developers to choose if they want to use window.alert or show a nice warning in their UI...

The hard-eol means the server returns ``410 Gone`` to every request. I think it will go through every ``.catch()`` error, no ?

n1k0:
> I would suggest that kinto.js let the developer provide a hook for this. By default it can be console.warn, but it can be useful for developers to choose if they want to use window.alert or show a nice warning in their UI...

Sounds like we'd need an event system for that. I suggest this is not part of this patch, but I'll create antoger issue for this.

> The hard-eol means the server returns 410 Gone to every request. I think it will go through every .catch() error, no ?

Yes, it's already tested, though I can add another explicit (though redundant) test for it if you really really want it.

Natim:
> Itt's already tested, though I can add another explicit (though redundant) test for it if you really really want it.

If it is already tested forget it :)

leplatrem:
r+ then

PR #80, opened 2015-07-16
No sudo for travis

magopian:


leplatrem:
Shouldn't you explicit ``sudo: false`` then ?

n1k0:
> Shouldn't you explicit sudo: false then ?

Well, that's the documented Travis default. Merging this, so now I officially owe you two pints of whatever beverage you like :)

PR #82, opened 2015-07-17
Fixes #68 - Handle Backoff header.

### LINKED ISSUE #68: Handle backoff indicators
As specified here http://cliquet.readthedocs.org/en/latest/api/backoff.html

The clients should be aware of ``Backoff`` and ``Retry-After`` response headers, in order to give the server a breathe :)

n1k0:
r=? @leplatrem @Natim


n1k0:
I just realized we need a request queue in order to avoid getting a disordered requests chain when the backoff release timestamp is reset. That's slightly more complicated than initially expected; I wish we'll actually use that feature :)

n1k0:
This is really a bunch of work; especially, we'd need to think what are the consequences not having the response immediately while requests are queued, and what happens precisely when a bunch of them are performed when the backoff release timestamp is reached.

A very simple solution would be to simply trigger and error when a Backoff header is received, and that would be developers' responsibility to hold on performing more requests and/or handling the queuing by themselves.

I'm voting for the latter, though would love getting some feedback :)

Natim:
Actually the Backoff is not an error. IMHO it is better not to do anything (like queuing or stopping requests) rather than raising error.

n1k0:
After discussion, here's the new plan:

- On Backoff release timestamp being set, reject any call to `Collection#sync()`;
- Provide an option to `Collection#sync()` to force the synchronization even if it's set.

n1k0:
@Natim r=?

src/collection.js
```diff
@@ -589,12 +589,18 @@ export default class Collection {
    *     Conflicting server records will be overriden with local changes.
    *   * `Collection.strategy.MANUAL`:
    *     Conflicts will be reported in a dedicated array.
+   * - {Boolean} forceBackoff: Force synchronization even if server is currently
```
Natim: Maybe ``ignoreBackoff`` is better?

src/collection.js
```diff
@@ -589,12 +589,18 @@ export default class Collection {
...
+   *   backed off.
    *
    * @param  {Object} options Options.
    * @return {Promise}
    */
-  sync(options={strategy: Collection.strategy.MANUAL, headers: {}}) {
-    // TODO rename options.mode to options.strategy
+  sync(options={strategy: Collection.strategy.MANUAL, headers: {}, forceBackoff: false}) {
+    // Handle server backoff: XXX test
```
Natim: Do you want me to setup the Backoff header on the dev server?

src/utils.js
```diff
@@ -127,3 +127,12 @@ export function partition(array, n) {
...
   return RE_UUID.test(uuid);
 }
+
+/**
+ * Returns the current UNIX timestamp.
+ *
+ * @return {Number}
+ */
+export function getUnixTime() {
+  return Math.floor(new Date().getTime() / 1000);
```
Natim: Maybe we should have everything in ms on the client side, since ``Date.now()`` return milliseconds, we can just put backoffRelease in ms and not bother about ms to s conversion.

test/http_test.js
```diff
@@ -14,161 +12,192 @@ chai.config.includeStack = true;
...
-    });
+        return http.on("backoff", value => {
+          expect(value).eql(2000);
+        }).request("/");
+      });
 
-    it("should handle deprecation header parse error", () => {
-      sandbox.stub(root, "fetch").returns(
-        fakeServerResponse(200, {}, {Alert: "dafuq"}));
+      it("should emit a backoff event on missing Backoff header", () => {
```
Natim: What does it means it should emit a backoff event when no Backoff header? To stop the backoff. Is this to release backoff when no more backoff are sent?

src/collection.js
n1k0: That's better indeed, will update.

src/collection.js
n1k0: That's an obsolete comment which should be removed. Maybe you can just explain me how to enable it on a local kinto instance so I can quickly test that everything works as expected.

src/utils.js
n1k0: Yeah, probably.

test/http_test.js
n1k0: > Is this to release backoff when no more backoff are sent?

Yes.

n1k0:
@Natim r=? regarding your comments.

I'll update this patch to allow running integration tests against a kinto server instance configured to send Backoff headers, for safety. Hold on merging in the meanwhile.

n1k0:
Note that this patch embeds a fix for https://github.com/mozilla-services/kinto/issues/148; We could land this as is and remove the fix when the related issue is closed, or wait for a new kinto version to be released (I really want to rely on stable releases for tests).

Natim:
r+ I agree we should not wait on mozilla-services/kinto#148 to land this PR

PR #87, opened 2015-07-21
Fixes #85 - Added integration test for deprecation headers.

### LINKED ISSUE #85: Add integration test for deprecation headers
Refs #82 which adds contextual test server configuration, so we now should add integration tests for deprecation headers.

n1k0:
r=? @Natim

### DIFF
--- a/test/integration_test.js
+++ b/test/integration_test.js
@@ -40,7 +40,7 @@ describe("Integration tests", () => {
   function flushServer(attempt=1) {
     return fetch(`${TEST_KINTO_SERVER}/__flush__`, {method: "POST"})
       .then(res => {
-        if (res.status !== 202)
+        if ([202, 410].indexOf(res.status) === -1)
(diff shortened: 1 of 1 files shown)


n1k0:
Hmm I should probably add a test for hard EOL. Hold on.

test/integration_test.js
```diff
@@ -320,4 +325,27 @@ describe("Integration tests", () => {
...
+        CLIQUET_EOS_URL: "http://www.perdu.com",
+        CLIQUET_EOS_MESSAGE: "Boom",
+      });
+    });
+
+    after(() => {
+      return stopServer();
+    });
+
+    it("should reject sync when the server sends a Backoff header", () => {
```
Natim: I think this test is not about Backoff header

n1k0:
Now with Hard EOL test added. r=? @Natim

Natim:
r+

PR #90, opened 2015-07-23
Documented current known limitations.

n1k0:
r=? @Natim

### DIFF
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -1,0 +1,66 @@
+# Current known limitations
+
+## Intended limitations
+
+### No revision tree
+
+Kinto.js doesn't track modification history in a form of a revision tree. Instead, it provides information about what eventually changed since the last time you asked, and only that.
+
+This allows covering 80% of common synchronization use cases, while being super-lightweight implementation wise.
+
+You can read more about the rationale [here](http://cliquet.readthedocs.org/en/latest/rationale.html).
+
+### No automatic confict handling
+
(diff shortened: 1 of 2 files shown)


docs/limitations.md
```diff
@@ -0,0 +1,66 @@
...
+
+As all local persistence operations are asynchronous, there might be situations where race conditions may occur.
+
+For example, if you're displaying a form for the user to update a record, and in the meanwhile that record is updated from elsewhere in the code, you might just save an obsolete/conflicting version of the record, overriding the last know valid one by "mistake".
+
+This is a [known issue](https://github.com/mozilla-services/kinto.js/issues/34) and we're in the process of finding a decent situation to handle these smoothly.
+
+### Take away →
+
+> *Pay high attention to race conditions when saving records in your app async flows.*
```
Natim: We may be explict about ServiceWorker synchronization while the APP is being used at the same time.

docs/limitations.md
```diff
@@ -0,0 +1,66 @@
...
+
+## In-memory filtering/ordering
+
+All the collection ordering and filtering is done in-memory; that means Kinto.js first loads a whole collection in memory, then process it, meaning you can easily get high-memory usage and degraded performances on very large datasets.
+
+### Take away →
+
+> *Don't use Kinto.js for very large datasets.*
```
Natim: yet? We do have plan to fix this right?

n1k0: Really depends on what you call *large datasets*; I don't have precise numbers to back that statement, but right now the design implies we load everything in memory, especially since we have adapters, which are defining a very tight shared feature scope regarding persistence.

PR #91, opened 2015-07-23
Enable CORS

ferjm:
On FxOS we enforce the same origin policy for packaged apps, so we need to enable CORS to make kinto.js work.

If you think that it's not acceptable to enable CORS by default for every request, we can make the distinction checking if we are on a FxOS device and the requester is an app: origin (although this last part is supposed to change in the future in favor of http(s) origins)

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -43,6 +43,7 @@ export default class HTTP {
     var response, status, statusText, headers;
     // Ensure default request headers are always set
     options.headers = Object.assign({}, HTTP.DEFAULT_REQUEST_HEADERS, options.headers);
+    options.mode = "cors";
     return fetch(url, options)
       .then(res => {
         response = res;


ferjm:
I've enabled CORS also on the Syncto server https://github.com/mozilla-services/syncto/pull/3

n1k0:
Sounds reasonable.

> If you think that it's not acceptable to enable CORS by default for every request

What could make it unacceptable?

ferjm:
> What could make it unacceptable?

I don't know :)

But maybe this is something that we can make optional. Just in case...

n1k0:
@ferjm Having an option sounds good indeed. Could you take care of that? I can do it if needed.

ferjm:
I can, but that'll need to wait. I am taking PTO for next week in about 10 minutes :)

n1k0:
@ferjm Lucky you :) Mine is planned next month, can't wait. In the meanwhile I'll take care of adding that option.

ferjm:
Thanks!

ametaireau:
Yes, using CORS only makes sense if we're in a browser environment (or FxOS), and doesn't in a node.js environment for instance, so having it as an option seems to make the most sense.

I believe the documentation could also mention when to disable it (I think we should have the CORS mode enabled by default to avoid surprises, or maybe we can detect if we're in the context of a browser or not)

n1k0:
Superseded by #98

PR #92, opened 2015-07-24
Local storage adapters.

n1k0:
The idea is to allow using different local persistence backends in lieu of only IndexedDB (using localStorage or creating other adapters conforming to the expected interface will then become possible).


n1k0:
r=? @Natim

.travis.yml
```diff
@@ -12,3 +12,5 @@ before_install:
 - .env/bin/pip install kinto==1.3.1
 env:
 - KINTO_PSERVE_EXECUTABLE=.env/bin/pserve
+after_script:
+- npm run report-coverage
```
Natim: :+1:

docs/extending.md
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  update(record) {
+    …
+  }
+
+  …
+}
+
+const kinto = new Kinto({adapter: MyAdapter});
```
Natim: We should say something like: "Then you can create the Kinto object passing you adapter and you are all set."

src/adapters/base.js
```diff
@@ -0,0 +1,83 @@
+"use strict";
+
+export default class BaseAdapter {
+  /**
+   * Deletes every records present in the database..
+   *
+   * @return {Promise}
+   */
+  clear() {
+    throw new Error("Implement me.");
```
Natim: nit: would raise (Not Implemented) but this might be a Python deformation.

test/adapters/IDB_test.js
```diff
@@ -0,0 +1,214 @@
...
+
+import IDB from "../../src/adapters/IDB.js";
+
+chai.use(chaiAsPromised);
+chai.should();
+chai.config.includeStack = true;
+
+const root = typeof window === "object" ? window : global;
+
+describe("adapters.IDB", () => {
```
Natim: Is it possible to make it so it is possible to reuse this test to make sure another adapter works?

Something like: https://github.com/mozilla-services/loop-server/blob/master/test/storage_test.js#L32
And: https://github.com/mozilla-services/loop-server/blob/master/test/storage_test.js#L1090-L1092

Natim:
Looks good to me with nits.

test/adapters/IDB_test.js
n1k0: I'll do that as soon as I'll add a localStorage adapter; that PR is quite large enough already :)

n1k0:
I'm gonna land this and start impelmenting a localStorage adapter.

PR #95, opened 2015-07-24
[WiP] Added LocalStorage adapter.

n1k0:
This is ugly as hell but works as a proof of concept. Will pursue next week.


src/adapters/localStorage.js
```diff
@@ -0,0 +1,136 @@
...
+  root.localStorage = require("localStorage");
+}
+
+export default class LocalStorage extends BaseAdapter {
+  constructor(dbname) {
+    super();
+    this._db = null;
+    // public properties
+    this.dbname = dbname;
+    this._keyStoreName = `${this.dbname}/__keys`;
```
ametaireau: Is this key a public one? (it's prefixed by `_`)

src/adapters/localStorage.js
```diff
@@ -0,0 +1,136 @@
...
+
+  /**
+   * Retrieve a record by its primary key from the localStorage datastore.
+   *
+   * @param  {String} id The record id.
+   * @return {Promise}
+   */
+  get(id) {
+    return Promise.resolve(
+      JSON.parse(localStorage.getItem(`${this.dbname}/${id}`)) || undefined);
```
ametaireau: Can ids be any type of value? (asking for the formatting, in order to avoid collision of names

ametaireau:
I really like to have different adapters, thanks for doing this.

The code looks good to me. r+ (with comments)

src/adapters/localStorage.js
Natim: At the moment ids can only be UUID4.

PR #97, opened 2015-07-27
Fixes #96 - Sync flow should stop when pushing changes fails.

### LINKED ISSUE #96: Sync flow should stop when pushing changes fails
Current flow is:

```js
    return this.db.getLastModified()
      .then(lastModified => this._lastModified = lastModified)
      .then(_ => this.pullChanges(result, options))
      .then(result => {
        if (!result.ok) {
          return result;
        } else {
          return this.pushChanges(result, options)
            .then(result => this.pullChanges(result, options));
        }
      });
```

We should probably resolve early when `pushChanges` result isn't `ok` as well. This is how the final flowchart should look like:

kinto js sync flow

n1k0:
r=? @ametaireau @leplatrem @Natim

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -452,12 +452,14 @@ export default class Collection {
       .then(lastModified => this._lastModified = lastModified)
       .then(_ => this.pullChanges(result, options))
       .then(result => {
-        if (!result.ok) {
+        if (!result.ok)
           return result;
-        } else {
-          return this.pushChanges(result, options)
-            .then(result => this.pullChanges(result, options));
-        }
+        return this.pushChanges(result, options)
+          .then(result => {
+            if (!result.ok)
+              return result;
+            return this.pullChanges(result, options)
+          });
       });
   }
 }
--- a/test/collection_test.js
+++ b/test/collection_test.js
@@ -884,35 +884,49 @@ describe("Collection", () => {
         .should.eventually.become(result);
     });
 
-    it("should transfer the headers option", () => {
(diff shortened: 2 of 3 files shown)


src/collection.js
```diff
@@ -456,7 +456,13 @@ export default class Collection {
           return result;
         } else {
           return this.pushChanges(result, options)
-            .then(result => this.pullChanges(result, options));
+            .then(result => {
+              if (!result.ok) {
+                return result;
+              } else {
```
leplatrem: nit: the else is superflous. Javascript idiom ?

test/collection_test.js
```diff
@@ -884,35 +884,49 @@ describe("Collection", () => {
...
-          expect(pullChanges.firstCall.args[1]).eql({headers: {Foo: "Bar"}});
-        })
+    it("should resolve early on push failure", () => {
+      const pullResult = new SyncResultObject();
+      const pushResult = new SyncResultObject();
+      pushResult.add("conflicts", [1]);
+      sandbox.stub(articles, "pullChanges").returns(Promise.resolve(pullResult));
+      sandbox.stub(articles, "pushChanges").returns(Promise.resolve(pushResult));
+      return articles.sync()
+        .should.eventually.become(pushResult);
```
leplatrem: nit: maybe we could change the test title to something like «pull is not executed if push fails». Just to help understanding this assertion :)

leplatrem:
r+

Issue #44, opened 2015-07-02
Handle synchronization when the server was flushed.


leplatrem:
The development server is flushed every day.

At first, we could think that the JS client would just reupload everything stored locally. Is it really the case?

Hints:
* If the server is flushed, the timestamps are reinitialized. 
* If the server is flushed, the permissions will be too (off topic?)

Just close this issue if cliquetis behaves as expected :)

n1k0:
TODO: I'll try flushing my local db and see what happens for the client.

Natim:
Maybe we need to detect the server flush and to ask the user for a strategy.

## How to detect the server was flushed?

It is when we get a collection timestamp updated with no records (either deleted or created)

## What to do when this occurs?

We ask the user if he wants to keep her local data or if he wants to start from scratch.

- In the first case, we mark as updated every record in the database (using resolve?)
- In the latter case we drop the local data

leplatrem:
«ask the user» -> «specify a strategy as param»

If pull changes returns 200 and updates the collection last modified, but contains no records (``data: []``), then it means the remote collection was emptied.

When this happen, we can either raise an exception (and let the dev catch it) or obtain them via the ``syncResult`` in the ``deleted`` attribute (like ``conflicts``). It shall contain the collection records marked as ``synced`` before synchronisation (?)

n1k0:
Would it be conceivable that the server would expose some header (or anything catchable) to notify it's been flushed?

Natim:
We already have a way to know that:

> If pull changes returns 200 and updates the collection last modified, but contains no records (data: []), then it means the remote collection was either flushed

Natim:
> For sync, if a node fail, the client is expected to reupload everything.

This means that with syncto the intended behavior is to reupload everything when this occurs.

Natim:
I think the reasonable behavior here is to always reupload and let the user delete and sync if it is what she wants.

n1k0:
So after having discussed this a little with @ametaireau, we think that:

- We should have available strategies defined to handle this stituation;
- On server flush detected, the default strategy would be to raise an error to properly notify the user;
- The other available strategy would be to reupload everything.

n1k0:
I think we went with raising an error so it's devs responsibility to handle the situation; is there anything missing to close this issue? @leplatrem @almet @natim @michielbdejong

leplatrem:
I'd rather leave it the way it is now and close.

n1k0:
Agreed.

PR #100, opened 2015-07-28
Added synchronization flow diagram.

n1k0:
r=? @ametaireau @Natim @leplatrem


Natim:
I don't know if it is out of this PR scope but we should add something about syncing from a flushed server. (collection timestamp mismatch with no changes)

n1k0:
There's issue #44 opened about that.

ametaireau:
I <3 diagrams. Thanks for doing this.?

leplatrem:
Neat ! 
I couldn't find any nit :) 

Maybe, you could commit the original file along (even if from proprietary software), so that anyone can contribute changes to the diagram eventually

n1k0:
> Maybe, you could commit the original file along

Updated the patch with the [draw.io](https://draw.io/) XML source.

PR #101, opened 2015-07-31
Prepare 1.0.0-rc.1

n1k0:
#### Pre PR merge

- [x] Check dist files
- [x] Check package.json info
- [x] Check outdated deps
- [x] Add `CHANGELOG.md`

#### Post PR merge

- [ ] Create tag
- [ ] Publish to npm
- [ ] Create Github release with changelog
- [ ] Publish updated demo


Natim:
We should release 1.0.0 instead of 0.6.0 because we are now supporting Kinto 1.x branch with this version of Kinto.js

Please file a 1.0.0.rc1 if you think this version is not good enough to have a 1.0.0 but I personally think that we can release 1.0.0 and release a 1.1.0 next if necessary.

Natim:
We should probably add a CHANGELOG such as https://github.com/Kinto/kinto/blob/master/CHANGELOG.rst

n1k0:
> We should release 1.0.0 instead of 0.6.0 because we are now supporting Kinto 1.x branch with this version of Kinto.js

0.6.0 is just another name for rc1 and is simpler to manage as a version name than *-rc1 with npm. Think of it as an intermediate version, I could just tag 1.0.0 tomorrow if needed.

Natim:
The important thing for me is to have a meaningful major version of the semver (http://semver.org/) schema.

n1k0:
Also, if possible (for now it is) I'd really like to stick with the [1.0.0 milestone](https://github.com/Kinto/kinto.js/milestones/1.0.0) scope for next major release.

Natim:
Yes it was the reason why I was proposing that we sail 1.0.0 for now and 1.1.0 for the next milestone?

My overall goal was to give people a meaningful version number that gave information about the supported Kinto API. (v0 === 0. major version and v1 === 1. major version) but maybe you have strong arguments against this.

In any case passing from 0.2.0 to 0.6.0 and then to 1.0.0 doesn't soons right for me. We should either go from 0.2.0 to 0.3.0 or from 0.2.0 to 1.0.0

n1k0:
I can probably investigate rc versionning using npm, though last time I checked it was a mess.

n1k0:
Okay prereleases seems to be [well supported now](https://docs.npmjs.com/misc/semver#prerelease-tags).

PR #105, opened 2015-07-31
Fixed npm published package.

n1k0:
For publishing an actually useable npm package.


Natim:
Argh should we commit them?

n1k0:
We could exclude them from the repo through `.gitignore`, and exclude sources from published npm package using `.npmignore`. That will need some build script tweaking though.

PR #106, opened 2015-08-06
[Ready for Review] Initial implementation of transformers.

n1k0:
## Transformers

Transformers are basically hooks for encoding and decoding records.

### Remote transformers

Remote transformers aim at encoding records before pushing them to the remote server, and decoding them back when pulling changes. Remote transformers are registered by calling the `Collection#use()` method, which accepts a `Kinto.transformers.RemoteTransformer`-derived object instance:

```js
import Kinto from "kinto";

function update(obj1, obj2) {
  return Object.assign({}, obj1, obj2);
}

class MyRemoteTransformer extends Kinto.transformers.RemoteTransformer {
  encode(record) {
    return update(record, {title: record.title + "!"});
  }

  decode(record) {
    return update(record, {title: record.title.slice(0, -1)});
  }
}

const kinto = new Kinto({remote: "https://my.server.tld/v1"});
coll = kinto.collection("articles");
coll.use(new MyRemoteTransformer());
```

Notice that the `decode` method should be the strict reverse version of `encode`. Calling `coll.sync()` here will store encoded records on the server; when pulling for changes, the client will decode remote data before importing them, so you're always guaranteed to have the local database containing data in clear:

```js
coll.create({title: "foo"}).then(_ => coll.sync())
// remotely saved:
// {id: "125b3bff-e80f-4823-8b8f-bfae10bfc3e8", title: "foo!"}
// locally saved:
// {id: "125b3bff-e80f-4823-8b8f-bfae10bfc3e8", title: "foo"}
```

#### Notes

> *This mechanism is especially useful for implementing a cryptographic layer, to ensure remote data are stored in a secure fashion. Kinto.js will provide one in a near future.*

### Local transformers

In a near future, Kinto.js will provide transfomers aimed at providing facilities to encode and decode records when persisted locally; you'll have to extend from `LocalTransformer` though.

### Async transformers

Transformers can also work asynchronously by returning a Promise:

```js
class MyAsyncRemoteTransformer extends Kinto.transformers.RemoteTransformer {
  encode(record) {
    return new Promise(resolve => {
      setTimeout(() => {
        resolve(update(record, {title: record.title + "!"}));
      }, 10);
    });
  }

  decode(record) {
    return new Promise(resolve => {
      setTimeout(() => {
        resolve(update(record, {title: record.title.slice(0, -1)}));
      }, 10);
    });
  }
}

coll.use(new MyAsyncRemoteTransformer());
```

### Multiple transformers

Transformers are stacked when `#use()` is called multiple times, in the order of the calls; that means you can chain multiple encoding operations, with the decoding ones being processed in the reverse order:

```js
class TitleCharTransformer extends Kinto.transformers.RemoteTransformer {
  constructor(char) {
    super();
    this.char = char;
  }

  encode(record) {
    return update(record, {title: record.title + this.char});
  }

  decode(record) {
    return update(record, {title: record.title.slice(0, -1)});
  }
}

coll.use(new TitleCharTransformer("!"));
coll.use(new TitleCharTransformer("?"));

coll.create({title: "foo"}).then(_ => coll.sync())
// remotely saved:
// {id: "125b3bff-e80f-4823-8b8f-bfae10bfc3e8", title: "foo!?"}
// locally saved:
// {id: "125b3bff-e80f-4823-8b8f-bfae10bfc3e8", title: "foo"}
```

### Limitations

There's currently no way to deal with adding tranformers to an already filled remote database; that would mean remote data migrations, and both Kinto and Kinto.js don't provide this feature just yet.

**As a rule of thumb, you should only start using transformers on an empty remote collection.**


n1k0:
This is shaping good, though I wonder how we should now deal with enabling transformers while using an unencoded remote database… Ideas?

n1k0:
So I just documented that limitation, I think it's now ready for review. r=? @Natim @leplatrem @ametaireau

n1k0:
Hmm now I think about it, and as we'll have to implement transformers for local CRUD operations as well, I wonder if the `#use()` signature isn't too limited.

Maybe we should pass transformer objects rather than encode and decode functions, so we can easily distinguish between remote transformations from local ones:

```js
class MyRemoteTransformer {
  get type() {
    return "remote";
  }
  encode(record) {
    ...
  }
  decode(record) {
    ...
  }
}

const remoteTransformer = new RemoteTransformer();
coll.use(remoteTransformer);
```

By having a type property, it's easy to determine if a transformer should be used for prepublication or as a decorator for local persistence.

Thoughts?

ametaireau:
A few things to discuss:

The function used to "install" the middleware is named `#use`, which isn't very explicit. (alternative names: `#useTransformers`, `#installTransformers`)

Passing transformers objects rather than encoders and decoders makes it clear that any encoder should have a mirror decoder in order to functionally work. Having a type attached to this object seems to make sense, because I cannot see a case where a remote transformer could be used locally and vice-versa. If such a case exists, we could overwrite the value passed by the object with the attribute of the function call.

ametaireau:
ccing @michielbdejong and @ferjm since they're working on the crypto layer for FxOS + Sync.

michielbdejong:
Ah, great! I'll try it out right now, with the code from this PR.

docs/api.md
```diff
@@ -340,3 +340,92 @@ kinto.events.on("deprecated", function(event) {
...
+function decode(record) {
+  return update(record, {title: record.title.slice(0, -1)});
+}
+
+const kinto = new Kinto({remote: "https://my.server.tld/v1"});
+coll = kinto.collection("articles");
+coll.use(encode, decode);
+```
+
+Notice that the `decode` function is the strict reverse version of `encode`. Calling `coll.sync()` here will store encoded records on the server; when pulling for changes, the client will decode remote data before importing them, so you're always guaranteed to have the local database containing data in clear:
```
leplatrem: nit: the `decode` function should be the strict reverse...

n1k0:
So I know I'm in holidays, but I wanted to push the transformer class thing while it was still fresh in my mind. From now on anybody can take over the PR, or just drop the commit if it feels wrong :)

michielbdejong:
Can you expose the `RemoteTransformer` class in the global scope or in the API somehow, so that I can include kinto.dev.js in my project and extend from it? Right now I'm getting `ReferenceError: RemoteTransformer is not defined` errors when trying out the example code (testing this in B2G-desktop).

n1k0:
> Can you expose the RemoteTransformer class in the global scope or in the API

Done, I've updated the docs accordingly.

michielbdejong:
@n1ko Awesome! :)

leplatrem:
Based on your feedback, I believe we can merge this :) I'm on it!

leplatrem:
I leave the branch in the repo since your code might depend on it.

michielbdejong:
You can safely delete the branch. Thanks for this work!

PR #108, opened 2015-08-11
Clarify docs about having to resolve incoming conflicts

leplatrem:
While working on the [kinto boiler plate app](https://github.com/Kinto/kinto-react-boilerplate), I realized that I misunderstood the meaning of ``CLIENT_WINS``. 

This PR is an attempt to clarify that **incoming** conflicts still may occur with the `CLIENT_WINS` policy.

@Natim @ametaireau @n1k0 r?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -205,9 +205,9 @@ Synopsis:
     - The developer has to handle them manually using [`#resolve()`](#resolving-conflicts), and call `#sync()` again when done;
 3. If everything went fine, publish local changes;
     - Fail on any publication conflict detected;
-        * If `strategy` is set to `Collection.strategy.SERVER_WINS`, no remote data override will be performed by the server;
-        * If `strategy` is set to `Collection.strategy.CLIENT_WINS`, conflicting server records will be overriden with local changes;
-        * If `strategy` is set to `Collection.strategy.MANUAL`, conflicts will be reported in a dedicated array.
+        * If `strategy` is set to `Collection.strategy.SERVER_WINS`, no client data will overwrite the remote data;
+        * If `strategy` is set to `Collection.strategy.CLIENT_WINS`, conflicting server records will be overwritten with local changes;
+        * If `strategy` is set to `Collection.strategy.MANUAL`, both incoming and outgoing conflicts will be reported in a dedicated array.
 
 ```js
 articles.sync()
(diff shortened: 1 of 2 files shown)


docs/api.md
```diff
@@ -217,39 +217,51 @@ articles.sync()
...
+import Collection from "kinto/lib/collection";
+
 articles.sync({
   strategy: Collection.strategy.CLIENT_WINS,
   headers: {Authorization: "Basic bWF0Og=="}
 })
-  .then(console.log.bind(console));
+  .then(result => {
+    console.log(result);
+  })
```
n1k0: Hmm, why this change?

n1k0:
Looks good though I feel like we should re-implement default behavior for automatic incoming conflict handling according to the global strategy mode used — like it was done some times ago.

That way, users selecting a given strategy would get consistent behavior all along the whole sync flow, which would make their experience more simple & more consistent. Thoughts?

docs/api.md
leplatrem: Just to explicit that the *result* we're talking about is coming here.
I had the feeling that newcomers might not get always get along with JavaScript brevity. But I might be wrong, and would be willing to revert this change :)

leplatrem:
This is exactly what was discussed in the [boilerplate PR](https://github.com/Kinto/kinto-react-boilerplate/pull/13).

Instead of changing the documentation, we might prefer to apply the policy for incoming conflicts as you said.

* ``CLIENT_WINS``: no precondition headers in PUT and incoming conflicts resolved with local version
* ``SERVER_WINS``: precondition headers and conflicts resolved with remote version

What bothers me is that the sync() should be *recursive* with above policies. Indeed, if we resolve the records implicitly, we should push them back to the server. Doing it twice might not be enough, if two clients are syncing simultaneously for example.

Thoughts ;) ?

docs/api.md
n1k0: It's mosty because the line after use `bind`. If you plan on keep this change, it's probably good to update the rest if the file as well :)

n1k0:
>  if we resolve the records implicitly

Just reading this sentence back from 2 weeks away, it's scary as hell :p FWIW, I'd like to keep users handling reconciliation in the most explicit fashion.

Natim:
Actually we should have three kind of conflict handing:

- ``CLIENT_WINS``: no precondition headers in PUT and incoming conflicts resolved with local version
- ``SERVER_WINS``: precondition headers and conflicts resolved with remote version
- ``BOTH_WINS``: Manual conflict handling (we could even try three way merge first)

docs/api.md
```diff
@@ -217,39 +217,53 @@ articles.sync()
...
-  .catch(console.error.bind(console));
+  .then(result => {
+    console.log(result);
+  })
+  .catch(error => {
+    console.error(error);
+  });
 ```
 
+The synchronization updates the local data, and provides information about performed operations.
```
ametaireau: nit: synchronisation.

ametaireau: Actually don't worry, this will be done in a later PR.

ametaireau:
LGTM.

PR #111, opened 2015-08-12
Introduce local DB prefix

leplatrem:
If my application allows to switch users (ie. logout/login), I want to isolate each local DB by user. 

The objective of this PR is to add a  prefix option for the local DB name, empty by default, that allows the developer to specify it upon its needs.

I first thought that we could imagine a better default based on hashing remote/authorization values. But I eventually preferred the default value that respect the current behaviour (ie. no prefix). Plus, blending records across servers (or users) could be a feature in some use-cases (e.g. backup, lost password, server outage, etc.)

* [x] Add prefixDB option on Collection
* [x] Pass it through Kinto constructor
* [x] Do not repeat default options from Kinto constructor
* [x] Update options docstrings


docs/api.md
```diff
@@ -12,6 +12,7 @@ const db = new Kinto(options);
 - `headers`: The default headers to pass for every HTTP request performed to the Cliquet server (eg. `{"Authorization": "Basic bWF0Og=="}`);
 - `adapter`: The persistence layer adapter to use for saving data locally (default: `Kinto.adapters.IDB`); alternatively, a `Kinto.adapters.LocalStorage` adapter is also provided; last, if you plan on writing your own adapter, you can read more about how to do so in the [Extending Kinto.js](extending.md) section.
 - `requestMode`: The HTTP [CORS](https://fetch.spec.whatwg.org/#concept-request-mode) mode. Default: `cors`.
+- `prefixDB`: The prefix for the local DB name (default: `""`). Use this option so that your application can switch remote server and authorization without combining records.
```
n1k0: How about `dbPrefix` instead?

n1k0: Also, the sentence is hard to understand from a user perspective. Maybe *Use this option to isolate user specific databases*?

src/api.js
```diff
@@ -40,7 +40,7 @@ export default class Api {
     this._backoffReleaseTime = null;
     // public properties
     this.remote = remote;
-    this.optionHeaders = options.headers;
+    this.optionHeaders = options.headers || {};
```
n1k0: Hmm, you don't like default option values? :)

test/index_test.js
```diff
@@ -78,6 +78,11 @@ describe("Kinto", () => {
       expect(new Kinto({requestMode}).collection("x").api.http.requestMode)
         .eql(requestMode);
     });
+
+    it("should propagate the prefixDB option to child dependencies", () => {
+      expect(new Kinto({prefixDB: "app--"}).collection("x").db.dbname)
+        .eql("app--default/x");
```
n1k0: Makes me think we should probably pay attention to possible db/store naming constraints in different adapaters (IDB/localStorage)… Maybe we should provide some validation for these?

n1k0:
Looks good minus the minor remark regarding db/collection naming constraints. This probably needs checking, and you ca always spawn a new issue to address it if it happens to be a problem. r+wc

PR #118, opened 2015-08-14
Remove mentions of Cliquet

leplatrem:
**Wait for https://github.com/Kinto/kinto/pull/156** (or some links will be broken)

Since *Cliquet* is only a building block of the Kinto server, it should not be mentionned at all in Kinto.js docs.
We noticed that this introduces a huge confusion between Cliquet and Kinto, and we'll be working on that clarification.

A first step is to remove every explicit mention of Cliquet in Kinto.js docs.

### DIFF
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -59,7 +59,7 @@ For now only a few standard fields are indexed by default in IndexedDB collectio
 Here's what's planned for future versions, outside of fixing the known limitations listed above:
 
 - Adding a **client-side crypto layer** to the API in order to bring secure & privacy-safe remote storage of user data;
-- Adding support for [**sharing & permissions**](http://kinto.readthedocs.org/en/latest/permissions.html);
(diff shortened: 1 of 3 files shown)


docs/limitations.md
```diff
@@ -59,7 +59,7 @@ For now only a few standard fields are indexed by default in IndexedDB collectio
 Here's what's planned for future versions, outside of fixing the known limitations listed above:
 
 - Adding a **client-side crypto layer** to the API in order to bring secure & privacy-safe remote storage of user data;
-- Adding support for [**sharing & permissions**](http://cliquet.readthedocs.org/en/latest/reference/permission.html);
+- Adding support for [**sharing & permissions**](http://kinto.readthedocs.org/en/latest/permissions.html);
```
Natim: http://kinto.readthedocs.org/en/latest/concepts.html#inherited

PR #128, opened 2015-08-26
Hint at http error reporting in docs example, fix #124

### LINKED ISSUE #124: Report http errors in a meaningful way
There [are events for](http://kintojs.readthedocs.org/en/latest/api/#events) backoff and deprecated, but what about network timeouts, 401 Unauthorized errors, and 500 Internal server errors from the Kinto servers? Now, these seem to result in a promise rejection by #sync(), but it would be nice to have more meaningful events to handle. For instance, in case of a 401, it might be necessary to obtain a new BrowserID assertion, which the app needs to handle.

michielbdejong:
This should be enough to point developers in the right direction, a full 'Handling sync rejections' section (listing all possible errors sync can reject) can always be added later.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -243,7 +243,11 @@ Synopsis:
 ```js
 articles.sync()
   .then(console.log.bind(console))
-  .catch(console.error.bind(console));
+  .catch(err => {
+    if (err.data === 401) {
+      console.error('HTTP status code indicates auth problem');
+    }
+  });
 ```
 
 ### Synchronization strategies
--- a/docs/api.md
+++ b/docs/api.md
@@ -244,11 +244,19 @@ Synopsis:
 articles.sync()
   .then(console.log.bind(console))
   .catch(err => {
-    if (err.data === 401) {
+    if (err.response && err.response.status === 401) {
       console.error('HTTP status code indicates auth problem');
     }
   });
 ```
+
+### Error handling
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -243,7 +243,11 @@ Synopsis:
 ```js
 articles.sync()
   .then(console.log.bind(console))
+  .catch(err => {
+    if (err.data === 401) {
```
n1k0: You can access the response through `err.response`; you should be able to get the status through `response.status`, which is probably clearer & more explicit than `data`, which is the json response body attached by to the response itself.

n1k0:
Also and while you're at it, could you please add a small general section about HTTP error handling? So we could properly document the `response` and `data` props attached to the error object.

michielbdejong:
Done! I did note that connecting to a non-existing server (I tried it with http://example.com:24012) takes as much as 75 seconds to time out in Firefox desktop, is that by design?

n1k0:
> Done!

Perfect, r+

> I did note that connecting to a non-existing server (I tried it with http://example.com:24012) takes as much as 75 seconds to time out in Firefox desktop, is that by design?

This might be handled that way by [isomorphic-fetch](https://www.npmjs.com/package/isomorphic-fetch)… That's probably worth filing a dedicated issue.

PR #129, opened 2015-08-26
Fix db transactions on safari.

QuentinRoy:
On Safari 8.0.8 (and priors, most likely), calling IDBDatabase.transaction with mode == undefined raises a TypeError (the browser won't use default mode value).
As a workaround, if mode is undefined, create the transaction without the mode argument.


n1k0:
Thanks. I'm suspicious about the amount of changes for dist files:

screen shot 2015-08-26 at 16 15 16

How did you generate them?

QuentinRoy:
Yes, I was surprised too.
I just run `npm build` and `npm dist`.

QuentinRoy:
I believe it is because of changes in the compilers. Probably babel. If we want to avoid that, we may consider removing the `^` before the dependencies' versions in `package.json` so that the exact same dependencies are installed instead of the most recent compatible ones.

n1k0:
> I believe it is caused by changes in the compilers

Could you please ensure that's the case, eg. sequentially using two distinct versions? I'm super surprised a new compiler version suddenly adds +50% generated code to the builds…

QuentinRoy:
OK it is not be because of the compilers. Here is what I investigated:
* I tried to use the exact versions provided in the package.json and it gave me the same compiled file than with the more recent versions.
* I checked out the previous commit (c5f55f0) and compiled (using once again the compiler versions of package.son) and it gave me a very different result from what was in the commit.

I have no explanation.

n1k0:
Superseded by #142.

PR #130, opened 2015-08-26
Add instructions on how to run functional tests.

ametaireau:
This helps to setup the environment to run the tests. It seems that node isn't able to find the pserve that's in the path when run with some specific shells.

cc @quentinroy

PR #132, opened 2015-08-26
Fix test failing when KINTO_PSERVE_EXECUTABLE is not set in the env.

QuentinRoy:
During the integration test, kinto was spawn with a virgin environment (or an environment containing only the provided variables).
While this ensured that any tester where given the same 'pure' kinto instance, it also (a) prevented them to test it in a custom configuration, (b) force them to set the `KINTO_PSERVE_EXECUTABLE` env variable as `pserve` could not be found anymore using the `PATH` environment variable.

This fix makes sure the provided env argument extends the parent environment instead of overwriting it.

This is linked to #130 and raised some discussion with @ametaireau. Keeping a virgin environment for Kinto may be wanted. In particular, the environment is used by the tests to communicate with the Kinto instance and, as a result, will raise uncontrolled behaviour if a user sets by himself some environment variables such as `CLIQUET_EOS` or `CLIQUET_BACKOFF`.
Another working solution can be to only relay the `PATH` of parent's env (which is required to find the `pserve` executable).

### DIFF
--- a/test/integration_test.js
+++ b/test/integration_test.js
@@ -22,7 +22,11 @@ describe("Integration tests", () => {
   var sandbox, server, tasks;
   const MAX_ATTEMPTS = 50;
 
-  function startServer(env={}) {
+  function startServer(env) {
+    // Add the provided environment variables to the child process environment.
+    // Keeping parent's environment is needed so that pserve's executable 
+    // can be found (with PATH) if KINTO_PSERVE_EXECUTABLE env variable was not provided.
+    env = Object.assign({}, process.env, env);
     server = spawn(PSERVE_EXECUTABLE, [KINTO_CONFIG], {env});
     server.stderr.on("data", function(data) {
       // Uncomment the line below to have server logs printed.


test/integration_test.js
```diff
@@ -22,7 +22,11 @@ describe("Integration tests", () => {
   var sandbox, server, tasks;
   const MAX_ATTEMPTS = 50;
 
-  function startServer(env={}) {
+  function startServer(env) {
+    // Add the provided environment variables to the child process environment.
+    // Keeping parent's environment is needed so that pserve's executable 
+    // can be found (with PATH) if KINTO_PSERVE_EXECUTABLE env variable was not provided.
```
n1k0: I'm reluctant landing this change as we lose control on what executable is actually used (imagine a globally installed `pserve` in a different version…)

I'd rather keep using an explicitly passed executable path, with proper documentation added (which is coming along #130 by @ametaireau).

Thoughts?

test/integration_test.js
```diff
@@ -22,7 +22,11 @@ describe("Integration tests", () => {
   var sandbox, server, tasks;
   const MAX_ATTEMPTS = 50;
 
-  function startServer(env={}) {
+  function startServer(env) {
```
n1k0: Nit: you can keep the default value here.

test/integration_test.js
Natim: I like the idea that having the VIRTUALENV activated automatically detect the right pserve command.
I would rather fail if the pserve command is wrong and doesn't find the server dependencies / is not able to run the server.

ametaireau: Well, it's actually useful if you *activate* your venv. Otherwise the pserve executable needs to be explicitely called out to be found.

test/integration_test.js
QuentinRoy: If we do Object.assign later, it is actually useless.

n1k0: What if I call `startServer([])`? Defaulting a required object to an object sounds fair to me.

QuentinRoy: You mean it gives users insight on what is intended?

n1k0: Just realized that `Object.assign({}, undefined)` doesn't raise any error — which is a little weird but, oh well. So scratch that comment we're *safe* here.

n1k0:
As commented on #130:

> I don't want to block on this too hard, so I'll just merge #132 while people seems happy with it. Thus 
> documentation should be updated to reflect the new situation, which is to ensure the `pserve` 
> executable extracted from `PATH` is the right one, with the option to override it defining 
> `KINTO_PSERVE_EXECUTABLE` as commented above.

So let me land this, #130 should see the documentation updated accordingly.

PR #139, opened 2015-08-31
Fixes #122 - Added Kinto.createRemoteTransformer().

### LINKED ISSUE #122: Using Transformers with prototypal inheritance
The `class` and `extends` keywords only work in ES6, and it would be nice if developers can define their Transformer using just ES5. I *think* this code is equivalent to the ES6 code in http://kintojs.readthedocs.org/en/latest/api/#transformers and that we should add it to the documentation?

In ES5:
````js
function MyRemoteTransformer() {};

MyRemoteTransformer.prototype = Kinto.transformers.RemoteTransformer.prototype;
MyRemoteTransformer.prototype.encode = function(record) {
  record.title += "!";
  return record;
};

MyRemoteTransformer.prototype.decode = function(record) {
  record.title = record.title.slice(0, -1);
  return record;
}

const kinto = new Kinto({remote: "https://my.server.tld/v1"});
coll = kinto.collection("articles");
coll.use(new MyRemoteTransformer());
````

In ES6:
````js
import Kinto from "kinto";

function update(obj1, obj2) {
  return Object.assign({}, obj1, obj2);
}

class MyRemoteTransformer extends Kinto.transformers.RemoteTransformer {
  encode(record) {
    return update(record, {title: record.title + "!"});
  }

  decode(record) {
    return update(record, {title: record.title.slice(0, -1)});
  }
}

const kinto = new Kinto({remote: "https://my.server.tld/v1"});
coll = kinto.collection("articles");
coll.use(new MyRemoteTransformer());
````

### PR DESCRIPTION
**Pull request by n1k0**
refs #122

### Creating transformers in ES5

If your JavaScript environment doesn't suppport [ES6 classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes) just yet, you can derive transformers in an ES5 fashion using `Kinto.createRemoteTransformer` static helper:

```js
const TitleCharTransformer = Kinto.createRemoteTransformer({
  constructor: function(char) {
    this.char = char;
  },

  encode: function(record) {
    return update(record, {title: record.title + this.char});
  },

  decode: function(record) {
    return update(record, {title: record.title.slice(0, -1)});
  }
});

coll.use(new TitleCharTransformer("!"));
coll.use(new TitleCharTransformer("?"));
```

> #### Notes
>
> Notice you can define a `constructor` method, though you don't need (and must not) call `super()` within it, unlike with ES6 classes.


n1k0:
r=? @michielbdejong

src/index.js
```diff
@@ -44,6 +44,27 @@ export default class Kinto {
   }
 
   /**
+   * Creates a remote transformer constructor, the ES5 way.
+   *
+   * @return {RemoteTransformer}
+   */
+  static createRemoteTransformer(proto) {
+    if (!proto || typeof proto !== "object")
+      throw new Error("Expected prototype object.");
```
ametaireau: I though it was a good practice to always enclose iffed instructions inside a block `{` and `}`.

ametaireau:
Looks good to me (with nitpick)

src/index.js
n1k0: Depends on your coding style; Gecko (and Mozilla)'s one states that for single line statements braces can be omitted. If we're planning on removing this rule, we'll need to update a bunch of files elsewhere; if that's the case, it's probably safe to spawn another issue.

n1k0:
Waiting for @michielbdejong final r+, ideally with checking that everything works fine on his end :)

src/index.js
ametaireau: They actually state that you should always use them:

> Always brace controlled statements, even a single-line consequent of an if else else. This is redundant typically, but it avoids dangling else bugs, so it's safer at scale than fine-tuning.
> — [Developer Coding Style guide](https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style#Control_Structures)

n1k0: Bah, this must have changed recently, I remember we conformed to it for
Loop.

Still, could you please find a distinct issue about this?

Nicolas Perriault
Le 31 août 2015 7:31 PM, "Alexis Metaireau" <notifications@github.com> a
écrit :

n1k0: s/find/file (cannot edit comment from mobile)

michielbdejong:
Will check today!

michielbdejong:
The `encode` and `decode` methods of my transformer return a promise (can't do crypto stuff synchronously of course), and this seems to work. But the example transformer returns a value synchronously. Do both these approaches work? If so, maybe worth a note in the docs about that?

n1k0:
> Do both these approaches work? 

Yes.

> If so, maybe worth a note in the docs about that?

Well, both approaches are documented:

- sync http://kintojs.readthedocs.org/en/latest/api/#remote-transformers
- async http://kintojs.readthedocs.org/en/latest/api/#async-transformers

n1k0:
I can probably add a specific note about the accepted return types though, will update.

michielbdejong:
Oh, great, I hadn't scrolled down to the async section, but now I do remember having seen it a while ago. Thanks a lot! r+

PR #143, opened 2015-09-03
Fixes #138 - Implement custom id schema's

### LINKED ISSUE #138: Pluggable ID schema's
Normally, Kinto.js will use UUID strings for item ID's, but there are situations where you want to control how these are generated. For instance, if you are importing an existing data set with integer ID's, or some other ID schema. The idea would be something like (just a draft idea, can be improved):

````js
var MyIdSchema = createIdSchema({
  generate: function() {},
  validate: function(id) {}
});
kinto.collection('foo').use(MyIdSchema);
````

Here, `generate` would return a string, and `validate` would return a boolean.
I'll see if I can implement this in a PR.

### PR DESCRIPTION
**Pull request by michielbdejong**
There are still variable names like `options.forceUUID` which refer to the default IdSchema, not sure if that's worth an API change?

I did not run the integration test locally (I don't have my environment set up correctly yet for that), but I added some unit tests, and unit tests all seem to pass.

The approach I took was:
* Move `src/transformers/` to `src/plugins/`.
* Move the UUID-related code into `src/plugins/IdSchema.js`.
* The collection creates an instance of that in `this._idSchema` in the collection constructor. 
* The IdSchema class is exposed as `Kinto.IdSchema` for people to derive their custom id schema from.
* I also added `Kinto.createIdSchema` in parallel with `Kinto.createRemoteTransformer`.
* I added a section to `doc/api.md`, let me know if it's clear
* I didn't add much in the way of code comments, I think it's pretty clear, but let me know if you want more comments / docs / unit tests / etc.

In any case, I thought maybe you can give me some early feedback and let me know if you like the API, what you think of variable naming, and what I should improve?

Cheers!
Michiel.

### CODE REVIEW

File: src/collection.js [hunk 1]
```diff
@@ -104,8 +106,14 @@ export default class Collection {
     });
   }
 
-  use(transfomer) {
-    this._transformers[transfomer.type].push(transfomer);
+  use(plugin) {
```
n1k0: I kinda like `plugin` :) Thought about `processor`, but `plugin` sounds even more generic.

src/collection.js
```diff
@@ -104,8 +106,14 @@ export default class Collection {
...
+    if (plugin.type == "idschema") {
+      this._idSchema = plugin;
+    } else if (plugin.type == "remotetransformer") {
+      this._transformers.remote.push(plugin);
```
n1k0: How about storing all this in `_plugins`?

n1k0:
> There are still variable names like options.forceUUID which refer to the default IdSchema, not sure if that's worth an API change?

It definitely needs at least a name change, eg. `forceId`.

> I did not run the integration test locally (I don't have my environment set up correctly yet for that), but I added some unit tests, and unit tests all seem to pass.

The good news is, integration tests pass just fine (as you can see in Travis anyway) :)

> - The collection creates an instance of that in this._idSchema in the collection constructor. 

Could we imagine having a single `_plugins` property containing both the IdSchema and transformers?

> - I added a section to doc/api.md, let me know if it's clear

Super clear! All in all this is shaping super good. Thanks again.

michielbdejong:
> [`options.forceUUID`] definitely needs at least a name change, eg. `forceId`.

OK! And then I'll also change the mentions of UUID to just 'id' in error messages. So technically, these are breaking changes, which implies a bump in the version string in package.json, correct?

docs/api.md
```diff
@@ -526,3 +527,66 @@ coll.use(new TitleCharTransformer("?"));
...
+class IntegerIdSchema extends IdSchema {
+  constructor() {
+    super();
+    this._next = 0;
+  }
+  generate() {
+    return this._next++;
+  }
+  validate(id) {
+    return ((id == parseInt(id, 10)) && (id >= 0));
```
Natim: I wonder if we should show an  example that doesn't handle unicity, maybe it worth adding a note or something.

michielbdejong:
> Could we imagine having a single _plugins property containing both the IdSchema and transformers?

OK, I'll move `this._idSchema` to `this._plugins.idSchema` and the `this._transformers.remote` array to `this._plugins.remoteTransformers` in src/collection.js, and then local transformers can go into a `this._plugins.localTransformers` array

docs/api.md
```diff
@@ -526,3 +527,66 @@ coll.use(new TitleCharTransformer("?"));
...
+coll.use(new IntegerIdSchema());
+```
+
+> #### Notes
+>
+> Notice you can define a `constructor` method, though you don't need (and must not) call `super()` within it, unlike with ES6 classes.
+
+### Limitations
+
+There's currently no way to deal with adding a custom id schema to an already filled remote database; that would mean remote data migrations, and both Kinto and Kinto.js don't provide this feature just yet.
```
Natim: > There's currently no way to deal with adding a custom id schema to an already filled remote database

Isn't it what we are doing with Syncto? I think we should specify that there are no way to switch from UUID4 to custom id schema because if your remote database already respect the schema there are no limitation.

docs/api.md
michielbdejong: So you mean just `return (typeof id == 'number');`? Sure, that's simpler to read, yet still shows how the validate function should work.

docs/api.md
michielbdejong: Right! True. The only thing to avoid is adding an id schema which the /local/ data does not adhere to. Will correct that.

src/collection.js
```diff
@@ -157,10 +165,10 @@ export default class Collection {
       return Promise.reject(new Error("Record is not an object."));
     }
     const newRecord = Object.assign({}, record, {
-      id:      options.synced || options.forceUUID ? record.id : uuid4(),
+      id:      options.synced || options.forceUUID ? record.id : this._idSchema.generate(),
```
Natim: I am not sure I understand this condition. ``if forceUUID then record.id`` I though forceUUID was to force an UUID generation.

Natim:
This looks good man :)

Natim:
> technically, these are breaking changes, which implies a bump in the version string in package.json, correct?

I would go for it :+1:

docs/api.md
Natim: Yes I like that, plus a note saying that:

    In a real application, you want to make sure you do not generate twice the same record id on a collection.
    This dummy example doesn't take care of id unicity. In case of id conflict you may loose data.

docs/api.md
Natim: Thanks.

n1k0:
> So technically, these are breaking changes, which implies a bump in the version string in package.json, correct?

Yes, rc4 it is. I'll package and publish it as soon as this is merged. 

> OK, I'll move this._idSchema to this._plugins.idSchema and the this._transformers.remote array to this._plugins.remoteTransformers in src/collection.js, and then local transformers can go into a this._plugins.localTransformers array

I'm wondering about `_plugins.transformers.local` and `_plugins.transformers.remote`… Yeah, that's a little long. Your call.

src/collection.js
michielbdejong: The way I understand it, getting assigned a new random UUID is the default, and the optional forceUUID parameter is so you can force your own /specific/ UUID.

n1k0: That's right.

docs/api.md
```diff
@@ -526,3 +527,66 @@ coll.use(new TitleCharTransformer("?"));
 There's currently no way to deal with adding tranformers to an already filled remote database; that would mean remote data migrations, and both Kinto and Kinto.js don't provide this feature just yet.
 
 **As a rule of thumb, you should only start using transformers on an empty remote collection.**
+
+## Id schemas
+
+By default, kinto.js uses UUID4 strings for record id's. If you want to work with an existing body of data, this may not be what you want.
+
+You can define a custom id schema on a collection by calling the `Collection#use()` method, which accepts a `Kinto.IdSchema`-derived object instance:
```
ametaireau: I feel having one `use` API which accepts different types of input is misleading. I've seen this pattern used for express (the node.js framework) in the past and it seems to be counter intuitive (e.g. really easy to use, not as easy to read).

Instead, we could have different methods to *install* the different configurations.

Here are different proposals:
- Have a `Collection#configure` to which we would pass an object with an `id-schema` key;
- Variant, pass two arguments to `Collection#setConfig`. e.g. `setConfig('id-schema', object)`
- Have a `Collection#setIdSchema` to which we pass an object;

Thoughts?

docs/api.md
ametaireau: Yes, I like that because it makes the user learn about common issues when reading the docs.

src/collection.js
ametaireau: It seems that this brings a situation where different extensibility mechanisms (e.g. plugins, adapters, transformers) overlaps, without a clear distinction in their names.

Until now, we had a clear scoping with each of the concepts, if we introduce "plugins", I fear that we might lose this. Should we consider `generators` (like we did for Kinto server) ?

src/collection.js
ametaireau: If this is uncealr from the code, we should add a comment.

src/collection.js
ametaireau: Oh, sorry I misunderstood this! We have different kinds of plugins (adapters, transformers, etc). Ignore these comments.

ametaireau:
Thanks for working on this @michielbdejong.

michielbdejong:
I like the first option! So that would mean, change `use` to be called `configure` and to take an object with keys instead of looking at the prototypes like it does now. So then instead of:

````js
collection.use(myIdSchema);
collecition.use(myRemoteTransformer);
````
it would become:
````js
collection.configure({
  idSchema: myIdSchema,
  remoteTransformers: [ myRemoteTransformer ]
});
````
Looks much more readable to me, indeed. Since this PR is going to introduce (small) breaking API changes anyway, like changes in error message strings, and already changes `use`, shall I just include this change as well?

**Comment by n1k0** [comment]
> Looks much more readable to me, indeed

Definitely, let's do this!

> shall I just include this change as well?

Hmm that's gonna be a rather big patch :) If you feel like it, do it — but please try to split the different facets of the implementation in distinct commits, for easier reviews.

**Comment by n1k0** [comment]
Of course we could also iterate here; landing this patch first, filing a new issue for the `Collection#configure` thing which could be addressed as soon as this lands.

**Comment by michielbdejong** [comment]
OK, created #148 for this change, will submit a separate PR for that, and then rebase this one on that.

File: src/collection.js (hunk 2 above)
Comment (michielbdejong) [inline src/collection.js]: Actually, now that we have `Collection#configure(config)`, maybe we should just store that exact object as `this._config`? Then we have `this._config.idSchema` and `this._config.remoteTransformers`. Or if you want to stick to 'plugins then we can call the configure method `Collection#configurePlugins(plugins)`.

**Comment by michielbdejong** [comment]
Rebased this branch on #148 and applied the following review comments:
  * name change forceUUID -> forceId, etc.
  * add comment about what forceId does
  * simplify example (just check for typeof 'number'), and add note about unicity
  * Correct phrase:
    "There's currently no way to deal with adding a custom id schema to an already filled remote database"

Did not yet apply the review comment about moving `this._idSchema` and `this._transformers.remote` into one data structure, because I'm not sure whether to call it `this._plugins` (and then maybe rename `Collection#configure` to `Collection#configurePlugins`), or calling it `this._config` so it's more logical in combination with the naming of the `Collection#configure` method.

**Comment by n1k0** [comment]
> or calling it `this._config` so it's more logical in combination with the naming of the `Collection#configure` method

That would make a lot of sense to me. `_config` sounds good.

**Comment by n1k0** [comment]
> That would make a lot of sense to me. `_config` sounds good.

BTW, how about supporting passing th options object as an arg to `Kinto#collection`? Eg.

```js
const kinto = new Kinto();
const collection = kinto.collection("foo", {
  idSchema: myIdSchema,
  remoteTransformers: [ myRemoteTransformer ]
});
```

michielbdejong:
Will improve test coverage, particularly for `Kinto.createIdSchema`, `Kinto.IdSchema`, and `kinto.collection("articles", { idSchema: myIdSchema });`.

michielbdejong:
Need to fix how UuidSchema inherits from IdSchema, and check/improve test coverage.

src/collection.js
```diff
@@ -1,13 +1,14 @@
...
 import { EventEmitter } from "events";
-import { v4 as uuid4 } from "uuid";
 import deepEquals from "deep-eql";
 
 import BaseAdapter from "./adapters/base";
-import { attachFakeIDBSymbolsTo, reduceRecords, isUUID4, waterfall } from "./utils";
+import { attachFakeIDBSymbolsTo, reduceRecords, waterfall } from "./utils";
 import { cleanRecord } from "./api";
 
+import UuidSchema from "./schemas/uuidschema";
```
n1k0: Nit: `UUIDSchema`?

src/collection.js
```diff
@@ -158,24 +160,26 @@ export default class Collection {
    * Adds a record to the local database.
    *
    * Options:
-   * - {Boolean} synced     Sets record status to "synced" (default: false).
-   * - {Boolean} forceUUID  Enforces record creation using any provided UUID
-   *                        (default: false).
+   * - {Boolean} synced   Sets record status to "synced" (default: false).
+   * - {Boolean} forceId  Forces the id field from the record to be used, instead
+   *                      of one that is generated automatically (default: false).
```
n1k0: Makes me think that for creations, we could raise if `forceId` is `false` and the method receives a record object containing an id; right now the error is silenced.

src/collection.js
```diff
@@ -217,15 +221,30 @@ export default class Collection {
...
+   *
+   * @param  {Object} conflict   The conflict object.
+   * @param  {Object} resolution The proposed record.
+   * @return {Promise}
+   */
+  resolve(conflict, resolution) {
+    return this.update(Object.assign({}, resolution, {
+      last_modified: conflict.remote.last_modified
+    }));
+  }
```
n1k0: For some reason you now have two versions of this method in the prototype. I recently moved it next to `_handleConflicts`, which should stay there.

n1k0:
Looks awesome to me; r=r+wc from me, though I'd like to hear from @Natim as well :)

src/collection.js
michielbdejong: Ah, merge mistake, sorry! Thanks for spotting it.

src/collection.js
michielbdejong: OK, I started doing this but it requires some code restructuring and extra tests, so I'll make it a follow-up ticket if that's OK (it's a good idea anyway to check for this, regardless of whether you use the default or a custom IdSchema).

n1k0: Yeah it's always fine and healthy to spawn new issues for not-directly related stuff.

n1k0:
Note that CI reported an error: `Error: Cannot find module './schemas/UUIDSchema'`; I think you need to reference the filename using the exact same case, here `uuidschema`. Or you could rename the file as `UUIDSchema.js`, your call.

docs/api.md
```diff
@@ -558,3 +559,70 @@ coll = kinto.collection("articles", {
 There's currently no way to deal with adding transformers to an already filled remote database; that would mean remote data migrations, and both Kinto and Kinto.js don't provide this feature just yet.
 
 **As a rule of thumb, you should only start using transformers on an empty remote collection.**
+
+## Id schemas
```
ametaireau: nit: use "ID", as id is used to refer to [something completely different](https://en.wikipedia.org/wiki/Id,_ego_and_super-ego#Id).

src/collection.js
```diff
@@ -158,24 +160,26 @@ export default class Collection {
    * Adds a record to the local database.
    *
    * Options:
-   * - {Boolean} synced     Sets record status to "synced" (default: false).
-   * - {Boolean} forceUUID  Enforces record creation using any provided UUID
-   *                        (default: false).
+   * - {Boolean} synced   Sets record status to "synced" (default: false).
+   * - {Boolean} forceId  Forces the id field from the record to be used, instead
```
ametaireau: `forceID` doesn't seem self-explanatory to me (without reading the documentation about it, I have no idea what this option does). Should we rename it to something different? `useRecordID`?

n1k0: I'm personally fine with `forceID` as it's properly documented, but `useRecordID` would work too. Delegating decision making to @michielbdejong.

n1k0:
Coverage being [okay on coveralls](https://coveralls.io/builds/3574224), I think we can now safely land this. @michielbdejong please do :)

n1k0:
Just take care of squashing the commits first.

michielbdejong:
> change the api.md document in other places where "id" is supposed to be "ID"

Ah sorry, hadn't searched properly. I did not change the ones in code comments in `src/`, since they're very close to the lower-case variable name there, and I thought changing phrases like "the id field" to "the ID field" would be confusing. But I now changed all the ones in api.md outside code snippets.

Thanks for your reviews!

PR #86, opened 2015-09-04
Remove the Backoff header handling exclusion for 304 responses 

Natim:
Refs #68, #82; for now backoff header handling is not performed when we receive 304 responses, because of https://github.com/mozilla-services/kinto/issues/148.

Once this issue is fixed and a new Kinto release is issues, we should remove that exclusion.

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -106,11 +106,6 @@ export default class HTTP {
   }
 
   _checkForBackoffHeader(status, headers) {
-    // XXX Temporary fix
-    // see https://github.com/mozilla-services/kinto/issues/148
-    if (status === 304) {
-      return;
-    }
     var backoffMs;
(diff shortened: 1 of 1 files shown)


ametaireau:
LGTM.

n1k0:
Hmm this has been merged a little quickly; is the fix available along Kinto 1.3 which is the version of the server we use in integration tests?

Natim:
Test are passing, so I guess yes.

n1k0:
I mean, we should have written a specific test along this patch.

Natim:
Do you want me to write one? I though test where already handling it and this was made in order for tests to pass.

n1k0:
We should upgrade the Kinto server version we use for integration testing, and indeed write a test to ensure the issue is no more using that new version :)

PR #146, opened 2015-09-04
Fixes #125, fixes #110: Drop reliance on Content-Length header.

### LINKED ISSUE #110: Error when server returns 400 error response
``Cannot read property 'errno' of null``

**To reproduce**:

```js
var collection = kinto.collection("ZjE1ZjgxODMtOTcyNy00ZDgzLWE2MDEtOTUyMTFlYTIzMzQ5OnMzY3IzdA==-items");
collection.sync();
```

Server will respond:
```
GET https://kinto.dev.mozaws.net/v1/buckets/default/collections/ZjE1ZjgxODMtOTcyNy00ZDgzLWE2MDEtOTUyMTFlYTIzMzQ5OnMzY3IzdA==-items/records

Access-Control-Allow-Origin:*
Connection:keep-alive
Content-Length:167
Content-Type:application/json; charset=UTF-8
Date:Wed, 12 Aug 2015 07:24:54 GMT
Server:nginx/1.4.6 (Ubuntu)

{"errno":107,"message":"path: Invalid record id","code":400,"details":[{"location":"path","name":null,"description":"Invalid record id"}],"error":"Invalid parameters"}
```

Curiously, it looks like in kinto.js code ``response.json()`` is `null`.

### LINKED ISSUE #125: Why is Content-Length header required on Kinto server responses?
In https://github.com/Kinto/kinto.js/blob/1deefd244da694b14320c4094c4248e5cdd26be9/test/http_test.js#L109 you test that requests fail if there is no Content-Length header in the http response.

The default for a nodejs server is (I think, or it seems) to respond without a Content-Length header, because this allows streamed response bodies (http/1.1 chunked encoding). I did not find any mention of this in the Kinto/Cliquet docs. Is this by design?

I can work around this by manually adding a Content-Length header to the mock server I'm using, but I thought it was worth a ticket, because the kinto.js code (and its [unit test](https://github.com/Kinto/kinto.js/blob/1deefd244da694b14320c4094c4248e5cdd26be9/test/http_test.js#L109)) seem to contradict the Kinto/Cliquet API documentation at this point.

n1k0:
This fixes issues #125 and #110.

r=? @natim @michielbdejong


Natim:
Looks good to me thanks.

michielbdejong:
LGTM, thanks!

So I'm learning from this that `res.json()` on the fetch API is not super useful unless you're absolutely sure that you are going to have some JSON coming in. I agree that consuming the response as text first, and then JSON-parsing it later, is necessary here, because the body can only be consumed once.

michielbdejong:
This comment is just for future reference:

I wrote:
> So I'm learning from this that res.json() on the fetch API is not super useful

But there is a nicer way described in [Jake's post](https://jakearchibald.com/2015/thats-so-fetch/) (scroll to the 'Stream readers & cloning' section), using stream cloning:

````js
fetch(url).then(function(response) {
  return response.clone().json().catch(function() {
    return response.text();
  });
});
````
We can leave the code as it is now, but wanted to complete with this info in case I or someone else ever reads this back. :)

PR #149, opened 2015-09-07
Fixes #148 - Extra arg on Kinto#collection to replace Collection#use.

### LINKED ISSUE #148: Move `Collection#use` to optional second argument of `Kinto#collection`
As discussed in #143, change `Collection#use` to be called `Collection#configure` and to take an object with keys instead of looking at the prototypes like it does now. So then instead of:

````js
collection.use(myIdSchema);
collection.use(myRemoteTransformer);
````

it would become:

````js
collection.configure({
  idSchema: myIdSchema,
  remoteTransformers: [ myRemoteTransformer ]
});
````

### PR DESCRIPTION
**Pull request by michielbdejong**

### CODE REVIEW

**Comment by michielbdejong** [comment]
@n1k0 maybe you prefer:
````js
coll.configure({
  transformers: {
    remote: [ myRemoteTransformer ]
  }
});
````
instead of:
````js
coll.configure({
  remoteTransformers: [ myRemoteTransformer ]
});
````

**Comment by n1k0** [comment]
The second version is fine, the first is quite verbose and not much more explicit.

File: src/collection.js [hunk 1]
```diff
@@ -104,8 +104,13 @@ export default class Collection {
...
 
-  use(transfomer) {
-    this._transformers[transfomer.type].push(transfomer);
+  configure(config) {
+    if (typeof config !== "object") {
+      throw new Error("First argument of configure should be an object.");
+    }
+    if (config.remoteTransformers instanceof Array) {
+      this._transformers.remote = config.remoteTransformers;
+    }
```
n1k0: We should probably raise instead of silently ignoring the passed value here.

michielbdejong:
I went for `this._plugins` in the end. Hope you like it! Let me know if test coverage is ok, as well.

r? @n1k0

docs/api.md
```diff
@@ -412,7 +412,7 @@ Transformers are basically hooks for encoding and decoding records, which can wo
 
 ### Remote transformers
 
-Remote transformers aim at encoding records before pushing them to the remote server, and decoding them back when pulling changes. Remote transformers are registered by calling the `Collection#use()` method, which accepts a `Kinto.transformers.RemoteTransformer`-derived object instance:
+Remote transformers aim at encoding records before pushing them to the remote server, and decoding them back when pulling changes. Remote transformers are registered through the optional second argument of `Kinto#collection()`, which accepts `Kinto.transformers.RemoteTransformer`-derived object instances in its remoteTransformers array.
```
n1k0: Nit: `remoteTransformers` (backticks)

docs/api.md
```diff
@@ -482,12 +483,14 @@ class MyAsyncRemoteTransformer extends Kinto.transformers.RemoteTransformer {
...
-coll.use(new MyAsyncRemoteTransformer());
+coll = kinto.collection("articles", {
+  remoteTransformers: [ new MyAsyncRemoteTransformer() ]
+});
 ```
 
 ### Multiple transformers
 
-Transformers are stacked when `#use()` is called multiple times, in the order of the calls; that means you can chain multiple encoding operations, with the decoding ones being processed in the reverse order:
+The remoteTransformers field of the plugins object passed to `Kinto#collection` is an Array. That means you can chain multiple encoding operations, with the decoding ones being processed in the reverse order:
```
n1k0: I'm still debating if these are *plugins*, or more generally *configuration*; in the eventuality of options being added in the future, we could more easily reuse the concept.

src/collection.js
```diff
@@ -60,7 +60,9 @@ export default class Collection {
     this._bucket = bucket;
     this._name = name;
     this._lastModified = null;
-    this._transformers = {remote: []};
+    this._plugins = {
```
n1k0: Note: If we go with configuration rather than plugins, that would be `_config`.

src/collection.js
```diff
@@ -60,7 +60,9 @@ export default class Collection {
...
+      remoteTransformers: []
```
n1k0: No default idSchema?

src/collection.js
```diff
@@ -104,8 +106,32 @@ export default class Collection {
...
+   * @param  {Mixed} value  The value to set
+   */
+  _setPlugins(type, value) {
+    // Input checks.
+    switch(type) {
+    case "remoteTransformers":
+      if (!Array.isArray(value)) {
+        throw new Error("remoteTransformers should be an Array");
+      }
+      value.map(item => {
```
n1k0: Nit: use `Array#some` for a tiny performance improvement.

src/index.js
```diff
@@ -95,12 +95,39 @@ export default class Kinto {
...
+   *     remoteTransformers: [ otherTransformer, yetAnotherTransformer ]
+   *   });
+   *   kinto.collection("foo", {
+   *     remoteTransformers: [ myTransformer ]
+   *   });
+   * ````
    *
    * @param  {String} collName The collection name.
+   * @param  {Object} plugins May contain the following fields:
+   *                         remoteTransformers: Array of RemoteTransformers
```
n1k0: idSchema missing?

src/collection.js
```diff
@@ -104,8 +106,32 @@ export default class Collection {
     });
   }
 
-  use(transfomer) {
-    this._transformers[transfomer.type].push(transfomer);
+  /**
+   * Configures plugins like RemoteTransformers, LocalTransformers, and
+   * IdSchemas. Will be called from `Kinto#collection`
```
n1k0: Ah, I realize this PR doesn't handle idSchema at all, hence my confusion; we should remove any mention of it in the patch though.

src/index.js
```diff
@@ -95,12 +95,39 @@ export default class Kinto {
   }
 
   /**
-   * Creates or retrieve a Collection instance.
+   * Creates or retrieve a Collection instance. The second (optional) parameter
+   * will set, or update, the collection's plugin configuration. So the
+   * following bits of code all result in `foo` being the "foo" collection, and
+   * in its remoteTransformers being set to `[ myTransformer ]`:
+   *
+   * ````js
```
n1k0: Nit: If we want md fenced codeblocks, we want three backticks only.

test/collection_test.js
```diff
@@ -174,6 +174,55 @@ describe("Collection", () => {
...
   });
 
+  describe("#_setPlugins", () => {
+    var articles;
+
+    beforeEach(() => articles = testCollection());
+
+    it("should check that first argument is a supported type", () => {
+      try {
+        articles._setPlugins("bogus", {});
```
n1k0: Not sure we want to write unit tests for the private implementation; I'd rather see this tested along `Collection`'s constructor.

test/collection_test.js
```diff
@@ -174,6 +174,55 @@ describe("Collection", () => {
...
+        expect(false, "Should not reach here.");
+      } catch (err) {
```
n1k0: Also, you can use `expect` directly to check it raises, eg. `expect(fn).to.Throw(Error, /err msg regexp/)`.

test/collection_test.js
```diff
@@ -174,6 +174,55 @@ describe("Collection", () => {
...
+        expect(articles._plugins.remoteTransformers).to.be.instanceOf(Array);
+        expect(articles._plugins.remoteTransformers.length).to.equal(0);
+        expect(err.message).to.equal("Plugin type bogus not supported");
+      }
+    });
+
+    it("should set this._plugins.remoteTransformers", () => {
+      articles._setPlugins("remoteTransformers",
+          [ new QuestionMarkTransformer() ]);
+      expect(articles._plugins.remoteTransformers).to.be.instanceOf(Array);
```
n1k0: Nit: `to.be.a("array")` is a little more chai-idiomatic.

test/collection_test.js
```diff
@@ -174,6 +174,55 @@ describe("Collection", () => {
...
+      expect(articles._plugins.remoteTransformers.length).to.equal(1);
+      expect(articles._plugins.remoteTransformers[0]).to.be.instanceOf(QuestionMarkTransformer);
```
n1k0: Nit: Generally speaking, we're trying to have a single assertion per test — with a proper label — so we can immediately identify which part of the expected functionality is broken.

test/index_test.js
```diff
@@ -195,5 +195,69 @@ describe("Kinto", () => {
...
+      const db = new Kinto();
+      const coll = db.collection("plop", {
+        remoteTransformers: [ new MyRemoteTransformer() ]
+      });
+
+      expect(coll._plugins.remoteTransformers).to.be.an.instanceOf(Array);
+      expect(coll._plugins.remoteTransformers.length).to.equal(1);
+      expect(coll._plugins.remoteTransformers[0])
+          .to.be.an.instanceOf(MyRemoteTransformer);
+    });
```
n1k0: I think we should try to avoid testing provate property getters, especially here where it's already tested by public API code using that specific piece of the impl.

test/index_test.js
```diff
@@ -195,5 +195,69 @@ describe("Kinto", () => {
...
+        expect(coll._plugins.remoteTransformers).to.be.an.instanceOf(Array);
+        expect(coll._plugins.remoteTransformers.length).to.equal(1);
+        expect(coll._plugins.remoteTransformers[0])
+            .to.be.an.instanceOf(MyRemoteTransformer);
+      });
+    });
+
+    it("should update the collection's plugins on all instances", () => {
+      const MyRemoteTransformer = class extends RemoteTransformer {};
+      const MyOtherRemoteTransformer = class extends RemoteTransformer {};
```
n1k0: These are ducplicated in some other tests, maybe it's worth factorizing and expose them at the higher scope level.

n1k0:
This is shaping good, I like the way we explicitly declare the transformers we plan to use for a given collection as soon as we retrieve an instance of it; some small concerns with the tests which are deviating a little from the general adopted strategy. Hope my comments help here, though feel free to reach out for help if needed :)

Thanks a bunch!

docs/api.md
michielbdejong: The reason I was hesitant about 'options' was that the [collection constructor](https://github.com/Kinto/kinto.js/blob/master/src/collection.js#L59) already [receives the Kinto options](https://github.com/Kinto/kinto.js/blob/master/src/index.js#L117-L121) in a variable named options. If that means that we should merge with that, I do not know. :) I do agree that there are a lot of names working together:
* options
  * adapters
  * other things...
* plugins
  * IdSchemas
  * transformers
    * remote
    * local

src/collection.js
michielbdejong: I left all the IdSchema stuff in #143, so this PR only considers RemoteTransformers.

test/collection_test.js
michielbdejong: Oh, I would have expected that to fail because `typeof(new Array()) == "object"`, but I just tried it and you're right, in chai an array is an array and not an object. :)

michielbdejong:
Thanks for the detailed feedback, will improve the structure of the tests!

docs/api.md
michielbdejong: Maybe we should document it like this:
### Kinto-level options:
* remote (String)
* headers (Object with Strings)
* adapter (plugin)
* requestMode (String?)
* bucket (String)
* dbPrefix (String)

### Collection-level options:
* idSchema (plugin)
* remoteTransformers (Array of plugins)
* localTransformers (Array of plugins)

And then for symmetry, instead of having `Kinto.createRemoteAdapter` and `Kinto.createIdSchema`, we could eventually have `Kinto.createPlugin` which would cover those two and also [custom adapters](http://kintojs.readthedocs.org/en/latest/extending/#custom-database-adapters), as an ES5 alternative to extending `Kinto.adapters.BaseAdapter`, `Kinto.transformers.RemoteTransformer`, and the planned `Kinto.transformers.LocalTransformer` and (I guess) `Kinto.idSchemas.BaseIdSchema`.

So if you agree that this is a sensible way to name things, I'll rephrase:

    The remoteTransformers field of the plugins object passed to `Kinto#collection`

as:

    The remoteTransformers field of the collection-level options object

n1k0: > we could eventually have `Kinto.createPlugin`

Sounds super good, but would you envision things implementation-wise? This would be responsible of creating types with very distinct use case and apis…

test/collection_test.js
michielbdejong: OK, so maybe `Collection#setPlugins` and `Collection#plugins` should be public, just like [these](https://github.com/michielbdejong/kinto.js/blob/148-collection-configure/src/collection.js#L72).

The reason to separate `setPlugins` from the constructor is that the collections are cached in index.js, so only one instance of each collection exists on the kinto object. This means that if you call `kinto.collection("articles", options)` twice, the `Collection` constructor is only called the first time, but the last call should (I think) win in setting the collection-level options for both instance references.

n1k0: > the collections are cached in index.js

That's why we should test for the `Collection`constructor directly:

```js
describe("Collection", function() {
  describe("#constructor", function() {
    it("should register and apply plugins", function() {
      sandbox.stub();
      const coll = new Collection("foo", {
        remoteTransformers: [new TitleCharTransformer("!")]
      });
      // … Assert that records are properly transformer
    });
  });
});
```

michielbdejong: So what would be the expected/desired behavior if someone does:
````js
let a = kinto.collection("articles");
let b = kinto.collection("articles", { remoteTransformers: [ myTransformer ] });
````
Should the second argument be ignored if the collection is already cached? Or should the collections not be cached at all, maybe? (given that there will now be many ways to parameterize them).

Comment (n1k0) [inline test/collection_test.js]: Wow, super nice point. We should stop caching collection instances methinks.

**Comment by michielbdejong** [comment]
Coverage increased (+0.4%) to 100.0% \o/ :)

**Comment by michielbdejong** [comment]
@n1k0 wrote:
> Wow, super nice point. We should stop caching collection instances methinks.

OK, that would simplify the code further. I think most apps will cache their constructed collections anyway (at least the Sync app does).

File: docs/api.md [hunk 16]
```diff
@@ -412,7 +412,7 @@ Transformers are basically hooks for encoding and decoding records, which can wo
 
 ### Remote transformers
 
-Remote transformers aim at encoding records before pushing them to the remote server, and decoding them back when pulling changes. Remote transformers are registered by calling the `Collection#use()` method, which accepts a `Kinto.transformers.RemoteTransformer`-derived object instance:
+Remote transformers aim at encoding records before pushing them to the remote server, and decoding them back when pulling changes. Remote transformers are registered through the optional second argument of `Kinto#collection()`, which accepts `Kinto.transformers.RemoteTransformer`-derived object instances in its `remoteTransformers` array.
```
n1k0: We should probably add a first-level **Plugins** heading and describe their underlying concept.

src/index.js
```diff
@@ -102,18 +102,20 @@ export default class Kinto {
...
+   * Creates a Collection instance. The second (optional) parameter
+   * will set the collection-level plugin configuration.
    *
    * @param  {String} collName The collection name.
+   * @param  {Object} plugins May contain the following fields:
+   *                          remoteTransformers: Array of RemoteTransformers
    * @return {Collection}
    */
-  collection(collName) {
+  collection(collName, plugins = {}) {
```
n1k0: As per IRC disucssion, `options` would be more consistent.

test/collection_test.js
```diff
@@ -22,13 +22,17 @@ const TEST_COLLECTION_NAME = "kinto-test";
 const FAKE_SERVER_URL = "http://fake-server/v1";
 
 describe("Collection", () => {
-  var sandbox, events, api;
+  var sandbox, events, remoteTransformers, api;
   const article = {title: "foo", url: "http://foo"};
 
-  function testCollection() {
+  function testCollection(plugins={}) {
```
n1k0: As per IRC disucssion, `options` would be more consistent.

src/index.js
```diff
@@ -124,16 +126,12 @@ export default class Kinto {
...
-      });
-    }
-
-    return this._collections[collName];
+    const bucket = this._options.bucket;
+    return new Collection(bucket, collName, api, {
+      events:              this._options.events,
+      adapter:             this._options.adapter,
+      dbPrefix:            this._options.dbPrefix,
+      remoteTransformers:  plugins.remoteTransformers
```
n1k0: As per IRC disucssion, `options` would be more consistent.

docs/api.md
n1k0: Scratch this comment; as per IRC disucssion, `options` would be more consistent.

src/collection.js
```diff
@@ -71,6 +71,9 @@ export default class Collection {
     this.db = db;
     this.api = api;
     this.events = options.events || new EventEmitter();
+    this.plugins = {
```
n1k0: As per IRC disucssion, `options` would be more consistent.

src/index.js
```diff
@@ -102,18 +102,20 @@ export default class Kinto {
...
     this._options = Object.assign(defaults, options);
-    this._collections = {};
     // public properties
     this.events = this._options.events;
   }
 
   /**
-   * Creates or retrieve a Collection instance.
+   * Creates a Collection instance. The second (optional) parameter
+   * will set the collection-level plugin configuration.
```
n1k0: s/plugin/options

n1k0:
@michielbdejong r=r+wc

src/collection.js
michielbdejong: Here, I used `this.remoteTransformers` instead of `this.options.remoteTransformers`, for consistency with for instance the line just above, which also uses `this.events` and not `this.options.events`.

michielbdejong:
Rebased and squashed.

src/collection.js
n1k0: Makes sense.

michielbdejong:
Thanks for all your help with this!!

PR #159, opened 2015-09-15
Section label resolving-conflicts -> resolving-conflicts-manually

michielbdejong:
I think the section label should be "resolving-conflicts-manually". Do you want me to open an issue first and prefix this commit with "Fixes #xxx - "?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -235,7 +235,7 @@ Synopsis:
 
 1. Fetch remote changes since last synchronization;
 2. Fail on any conflict encountered;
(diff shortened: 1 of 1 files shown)


n1k0:
Actually this part of the docs is still misleading; depending on the strategy adopted, fetched conflicts might get resolved automatically (`*_WINS` strategies). We should revamp this whole section imho.

Natim:
We can do that in another PR, at the moment this PR only fixes existing links.

Issue #160, opened 2015-09-15
Non-empty syncResults.updated when syncing to an empty remote collection


michielbdejong:
When you run `go()` in the console of http://michielbdejong.github.io/kinto-encrypter-les-donnees/ you see the syncUp step (which uploads data to an empty collection on the public test server) results in:
````
Sync up Object { ok: true, lastModified: 1442328395941, errors: Array[0], 
created: Array[0], updated: Array[1], deleted: Array[0], published: Array[1], 
conflicts: Array[0], skipped: Array[0], resolved: Array[0] }
````
Here, the `published` array contains the encrypted data (which makes sense, I think), but the `updated` array contains that same element. It also happens if you don't use a RemoteTransformer. Is this because `_synced: true` is updated on that record?

n1k0:
> Is this because `_synced: true` is updated on that record?

[Yes](https://github.com/Kinto/kinto.js/blob/82b00c26ce754058ef013ded37e025583ef926f8/src/collection.js#L337-L339). Do you feel it shouldn't be that way?

michielbdejong:
Hm, if I think more about it, there are lots of different cases, in the cross-product of {incoming,outgoing,conflict,resolved,skipped} with {create,update,delete} and {ok,error}. I would expect syncResults to contain the following (covering each combination once):
* errors: {incoming,outgoing,conflict,resolved,skipped} x {create,update,delete} x {error}
* created: {incoming} x {create} x {ok}
* updated: {incoming} x {update} x {ok}
* deleted: {incoming} x {delete} x {ok}
* published: {outgoing} x {create,update,delete} x {ok}
* conflicts: {conflict} x {create,update,delete} x {ok}
* resolved: {resolved} x {create,update,delete} x {ok}
* skipped: {skipped} x {create,update,delete} x {ok}

Here, the first dimensions ({incoming,outgoing,conflict,resolved,skipped}) would be:
* incoming - the record was only changed remotely
* outgoing - the record was only changed locally
* conflict - if the record was changed both locally and remotely, and the conflict was not resolved automatically.
* resolved - if the record was changed both locally and remotely, but the conflict was resolved automatically.
* skipped - *almost* the same as resolved, except that this is the special case where `conflict.local` and `conflict.remote` are identical (both deleted, both created equally, both updated equally), and so we don't even want to call this a "conflict", because its resolution strategy is moot.

[...]

Issue #144, opened 2015-09-04
Avoid redownloading changes on Sync


Natim:
Refs: mozilla-services/readinglist#68

Basically, when doing a Batch before syncing, in the next update you will get all the records you've just changed that are already up-to-date with your local version. (When using ``If-Match`` headers to handle races)

We've just landed something, that let you exclude a list of values when syncing: https://github.com/mozilla-services/cliquet/pull/432

This is something we may want to use in kinto.js sync to optimize the second fetch.

n1k0:
That would be basically adding an `exclude` query string parameter to the second fetch for changes request with the already processed record ids, right?

leplatrem:
Using `?exclude_id=id1,id2,id3` it should work.

What is the limit of the URL length by the way ?

leplatrem:
> What is the limit of the URL length by the way ?

Staying under 2000 chars looks like the most reasonable approach ([source](http://stackoverflow.com/a/417184))

PR #161, opened 2015-09-16
Refs #160: Sync flow and result object format optimizations.

n1k0:
Following discussion on #160 ([this comment](https://github.com/Kinto/kinto.js/issues/160#issuecomment-140712336)), I've experimented with removing the second call to `pullChanges`, and to my great surprise all the tests were mostly passing.

I went further, and now locally touched records after that second pull are no more listed in the `updated` array property of the sync result object — that was clearly misleading.

I've also found a bug which was preventing encoded records to be decoded when updated after having been published; also, the `published` sync result property now lists records in their decoded flavor — which is probably more user-friendly.

I'm having hard times figuring out if I'm missing any obvious use case with this patch, but I'm still pretty confident its viable when it comes to all functional tests being green.

Also admire the concision of that new version of the`sync` method; no mention we're gaining one HTTP request, which is good from a performances standpoint.

Would love hearing thoughts from @michielbdejong @Natim @ametaireau @leplatrem and anybody interested in this topic.

**Update**

Had to reintroduce that second call to `pullChanges` in order to address possible race conditions in-between synchronization steps. Other addressed points mentioned above remain.


n1k0:
Gosh, it's picky.

screen shot 2015-09-16 at 16 52 27

src/collection.js
```diff
@@ -506,7 +506,8 @@ export default class Collection {
             });
           } else {
             // Remote update was successful, reflect it locally
-            return this.update(record, {synced: true});
+            return this._decodeRecord("remote", record)
```
Natim: Nice catch.

leplatrem:
> anybody interested in this topic.

I'm sorry I cannot dig into this deeply. But it looks good to me. AFAIU one drawback is that you will download your own pushed records during the next sync (because `this._lastModified` is only updated during `importChanges()`).

Side note: don't forget to update the diagram :)

n1k0:
> (because `this._lastModified` is only updated during `importChanges()`).

Yeah, that was basically the main purpose of that second pull, as well as force-marking remote version as synced locally.

> Side note: don't forget to update the diagram :)

Good point!

n1k0:
After playing with the demo using that patch, I'm finding annoying to have to sync twice in order to have an all clear sync result object. I still need to work out how to properly mark published records as synced locally, with the proper `last_modified` value updated with the right data (which are obtained in the batch request individual responses I think).

n1k0:
Well, it appears the only way I have to update the remote collection `lastModified` value from the server is indeed by pulling changes from it; and we absolutely need to update this value to ensure next sync operation will use the latest known timestamp… Would it be conceivable to have the latest collection `lastModified` value retrieved from a BATCH main response?

docs/api.md
```diff
@@ -291,24 +291,40 @@ articles.sync({
...
+  deleted:   [],
+  skipped:   [],
+  published: [],
+  resolved:  [],
 }
 ```
 
+The synchronization result object exposes the following properties:
+
+- `ok`: The boolean status of the synchronization operation; `true` if no unsolvable conflicts and no errors were encountered.
```
michielbdejong: unsolvable -> unresolved - presumably, conflicts that are handled manually are in themselves solvable, it just hasn't happened yet. :)

docs/api.md
```diff
@@ -291,24 +291,40 @@ articles.sync({
...
+- `lastModified`: The timestamp of the latest revision of server data.
```
michielbdejong: Is that one timestamp for the whole collection? This timestamp is in milliseconds since 1970, right? Does the server guarantee that any changes that happen immediately after one of the clients synced happen in the next millisecond?

docs/api.md
```diff
@@ -291,24 +291,40 @@ articles.sync({
...
+- `conflicts`: The list of encountred conflicts during both import and export operations (see *[Resolving conflicts manually](#resolving-conflicts-manually)*);
```
michielbdejong: But only the ones that were encountered and that require manual resolution, right?

Nit: encountred -> encountered

docs/api.md
```diff
@@ -291,24 +291,40 @@ articles.sync({
...
+- `errors`:    The list of encountered errors, if any.
```
michielbdejong: Per record (so one 401 error at the http-level can result in 20 errors at the record-level here). Is it guaranteed that a record that is listed here was left unchanged? For instance, if a record is pulled in, but the RemoteTransformer throws an error, is the lastModified value reset to (or left at) what it was, so that it will be retried next time?

docs/api.md
```diff
@@ -291,24 +291,40 @@ articles.sync({
...
+- `created`:   The list of remote records which have been succesfully imported into the local database.
+- `updated`:   The list of remote record updates which have been succesfully reflected into the local database.
+- `deleted`:   The list of remotely deleted records which have been succesfully deleted as well locally.
+- `skipped`:   The list of identical records which have been declared as synced automatically.
+- `published`: The list of locally modified records which have been succesfully pushed to the remote server.
```
michielbdejong: "modified" here meaning "created, updated, or deleted". But I guess that's clear enough.

docs/api.md
```diff
@@ -291,24 +291,40 @@ articles.sync({
...
+- `resolved`:  The list of conflicting records which have been succesfully resolved according to the selected strategy (note that using the default `MANUAL` strategy, this list is always empty).
```
michielbdejong: Nit: succesfully -> successfully

src/collection.js
```diff
@@ -506,7 +506,8 @@ export default class Collection {
             });
           } else {
             // Remote update was successful, reflect it locally
```
michielbdejong: Nit: Remote create or update

michielbdejong:
> Would it be conceivable to have the latest collection lastModified value retrieved from a BATCH main response?

That wouldn't help you. Say you pulled at 12:00, and then publish a new record at 12:07, and learn that the lastModified of the collection as a whole now equals the `response.data.last_modified` from [the response body of the POST](http://kinto.readthedocs.org/en/latest/api/records.html#records-post) that you just did. That still tells you nothing, another client may have made some other change at 12:03.

I think the best option is to leave the "last full sync" timestamp for the collection at 12:00 and set `this.lastModified` of just the record that you published at 12:07, to 12:07 (using the `response.data.last_modified` value from that request response). Then, at 12:10 do another sync, in which you will see for instance:
* new collection timestamp: 12:07
* record created at 12:03 (this is a change that still needs to be reflected locally)
* record created at 12:07 (the one you published yourself, so ignore)

michielbdejong:
It's just a pity that all data that you publish is also sent back to you again, which can be quite inefficient if records contain big blobs of data. You're basically guaranteed to have twice as much traffic as needed for outgoing changes.

If batch operations are atomic, then maybe it would make sense to do a GET at the beginning of the batch, and then thanks to batch atomicity you could safely set the collection timestamp to the highest one from the batch.

Or maybe add per-record _since timestamps to the GET query, but that would become quite an ugly API, probably.

docs/api.md
Natim: > Is that one timestamp for the whole collection?

Yes

> This timestamp is in milliseconds since 1970, right?

Yes

> Does the server guarantee that any changes that happen immediately after one of the clients synced happen in the next millisecond?

Yes.

docs/api.md
Natim: Changes are individual for the list so if 1 worked over 20, the collection timestamp changed and the record that worked changed. But the other records are left unchanged.

Natim:
> It's just a pity that all data that you publish is also sent back to you again, which can be quite inefficient if records contain big blobs of data.

Yes this is a known limitation and we have a way to fix it. See #144

n1k0:
I think I'll just revert the change removing that second call, which is the best way to ensure any race is properly addressed during the flow; #144 will address part of the performance concerns. We may come back to this later anyway.

docs/api.md
michielbdejong: I guess the collection timestamp always stays before the oldest record that errored, to make sure all errors get endless retry until they succeed?

n1k0: Nope, and that's a good point, I'll investigate that.

n1k0: Filed #163 about it.

n1k0:
Failed build says:

```
$ nvm install 0.12

Version '0.12' not found - try `nvm ls-remote` to browse available versions.
```

Now that's new -_-

n1k0:
This was a temporary issue on Travis it seems. This is now ready for review; r=? @michielbdejong @Natim

Natim:
It looks good to me.

docs/api.md
```diff
@@ -291,24 +291,40 @@ articles.sync({
...
+  skipped:   [],
+  published: [],
+  resolved:  [],
 }
 ```
 
+The synchronization result object exposes the following properties:
+
+- `ok`: The boolean status of the synchronization operation; `true` if no unresolved conflicts and no errors were encountered.
+- `lastModified`: The timestamp of the latest revision of server data.
```
michielbdejong: This description can be expanded and made more precise as part of #163. For now it's close enough, I think.

docs/api.md
```diff
@@ -291,24 +291,40 @@ articles.sync({
...
+- `conflicts`: The list of encountered conflicts during both import and export operations (see *[Resolving conflicts manually](#resolving-conflicts-manually)*);
```
michielbdejong: Only the unresolved ones, not the resolved ones. But the reader will probably understand that.

test/integration_test.js
```diff
@@ -167,10 +167,8 @@ describe("Integration tests", () => {
           expect(cleanRecord(syncResult.published[0])).eql(testData.localUnsynced[0]);
         });
 
-        it("should mark local records as synced", () => {
```
michielbdejong: Are we still testing that it does actually mark the local record as synced? I guess we were testing that implicitly through the inclusion in`updated`, and now we're not testing it at all anymore? (at least not here).

docs/api.md
michielbdejong: An example of when this description is incorrect is if there are unresolved conflicts in the SyncResult for instance, see this test https://github.com/Kinto/kinto.js/blob/sync-result-updated-list/test/integration_test.js#L352

test/integration_test.js
```diff
@@ -779,10 +1059,10 @@ describe("Integration tests", () => {
             ]);
           });
 
-          it("should encode records when pushed to the server", () => {
```
michielbdejong: Are we still testing that it actually does do the encoding. The single biggest bug risk in terms of implications is if unencrypted data leaks out of a Kinto.js-based app due to some bug where the RemoteTransformer does not get applied in some edge case, so we should have supergood integration tests for that.

michielbdejong: Ah, found it. https://github.com/Kinto/kinto.js/blob/sync-result-updated-list/test/integration_test.js#L1068

michielbdejong:
r+ from me too. (maybe create a follow-up for https://github.com/Kinto/kinto.js/pull/161/files#r39848183 if needed).

Thanks for addressing this! It makes the SyncResults a lot clearer.

test/integration_test.js
n1k0: Yes, in the [very next test](https://github.com/Kinto/kinto.js/blob/sync-result-updated-list/test/integration_test.js#L174-L187).

test/integration_test.js
n1k0: Yeah, using transofrmers end to end integration tests, we're pretty much safe here.

docs/api.md
n1k0: It's worth a precision, will update.

PR #164, opened 2015-09-18
Fixes #163: Ensure lastModified isn't saved on import errors encountered

### LINKED ISSUE #163: Ensure lastModified value isn't updated if errors are encountered
From comment https://github.com/Kinto/kinto.js/pull/161#discussion_r39720345

> I guess the collection timestamp always stays before the oldest record that errored, to make sure all errors get endless retry until they succeed?

We should probably ensure lastModifed value isn't bumped on errors encountered.

n1k0:
r=? @michielbdejong


michielbdejong:
r+wc: Just update the two comments to say "(no) conflicts *or errors* occured".

Great!

PR #166, opened 2015-09-18
Correct code comment for `Collection#pullChanges`

michielbdejong:
The statement that this function will reject if it encounters conflict seemed incorrect to me. Also, to say it imports changes, if there is also a function called `importChanges` seemed confusing.


n1k0:
r+

PR #168, opened 2015-09-21
Fixes #158 - Collection#create Id requirements validation.

### LINKED ISSUE #158: Raise an error when an id is passed to Collection#create but not necessary
If you set the `useRecordId` option to true in `Collection#create`, then the `record` object you pass should already contain the `id` field to be validated. Same if you set the `synced` option to true.

If you set neither of these options, an id is generated using the *IdSchema* of the collection. In this case, the record passed should not have an `id` field, and we should reject with an error if it does.

This is an idea that came up in https://github.com/Kinto/kinto.js/pull/143#discussion_r39418146

n1k0:
refs #158 - r=? @michielbdejong

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -159,6 +159,11 @@ export default class Collection {
   /**
    * Adds a record to the local database.
    *
+   * Note: If either the `useRecordId` or `synced` options are true, then the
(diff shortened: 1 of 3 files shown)


michielbdejong:
r+, nice catch in the integration tests! :)

Issue #155, opened 2015-09-09
Avoid typing and extending classes for transformers


ametaireau:
Transformers currently need to extend from a known base class. Rather than doing it this way, we could  use duck typing rather than typing, and accept any object with the correct methods.

michielbdejong:
Yes, I was thinking the same thing. That would actually also make the whole API after #149 simpler, because we can get rid of `Kinto#createRemoteTransformer` (it might still be useful to people if we keep exposing `Kinto.transformers.RemoteTransformer`, but at least it makes it easier to define a transformer). The same would apply to `IdSchema` and `Kinto#createIdSchema` from #143.

n1k0:
The useful thing classes add is constructors, so you can pass arguments to the created instances. A good example how it's useful is the [`TitleCharTransformer`](https://github.com/Kinto/kinto.js/blob/741b24e9a3849abdb0d5d67c8229af4b2196af7b/test/integration_test.js#L713-L724).

Sure we could keep providing a `constructor` hook to the passed object literal, though that would be quite dirty and out of the common js practices.

Of course it's still possible to pass a built object referencing some of its property, though that would harm reusability; how would you guys preserve the ability to construct transformers, and/or think of a way of achieving the very same functionality?

michielbdejong:
The way I did it in the end for RemoteTransformers is that `kinto.createRemoteTransformer` still handles the constructor as desired, so you the TitleCharTransformer snippets from http://kintojs.readthedocs.org/en/latest/api/#transformers (both the ES6 one and the ES5 one) still work the same way, but if you want something simpler without constructor, the following is now also allowed:
````js
kinto.collection('articles', { remoteTransformers: [ {
  encode(record) { ... },
  decode(record) { ... }
} ]);
````

**Comment by n1k0** [comment]
Yeah I get that, though declaring a `constructor` prototype property is a little weird in the sense it's usually perceived as a reserved keyword, and people might expect a given behavior… Maybe that's just a minor issue. I suspect users will simply create a function to generate their own *constructed* transformers:

```js
function createTitleCharTransformer(char) {
  return {
    type: "remote",
    encode: function(record) {
      // we can use `char` here
    }
  };
}

const coll = kinto.collection("foo", {
  remoteTransformers: [ createTitleCharTransformer("?") ]
});
```

[...]

Issue #154, opened 2015-09-09
Add a CollectionTransformer concept


almet:
For the collection signing, we need a way to alter and retrieve metadata information contained in the collection.

After a quick chat with @n1k0 it seems to make sense to add a concept of `CollectionTransformers`.

You could then define a transformer that would be called before pushing new data to the collection, or just before retrieving new data from the collection.

leplatrem:
How would that differ from #153 ?

almet:
That's a good question. They seem very similar.

The event approach seems to just be a notification system (that can't intercept the logic going on), while the second (this one) can.

Anyway, I don't think we need the two approach at the same time, and we should pick one or the other.

leplatrem:
Implemented in #302

PR #171, opened 2015-09-21
Fixes 155: Drop use of classes for transformers and IdSchema.

n1k0:
Refs #155.

- [x] Drop use of classes for remote transformers
- [x] Ensure provided transformers comply to the expected signature
- [x] Drop use of classes for IdSchema
- [x] Ensure provided IdSchema complies to the expected signature
- [x] Update docs
- [x] Update dist files


n1k0:
This is now ready for review. r=? @michielbdejong

ametaireau:
Great. Once again, less is more.

docs/api.md
```diff
@@ -429,7 +429,7 @@ Transformers are basically hooks for encoding and decoding records, which can wo
 
 ### Remote transformers
 
-Remote transformers aim at encoding records before pushing them to the remote server, and decoding them back when pulling changes. Remote transformers are registered through the optional second argument of `Kinto#collection()`, which accepts `Kinto.transformers.RemoteTransformer`-derived object instances in its `remoteTransformers` array.
+Remote transformers aim at encoding records before pushing them to the remote server, and decoding them back when pulling changes. Remote transformers are registered through the optional second argument of `Kinto#collection()`, which accepts transformer objects in its `remoteTransformers` array.
```
ametaireau: Should we specify that a transformer object has two functions `encode` and `decode`?

n1k0: Indeed, will add a section about the expected signatures.

n1k0:
I'm now thinking that we don't check that a transformer object complies to the expected signature. It's probably worth ensuring it provides at least both `encode` and `decode` methods.

n1k0:
Now need to do the same for IdSchema :)

docs/api.md
```diff
@@ -429,7 +429,9 @@ Transformers are basically hooks for encoding and decoding records, which can wo
...
+
+A transformer object is basically an object literal having and `encode` and a `decode` method, both accepting a `record` object and returning that record transformed, or a Promise resolving with that record transformed:
```
michielbdejong: Note that `record.id` should always be left unchanged by a transformer, and the `encode` should have the exact reverse effect of `decode`.

docs/api.md
```diff
@@ -475,21 +477,21 @@ coll.create({title: "foo"}).then(_ => coll.sync())
 
 ### Local transformers
 
-In a near future, Kinto.js will provide transfomers aimed at providing facilities to encode and decode records when persisted locally; you'll have to extend from `LocalTransformer` though.
+In a near future, Kinto.js will provide transfomers aimed at providing facilities to encode and decode records when persisted locally.
```
michielbdejong: Maybe lump local transformers and [collection transformers](https://github.com/Kinto/kinto.js/issues/154) into one "transformers in general" remark, and move that to line 148 (inside "Transformers" section, but before "Remote transformers" section)

docs/api.md
```diff
@@ -429,7 +429,9 @@ Transformers are basically hooks for encoding and decoding records, which can wo
```
michielbdejong: "For now, only remote transformers have been implemented, but there are plans to implement local transformers and collection transformers in a next version." (and then remove the 'local transformers' section, see https://github.com/Kinto/kinto.js/pull/171/files#r40066296)

src/collection.js
```diff
@@ -114,6 +127,48 @@ export default class Collection {
...
+  }
+
+  /**
+   * Validates a list of remote transformers.
+   *
+   * @param  {Array} remoteTransformers
+   * @return {Array}
+   */
+  _validateRemoteTransformers(remoteTransformers) {
+    if (!remoteTransformers || remoteTransformers.length === 0) {
```
michielbdejong: You should check here if remoteTransformers is an error, otherwise this will throw `remoteTransformers.map is not a function` if people pass `{ remoteTransformers: { length: 1 } }` as the collection options. OK, nobody in their right mind would ever pass that of course, so it's all academic, but still. ;)

src/collection.js
```diff
@@ -114,6 +127,48 @@ export default class Collection {
   }
 
   /**
+   * Validates an idSchema.
+   *
+   * @param  {Object} idSchema
+   * @return {Object}
+   */
+  _validateIdSchema(idSchema) {
+    if (!idSchema) {
```
michielbdejong: `if (typeof idSchema === "undefined") {` would be slightly more correct (although nobody should in practice specify `false` or `''` as their id schema, if we're checking the input then we might as well do it fully correctly).

Natim: or ``idSchema === undefined``

test/integration_test.js
```diff
@@ -1018,76 +1017,55 @@ describe("Integration tests", () => {
...
-              })
-              .then(res => res.json())
-              .then(res => res.data.map(x => x.title).sort())
-              .should.become(["abc!?", "def!?"]);
-          });
+      it("should list published records unencoded", () => {
+        return tasks.sync()
+          .then(res => res.published.map(x => x.title).sort())
+          .should.become(["abc", "def"]);
+      });
```
michielbdejong: Merge conflict?

michielbdejong:
r+wc from me! :) Great to see so much code disappear ;)

docs/api.md
n1k0: I guess you meant line *481*.

n1k0: Ah, no, line 431 it is.

docs/api.md
n1k0: > and that encode should have the exact reverse effect of decode.

There's a note about this already just after the code snippet.

test/integration_test.js
n1k0: Huh. Investigating.

n1k0: Ah no, it's just test refactoring now we don't need two test suites for es5/6 transformers :)

PR #174, opened 2015-09-23
Fixes #165: Expose per-record import errors.

### LINKED ISSUE #165: Handling errors during the this.create step of _importChange
If https://github.com/Kinto/kinto.js/blob/163-prevent-lastModified-bump-on-errors/src/collection.js#L378 rejects for one record (e.g. record is not an object, id cannot be validated, disk is full), from how I read the code, it will skip everything after https://github.com/Kinto/kinto.js/blob/163-prevent-lastModified-bump-on-errors/src/collection.js#L393-L413, including the call `pushChanges`.

That seems a bit drastic to me, I feel this should simply be added to `SyncResults#errors`, as a per-record error?

Now, 

Are there tests for e.g. disk full errors from IndexedDB?

n1k0:
Refs #165. r=? @michielbdejong


michielbdejong:
LGTM, r+

One a side node (maybe for a follow-up): it's a pity that the error format [for outgoing](https://github.com/Kinto/kinto.js/blob/165-expose-per-record-import-errors/src/api.js#L207-L211) is so different from the errors format [for incoming](https://github.com/Kinto/kinto.js/blob/165-expose-per-record-import-errors/src/collection.js#L440), but with this patch  at least all incoming errors have the same format, regardless of whether they come from e.g. `RemoteTransformer#decode` or from e.g. `IdSchema#verify`.

n1k0:
> it's a pity that the error format for outgoing is so different from the errors format for incoming

I agree, though these are two very different kinds of error (http vs idb); could you think of a more homogeneous format?

n1k0:
Anyway let me land this first, we should file another issue to address this specific problem :)

PR #175, opened 2015-09-23
Fixes #169 - Instantiate Api only once

### LINKED ISSUE #169: Avoid checking the server settings for each collection
For the FxOS Sync app, currently (or actually after we fix bug [1206694](https://bugzilla.mozilla.org/show_bug.cgi?id=1206694) and [bug 1206695](https://bugzilla.mozilla.org/show_bug.cgi?id=1206695)), even if no data changed on the server, nor locally, each sync cycle incurs `4n+4` requests, for `n` the number of collections (currently three, so 16 requests). They are:

* 4 GET requests to 'hello' URL on http://localhost:8000/v1/, to check the server version
* 1 GET request to meta/global (this is mandatory)
* 3 GET requests to ${collection}/records (one for each collection)
* one OPTIONS request for each of these GET requests, (unavoidable, but can be served statically by Syncto or Kinto)

It seems a bit overkill to check http://localhost:8000/v1/ four times within a few seconds. The load impact on the Syncto or Kinto server is probably minimal, since this is probably just a static page, but these requests need to complete before the next thing happens, so it costs time. Their response looks like this, each time:

````js
{"settings":{"cliquet.batch_max_requests":25},"url":"http:\/\/localhost:8000\/v1\/","documentation":"https:\/\/syncto.readthedocs.org\/","version":"1.0.0.dev0","commit":"ebc27e3","hello":"syncto"}
````

Would it be an option to cache these responses (per server URL) at the Kinto level?

michielbdejong:
Looks like the Api object only depends on Kinto-level options, not on Collection-level options.

So is this safe this way? @n1k0 

(didn't run tests yet)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -66,6 +66,12 @@ export default class Kinto {
       remote: DEFAULT_REMOTE,
     };
     this._options = Object.assign(defaults, options);
+    this._api = new Api(this._options.remote, {
+      headers:     this._options.headers,
+      events:      this._options.events,
+      requestMode: this._options.requestMode,
+    });
+
     // public properties
     this.events = this._options.events;
   }
@@ -84,14 +90,8 @@ export default class Kinto {
       throw new Error("missing collection name");
     }
 
-    const remote = this._options.remote;
-    const api = new Api(remote, {
-      headers:     this._options.headers,
-      events:      this._options.events,
-      requestMode: this._options.requestMode,
-    });
     const bucket = this._options.bucket;
-    return new Collection(bucket, collName, api, {
+    return new Collection(bucket, collName, this._api, {
       events:              this._options.events,
       adapter:             this._options.adapter,
       dbPrefix:            this._options.dbPrefix,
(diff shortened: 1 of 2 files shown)


n1k0:
Sounds good. We should now add a test to ensure settings are retrieved a single time *per instance* after the first request is being issued.

michielbdejong:
@n1k0 Any idea why the Integration tests "before each" hook is failing now on Travis? Same thing is happening on #179. Is it also failing for you locally? (for me it is, but that could also be a problem with my KINTO_PSERVE_EXECUTABLE and things like that).

n1k0:
> Any idea why the Integration tests "before each" hook is failing now on Travis? 

No idea, we're discussing about it with @Natim and we're clueless so far :(

n1k0:
> Is it also failing for you locally?

No, works fine on my local machine.

n1k0:
r+ :)

michielbdejong:
Rebased for green tests.

n1k0:
\o/

PR #179, opened 2015-09-25
'git co' -> 'git clone' in install instructions

michielbdejong:
Also added use of \`pwd\` in `KINTO_PSERVE_EXECUTABLE` instructions, so that it's automatic.

Btw, I could not find any mention of `KINTO_PSERVE_EXECUTABLE` in `.env/`, `.venv/` or `./venv/` - so is this step not always necessary by definition?

### DIFF
--- a/docs/hacking.md
+++ b/docs/hacking.md
@@ -6,7 +6,7 @@ Hacking on Kinto.js requires a working [NodeJS v0.12x environment](https://nodej
 
 Source code is [hosted on Github](https://github.com/mozilla-services/kinto.js).
 
-    $ git co https://github.com/mozilla-services/kinto.js
+    $ git clone https://github.com/mozilla-services/kinto.js
     $ cd kinto.js
     $ npm install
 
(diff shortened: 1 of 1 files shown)


Natim:
> I could not find any mention of KINTO_PSERVE_EXECUTABLE

The mention is here: ./test/integration_test.js:18

Also the value of ``KINTO_PSERVE_EXECUTABLE`` default to pserve in the path (which might not be what we want)

Natim:
r+, I asked @n1k0 to have a look at travis-ci. Thank you.

docs/hacking.md
```diff
@@ -34,7 +34,7 @@ Then you can run the tests by running:
 
 You can alternatively run the tests in a more explicit fashion by passing the `KINTO_PSERVE_EXECUTABLE` environment variable, which can be found in the virtual environment's `bin` directory:
 
-    $ KINTO_PSERVE_EXECUTABLE="/path/to/venv/bin/pserve" npm run test
+    $ KINTO_PSERVE_EXECUTABLE="`pwd`/venv/bin/pserve" npm run test
```
n1k0: This really depends on what's the actual output of `pwd`; users might still be in their `kinto.js` directory, and that wouldn't work.

michielbdejong: If I have kinto.js checked out in `/Users/Michiel/repos/kinto.js`, then venv command installs pserve in `/Users/Michiel/repos/kinto.js/venv/bin/pserve`, right? So then it's correct?

n1k0:
r+

PR #180, opened 2015-09-28
[Parked] Batch updates.

n1k0:
This is an attempt at supporting transactional bulk updates at the adapter level. The final plan is to journalize sync persistence operations and perform them all at once within a single indexedDB transaction ~~, but for now the work has been initiated at the adapter level only~~; I'm requesting @michielbdejong early feedback to know if I'm missing anything obvious.

- [x] Implement `IDB#batch`
- [x] Wrap all sequential db updates in `Collection` within a single transaction wherever applicable
- [x] Expose the list of batch planned operations in the result
- [x] Allow exposing the result returned by a passed batch function
- [x] Ensure consistent batch error reporting
- [x] Migrate `IDB#get()` to use `IDB#batch()`
- [x] Migrate `IDB#create()` to use `IDB#batch()`
- [x] Migrate `IDB#update()` to use `IDB#batch()`
- [x] Migrate `IDB#delete()` to use `IDB#batch()`
- [x] Migrate `IDB#clear()` to use `IDB#batch()`
- [x] Migrate `IDB#list()` to use `IDB#batch()`
- [x] Implement `IDB#batch.abort()`
- [ ] Implement batch at the `BaseAdapter` level for non-transactional adapters
- [x] Update docs
- [x] Update dist files


src/collection.js
```diff
@@ -460,6 +460,8 @@ export default class Collection {
    * @return {Promise}
    */
   importChanges(syncResultObject, changeObject) {
+    // XXX: Ensure all imports are done within a single transaction
+    // Remove this and replace by batch(changes.map(change => batch[action](change)))
     return Promise.all(changeObject.changes.map(change => {
       return this._importChange(change);
     }))
```
michielbdejong: It would be nice if `batch(changes.map(change => batch[action](change)))` could return a promise for a results object like the one `Promise.all` returns. Alternatively, change the code in line https://github.com/Kinto/kinto.js/pull/180/files#diff-d2dbe7779d59d4dd594e92248caba0a5R469 to match the `{operations, errors}` syntax.

michielbdejong: Also, `_importChange` includes a [`this.get` call](https://github.com/Kinto/kinto.js/pull/180/files#diff-d2dbe7779d59d4dd594e92248caba0a5R434), which also needs to be done within the batch to keep things safe (actually I think that code is not even safe right now?). But you can't just include it as an extra operation in the batch, because then you won't get the result of the `get` operation out.

Maybe an alternative approach could be:

````js
batch(function executor(transaction) {
  // do lots of stuff here,
  // but call transaction.get, transaction.create, etc.
  // instead of this.get, this.create.
});
````
And then `batch` does:
* prepare transaction,
* execute the executor function passed to it
* resolve from the transaction's `onsuccess` etc. handlers
* the transaction object with which the executor function is called has a `.get` function that works

**Comment by n1k0** [comment]
@michielbdejong implemented what you suggested in 7e84ba722e5b42d47de47c6d88cb10eab02bc0c9, feedback stil warmly welcomed :)

**Comment by n1k0** [comment]
So 484a87ba193dbb27fd04a6d2f91fbfd120e40d0e provides green tests, though I suspect they'll lack coverage. While there's a whole lot of room from refactoring, I would love hearing feedback from @michielbdejong at this point.

File: src/adapters/IDB.js [hunk 2]
```diff
@@ -81,6 +87,62 @@ export default class IDB extends BaseAdapter {
...
+   * @param  {Function} fn The operation function, which must return a Promise.
+   * @return {Promise}
+   */
+  batch(fn) {
+    return this.open().then(() => {
+      const errors = [], operations = [];
+      const {transaction, store} = this.prepare("readwrite");
+      const batchObject = {
+        get(id) {
+          if (!id) {
```
michielbdejong: This same check does not seem to be present in https://github.com/Kinto/kinto.js/blob/172-bulk-operations/src/adapters/IDB.js#L203 ?

src/adapters/IDB.js
```diff
@@ -81,6 +87,62 @@ export default class IDB extends BaseAdapter {
   }
 
   /**
+   * Batch operations.
+   *
+   * @param  {Function} fn The operation function, which must return a Promise.
```
michielbdejong: Maybe give `fn` a more descriptive name? Although I can't really think of a good name for it. :)

src/adapters/IDB.js
```diff
@@ -5,6 +5,12 @@ import { attachFakeIDBSymbolsTo } from "./../utils";
 
 attachFakeIDBSymbolsTo(typeof global === "object" ? global : window);
 
+const BATCH_STORE_METHODS = {
+  create: "add",
+  update: "put",
+  delete: "delete"
```
michielbdejong: Maybe add `clear` as well?

src/adapters/IDB.js
```diff
@@ -81,6 +87,62 @@ export default class IDB extends BaseAdapter {
...
+        get(id) {
+          if (!id) {
+            return Promise.reject(new Error("Id not provided."));
+          }
+          return new Promise((resolve, reject) => {
+            const request = store.get(id);
+            request.onerror = event => reject(new Error(event.target.error));
+            request.onsuccess = () => resolve(request.result);
+          });
+        }
```
michielbdejong: Maybe add `list` as well?

src/adapters/IDB.js
```diff
@@ -81,6 +87,62 @@ export default class IDB extends BaseAdapter {
...
+      };
+      for (let type of Object.keys(BATCH_STORE_METHODS)) {
+        batchObject[type] = data => {
+          const operation = {type, data};
+          operations.push(operation);
+          return operation;
+        };
+      }
+      const result = fn(batchObject);
+      return Promise.resolve(result)
```
michielbdejong: `fn` must return a Promise,  so these two lines could be simply be:
````js
const promise = fn(batchObject);
return promise.then(result => {
...
````

File: src/adapters/IDB.js [hunk 7]
```diff
@@ -81,6 +87,62 @@ export default class IDB extends BaseAdapter {
...
+        .then(result => {
```
michielbdejong: Actually, `result` is not used in this function, so lines 118-120 can just be `return fn(batchObject).then(() => {`

src/adapters/IDB.js
```diff
@@ -81,6 +87,62 @@ export default class IDB extends BaseAdapter {
   }
 
   /**
+   * Batch operations.
+   *
+   * @param  {Function} fn The operation function, which must return a Promise.
+   * @return {Promise}
```
michielbdejong: Maybe explain that it returns a promise for `{operations, errors}`. What is the use of returning this data, is that for the unit tests?

src/collection.js
```diff
@@ -460,15 +462,31 @@ export default class Collection {
...
+          } else if (importResult.type === "skipped") {
+            skipped.push(importResult.data);
           }
-        }
+          return importResult;
+        });
+      }));
+    });
+    return batchResult
+      .then(({operations, errors}) => {
```
michielbdejong: Why not directly `return this.db.batch( ... ).then(({operations, errors}) => {` from line 467?

src/adapters/IDB.js
michielbdejong: Ah, found it on line 487, sorry.

test/adapters/IDB_test.js
```diff
@@ -19,6 +20,102 @@ describe("adapter.IDB", () => {
...
+              {id: 2, name: "bar"},
+            ]);
+        });
+
+        it("should perform different crud operations in order", () => {
+          return db.create({id: 1, name: "foo"})
+            .then(_ => db.batch(batch => {
+              batch.delete(1);
+              batch.create({id: 2, name: "bar"});
+              batch.update({id: 2, name: "baz"});
```
michielbdejong: This test would be even more fun with `id: 1` on all operations. :)

test/adapters/IDB_test.js
```diff
@@ -19,6 +20,102 @@ describe("adapter.IDB", () => {
...
+        it("should expose failing operation errors", () => {
+          return db.batch(batch => {
+            batch.create({id: 1, name: "foo"});
+            batch.create({id: 2, name: "bar"});
+            batch.create(1);
+            batch.create(2);
+          })
+            .then(res => {
+              expect(res.errors).to.have.length.of(2);
+              expect(res.errors[0].error.name).eql("DataError");
```
michielbdejong: We should really document a good list of different errors that may appear in `SyncResults.errors`, and what they mean.

test/adapters/IDB_test.js
```diff
@@ -19,6 +20,102 @@ describe("adapter.IDB", () => {
...
+        it("should not alter database on transaction error", () => {
+          return db.create({id: 1, name: "foo"})
+            .then(_ => db.batch(batch => {
+              batch.create({id: 2, name: "bar"});
+              batch.create({id: 3, name: "baz"});
+              batch.create({id: 1, name: "foo-dupe"}); // dupe
+              batch.create({id: 4, name: "qux"});
+            }))
+            .then(_ => db.list())
+            .should.become([{id: 1, name: "foo"}]);
```
michielbdejong: Nice! So transactions actually work! :)

test/adapters/IDB_test.js
```diff
@@ -19,6 +20,102 @@ describe("adapter.IDB", () => {
...
+        it("should allow performing a get operation within a batch", () => {
+          return db.create({id: 1, name: "foo"})
+            .then(_ => {
+              return db.batch(batch => {
+                batch.create({id: 2, name: "bar"});
+                return batch.get(1)
+                  .then(res => {
+                    expect(res).eql({id: 1, name: "foo"});
+                  });
+              });
```
michielbdejong: A simpler/nicer test might be:
````js
return batch.get(1)
  .then(res => {
    batch.update({id: 1, name: `Hello ${res.name}`});
});
````
And then expect 'Hello foo' afterwards.

File: test/collection_test.js [hunk 14]
```diff
@@ -827,9 +827,7 @@ describe("Collection", () => {
       it("should resolve with imported deletions", () => {
         return articles.pullChanges(result)
           .then(res => res.deleted)
-          .should.eventually.become([
-            {id: id_4}
-          ]);
+          .should.eventually.become([id_4]);
```
michielbdejong: Is this a breaking API change because we can no longer access the old record? Can't the API stay the same?

test/collection_test.js
```diff
@@ -764,7 +764,7 @@ describe("Collection", () => {
         {id: id_3, title: "art3"},   // to be created
         {id: id_4, deleted: true},   // to be deleted
         {id: id_6, deleted: true},   // remotely deleted & missing locally, skipped
-        {id: id_7, title: "art7-b"}, // remotely conflicting
+        {id: id_7, title: "art7-b"}, // remotely updated
```
michielbdejong: Did something actually change in this test or was this just an imprecise comment?

test/collection_test.js
```diff
@@ -869,20 +879,6 @@ describe("Collection", () => {
...
-          sandbox.stub(articles, "create")
-            .onCall(0).returns(Promise.reject(err1))
-            .onCall(1).returns(Promise.reject(err2));
-
-          return articles.pullChanges(result)
-          .then(res => res.errors)
-          .should.become([err1, err2]);
-        });
-      });
     });
```
michielbdejong: This test was moved to the 'should expose failing operation errors' test, right?

test/integration_test.js
```diff
@@ -27,7 +27,7 @@ describe("Integration tests", () => {
     // Keeping parent's environment is needed so that pserve's executable
     // can be found (with PATH) if KINTO_PSERVE_EXECUTABLE env variable was not provided.
     env = Object.assign({}, process.env, env);
-    server = spawn(PSERVE_EXECUTABLE, [KINTO_CONFIG], {env, detached: true});
+    server = spawn(PSERVE_EXECUTABLE, [KINTO_CONFIG], {env});
```
michielbdejong: Merge error?

michielbdejong:
Wow, awesome! :)

#### Remark 1:
We could refactor the 100 lines of individual operations code in https://github.com/Kinto/kinto.js/blob/172-bulk-operations/src/adapters/IDB.js#L145-L253 to:
````js
create(record) { return this.batch(trans => trans.create(record)); }
update(record) { return this.batch(trans => trans.update(record)); }
delete(id) { return this.batch(trans => trans.list(id)); }
get(id) {
  let result;
  return this.batch(trans => {
     result = trans.get(id);
  }).then(() => result);
});
list() {
 let result;
  return this.batch(trans => {
     result = trans.list();
  }).then(() => result);
});
````
#### Remark 2:
We could add a dummy `batch` method to the `BaseAdapter`, something like:
````js
batch(fn) {
  return fn(this);
}
````

File: src/adapters/IDB.js (hunk 2 above)
Comment (n1k0) [inline src/adapters/IDB.js]: Indeed. If we want 100% mirroring of the API and behavior, should I add this check to adapter's `#get()`, or should I remove the check from the batch one?

File: src/adapters/IDB.js (hunk 3 above)
Comment (n1k0) [inline src/adapters/IDB.js]: I'll go with `operationsFn`.

File: src/adapters/IDB.js (hunk 6 above)
Comment (n1k0) [inline src/adapters/IDB.js]: Actually there's no point at requiring a Promise to be returned. I'll leave this as optional in the code, especially as I don't clearly see what to do with the returned result value :)

Note: if you have ideas about it, I'm all ears though. I feel like it could be useful, but how?

File: src/adapters/IDB.js (hunk 7 above)
Comment (n1k0) [inline src/adapters/IDB.js]: See previous comment about that :)

File: src/adapters/IDB.js (hunk 8 above)
Comment (n1k0) [inline src/adapters/IDB.js]: I definitely feel we could v-carry data over the returned value, though I don't know where to put it… `{operations, errors, result}`?

File: src/collection.js (hunk 9 above)
Comment (n1k0) [inline src/collection.js]: That was a debugging leftover, will fix ;)

File: test/adapters/IDB_test.js (hunk 10 above)
Comment (n1k0) [inline test/adapters/IDB_test.js]: Well that would raise a transaction error, which is topic for another dedicated test in the suite.

File: test/adapters/IDB_test.js (hunk 11 above)
Comment (n1k0) [inline test/adapters/IDB_test.js]: Yeah, for now that mixes http and idb errors… We should probably type them or create ad-hoc sections. Feel free to file an issue about that.

File: test/adapters/IDB_test.js (hunk 12 above)
Comment (n1k0) [inline test/adapters/IDB_test.js]: Let's hope so :p

File: test/adapters/IDB_test.js (hunk 13 above)
Comment (n1k0) [inline test/adapters/IDB_test.js]: Love it!

File: test/collection_test.js (hunk 14 above)
Comment (n1k0) [inline test/collection_test.js]: I debated that; yes it's a breaking change but couldn't see the necessity of keying the id. Can definitely revert to the old behavior.

Comment (n1k0) [inline test/collection_test.js]: Reverting to old format implies:

```js
        operations.forEach(operation => {
          if (operation.type !== "void") {
            // operations can only be create(d), update(d), delete(d)
            const data = operation.type === "delete" ? {id: operation.data} : operation.data;
            syncResultObject.add(`${operation.type}d`, data);
          }
        });
```

Yuk :(

n1k0: After discussion over IRC, I'll keep it that way.

test/collection_test.js
n1k0: The comment was misleading; for that test, it's a classical update from the server (we're in the *“When no conflicts occured”* section of the tests ;)).

test/collection_test.js
n1k0: Yup.

test/integration_test.js
n1k0: Nope, detaching the server process was preventing to kill it automatically along the test process ;)

n1k0:
@michielbdejong did factorize create, update and delete to reuse batch() in 8e66c79884ddb30593da398ed15e41e1c71a37bd, thoughts?

The main issue is we can't handle errors the very same way, so I went with picking the only one present if any, which should be safe as we're doing a single operation (is it?).

n1k0:
This is getting better now we're leveraging batch function result. I'll probably refactor the pull/push collection flow to benefit from it. As always, comments, feedback, suggestion are welcome :)

src/collection.js
```diff
@@ -641,6 +672,7 @@ export default class Collection {
     if (strategy === Collection.strategy.MANUAL || result.conflicts.length === 0) {
       return Promise.resolve(result);
     }
+    // XXX: Ensure all updates are done within a single transaction
```
n1k0: Oops, missed that one. Commenting here so I don't forget about it.

michielbdejong:
This is currently throwing `TransactionInactiveError`s. To replicate:
Run `npm run demo`
Run `cd demo/ ; python -m SimpleHTTPServer 8001`
On http://localhost:8001/ add a task
You will see "Error: create() A request was placed against a transaction which is currently not active, or which is finished." in the console.

michielbdejong:
@n1k0 Note that my dist files are different from the ones you just pushed: https://github.com/Kinto/kinto.js/commit/b8f4245d11b734ffdb404c64c92b59beb54f8cef

n1k0:
> @n1k0 Note that my dist files are different from the ones you just pushed: b8f4245

Yeah, forgot to update them. Cannot reproduce the issue you describe with these latest versions, can you please retest using them?

n1k0:
I've just noticed that from my tests, listing store entries from within a transaction doesn't seem to work; sample test:

```js
        it("should allow listing from within batch", () => {
          return db.batch(batch => {
            batch.create({id: 1, name: "foo"});
            return batch.list();
          }).should.eventually.have.property("result").eql([
            {id: 1, name: "foo"}
          ]);
        });
```

Logging the flow steps and intermediate states makes it clear that the transaction `complete` event is triggered *before* that `list` promise is even fulfilled, leading in inconsistent results; here, the list returned is empty.

I'm having hard times figuring out if that's an expected behavior, of if I'm missing something obvious.

n1k0:
> it's even more surprising that get actually works

Just checked, tthe behavior is the same as with `list`: a batch#create call then a `batch#get` of it will fail as well.

Now thinking about it, transaction requests being *scheduled*, it makes sense they haven't been executed at the time the `list` promise is fulfilled, and I suspect this is actually expected behavior; though API might get highly confusing for end users, as I was as an implementer (ok it's late, but still). Sure, docs will probably help providing guidance with this, though I'm really challenging the idea of providing `batch#list` and friends, which will retrieve results unsynced with the current virtual data state within the transaction.

Thoughts?

michielbdejong:
Ah, the magic of auto-commit, gotta love it. :)

https://duckduckgo.com/?q=indexeddb+auto-commit&t=ffab
http://stackoverflow.com/questions/10484965/how-can-i-put-several-requests-in-one-transaction-in-indexeddb
http://stackoverflow.com/questions/27326698/indexeddb-transaction-auto-commit-behavior-in-edge-cases
http://w3c.github.io/IndexedDB/#dfn-commit

I think the problem is that once you start doing the first operation on a transaction, from that moment on it will try to find an opportunity to auto-commit before you can do the second operation. So make sure to always keep the callback chain alive until you're done. I expect (didn't test) it will work if you do:
````js
return batch.create({id: 1, name: "foo"}).then(() => batch.list());
````

n1k0:
>  I expect (didn't test) it will work if you do:
>
> ```js
> return batch.create({id: 1, name: "foo"}).then(() => batch.list());
> ```

Problem is that `batch.create` doesn't return a Promise. Tried to `Promise.all` all store requests (resolving on `req.onsuccess`) but it didn't work better.

michielbdejong:
What if you make `batch.create` return a promise from `store.add.onsuccess`?

n1k0:
> What if you make batch.create return a promise from store.add.onsuccess?

That's what I've just described, no more luck with that approach.

michielbdejong:
Weird! I'll have a look soon and see if I can work out what's going on.

n1k0:
I think I've found my way around by wrapping batch request in their very own promise, great result so far and solves the issue mentioned above; *lots* of refactoring now :)

n1k0:
@michielbdejong would love hearing feedback about the approach in 37c369b613d90c28a1993946677cadc2a8c32276

michielbdejong:
Cool! Will test & review on Monday.

n1k0:
For `Collection#batch()`, we'll need to think of a way to reuse the validation checks we use for `Collection#create` and friends. Will do on Monday.

michielbdejong:
Still getting an TransactionInactiveError :( Let's discuss on irc!

michielbdejong:
To reproduce, copy kinto.min.js from https://github.com/Kinto/kinto.js/commit/66fc4a2af2582d15a189ca38f7f69b37fdd559c3 to apps/sync/js/ext/ on https://github.com/michielbdejong/gaia/tree/1200539-sync-app-integration-tests/apps/sync#trying-out-the-sync-app

n1k0:
r=? @leplatrem

n1k0:
> For `Collection#batch()`, we'll need to think of a way to reuse the validation checks we use for `Collection#create` and friends. Will do on Monday.

I suggest we postpone landing generic validation after this patch eventually lands; current diff is huge, and adding more to it seems wrong. I've rebased removing the validation stuff, will move to update documentation bits.

@michielbdejong keep me informed if the current branch still breaks for you, and how.

n1k0:
For supporting a default implementation of `batch()` at the `BaseAdapter` level, the main issue us that it should return the same result object as its child implementations, which would be possible but a little hard to achieve at this point; could we consider spawning a new issue about this, and do we really want it?

michielbdejong:
BaseAdapter doesn't strictly need `batch()`, but then `LocalStorage` adapter does need its own (dummy) implementation of it, right?

n1k0:
> then LocalStorage adapter does need its own (dummy) implementation of it, right?

Sure, I just wanted to ensure that approach would work first. Will implement batch for LocalStorage and add supplementary tests.

n1k0:
Fixed by #303. So long, branch.

PR #167, opened 2015-09-21
Upgraded Kinto to 1.4.0.

n1k0:
We should run our integration tests against the very [latest stable](https://github.com/Kinto/kinto/releases/tag/1.4.0) release of Kinto.

r=? @Natim @leplatrem @ametaireau

PR #183, opened 2015-09-29
Do not merge (just testing Travis with this PR)

michielbdejong:


michielbdejong:
Looking at https://travis-ci.org/Kinto/kinto.js/jobs/82710400 I see `.env/bin/pip install kinto==1.4.0` even though https://github.com/michielbdejong/kinto.js/blob/8f357ffd1512f402bbac4cdd5dbd82fba47783d7/.travis.yml#L12 in this PR has version 1.3.1.

If you think about it, it makes sense that Travis looks at master (not at the branch being merged) for its `.travis.yml` file, because it's safer (in master, you specify the authoritative conditions for merging into master).

What is still weird is that [all checks passed](https://github.com/Kinto/kinto.js/pull/167) on `dc8916bca32`, and they ran with 1.4.0.

I'll see if anything changed after the Kinto upgrade which might have caused Travis to start failing.

michielbdejong:
Maybe something changed in Kinto on Wednesday (Travis will `pip install` latest Kinto in each run, I guess). Although I guess the version numbers should take care of that, Kinto 1.4.0 should always be Kinto 1.4.0...

Also did not see any announcements from Travis about things changing on their side last week.

Will investigate further.

michielbdejong:
OK, I think I found the problem: When I run `.env/bin/pip install kinto==1.3.0 --upgrade` locally in master, `.env/bin/pserve test/kinto.ini` it works, but when I do the same with `kinto==1.4.0`, I see this error:
````
[...]
File "/Users/Michiel/repos/kinto.js/.env/lib/python2.7/site-packages/kinto/views/groups.py", line 6, in <module>
    from cliquet.authorization import get_object_id
ImportError: cannot import name get_object_id
````
@n1k0 do you see the same behavior on your localhost?

michielbdejong:
Confirmed, it's a problem with Kinto 1.4.0, it's fixed in Kinto 1.5.0.

PR #184, opened 2015-09-29
Upgraded Kinto to 1.5.0.

n1k0:
This is most likely to fix our broken builds, because Kinto 1.4.0 has troubles starting from our test env, while 1.5.0 seems to have fixed the situation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_install:
 - export PATH=$HOME/.local/bin:$PATH
 - pip install --user `whoami` virtualenv
 - virtualenv .env
-- .env/bin/pip install kinto==1.4.0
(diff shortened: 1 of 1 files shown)


Natim:
Is because the config is not up-to-date?

n1k0:
> Is because the config is not up-to-date?

Error we get with 1.4.0 is `ImportError: cannot import name get_object_id`. I was about to file a proper issue for Kinto.

Natim:
Find the problem! While Kinto 1.4.0 is tagging the compatible version of cliquet, the tagging is wrong: ``'cliquet>=2.6.2,<=2.7'`` it should be ``<2.7``

n1k0:
@Natim filed https://github.com/Kinto/kinto/issues/205 about it.

Issue #190, opened 2015-10-05
Remove unneeded external dependencies


mozmark:
There are a number of libraries and adapters which are loaded by Kinto that will never be used in some circumstances. E.g. polyfills such as isomorphic fetch are not required where a real fetch implementation exists, kinto.js running as privileged JS will not need IDB or local storage adapters.

Can we re-work so such dependencies are injected in, allowing separate distributed scripts to be prepared for different intended uses?

n1k0:
As per IRC discussion

> mgoodwin: kinto.js seems to include Buffer - is that just because it's a dep. of deep-eql? If so, might you be OK with us shipping a light version version of deep-eql that doesn't use it?
> mgoodwin: (about 50% of the outout file is Buffer related - which seems excessive. Also, it does funky things with mutating object prototypes which is bad for perf)

n1k0:
At first glance, [browserify's `browser` field](https://github.com/substack/browserify-handbook#browser-field) is what we need here; but:

> The browser field only applies to the current package. Any mappings you put will not propagate down to its dependencies or up to its dependents.

So we'll have to ship our own version of `deep-eql`, I'm afraid.

PR #199, opened 2015-10-07
Inject dependencies

mozmark:
Trying again (see comment on https://github.com/Kinto/kinto.js/pull/197 re. issue #190 for context).


src/adapters/IDB.js
```diff
@@ -238,4 +235,32 @@ export default class IDB extends BaseAdapter {
...
   }
+
+  /**
+   * In FakeIndexedDB, symbols are exposed using ``FDB`` prefixes in names.
+   * This piece of code will register them with the same names as native API,
+   * only if indexedDB is not already available.
+   *
+   * @param {Object} obj The object to check and atttach indexedDB symbols to.
+   */
+  static attachFakeIDBSymbolsTo(obj) {
```
n1k0: Now I think about it, the fake IDB stuff is only needed for test code ran in a nodejs environment, so we should probably move this to `test_utils.js` and call that function at test startup time.

n1k0: Created #201 about this.

src/api.js
```diff
@@ -70,7 +69,7 @@ export default class Api {
      * The even emitter instance.
      * @type {EventEmitter}
      */
-    this.events = options.events || new EventEmitter();
```
n1k0: I really like dropping instantiation from within constructors :)

src/KintoBase.js
```diff
@@ -0,0 +1,99 @@
...
+   * @param  {Object} options The options object.
+   */
+  constructor(options={}) {
+    const defaults = {
+      bucket: DEFAULT_BUCKET_NAME,
+      remote: DEFAULT_REMOTE,
+    };
+    this._options = Object.assign(defaults, options);
+    if (!this._options.adapter) {
+      throw new Error("No adapter provided");
```
n1k0: We should add a test for this at the `KintoBase` level (right now we only have a test for the Collection).

src/index.js
```diff
@@ -31,76 +26,13 @@ export default class Kinto {
...
-    const bucket = this._options.bucket;
-    return new Collection(bucket, collName, this._api, {
-      events:              this._options.events,
-      adapter:             this._options.adapter,
-      dbPrefix:            this._options.dbPrefix,
-      idSchema:            options.idSchema,
-      remoteTransformers:  options.remoteTransformers
-    });
+    let expandedOptions = Object.assign(defaults, options);
+    super(expandedOptions);
```
n1k0: Nit: maybe directly 

```js
super(Object.assign({}, defaults, options));
```

Note that I'm also suggesting we pass it a copy instead of mutating the `default` object; not that it's impacting here, just I consider good practice to generally avoid mutations.

src/api.js
```diff
@@ -70,7 +69,7 @@ export default class Api {
...
+    this.events = options.events;
```
n1k0: As events is now a string requirements, we should probably throw when not provided — as we did for `KintoBase` and `Collection`. Also, we could consider this not being an *option* anymore, but a required argument; though this would be an important API change… Thoughts?

src/index.js
```diff
@@ -31,76 +26,13 @@ export default class Kinto {
...
-   *
-   * @param  {Object} options The options object.
-   */
   constructor(options={}) {
     const defaults = {
       adapter: Kinto.adapters.IDB,
-      bucket: DEFAULT_BUCKET_NAME,
-      events: new EventEmitter(),
-      remote: DEFAULT_REMOTE,
+      events: new EventEmitter()
```
n1k0: Maybe we could allow users to pass in their own `events` instance? The previous jsdoc was misleading btw.

n1k0: Scratch all this, I misread the code… the `events` option is actually supported, let's keep it that way.

n1k0:
Looks good! Now the question is, how do you trigger distinctly targeted builds? And we should document this too.

For coverage, have you tried running `npm run test-cover-html` to investigate which part of the code aren't covered? Istanbul html output is a little more convenient to read than the ones provided online by coveralls imoe.

src/api.js
mozmark: Personally, I'd prefer the API change since it'd be more obvious it's expected. I don't mind either way, though. Let me know what you prefer and I'll do it. /cc @michielbdejong

michielbdejong: I don't consider a change to the API of `Api` really a Kinto API change. ;) It just means that instead of providing the default in `api.js`, we provide it in `index.js`, which is indeed kind of nice because it means the constructor has less active code.

(unrelated to this PR) I think the code comment `- {String}       events  The HTTP request mode.` on line 40/39 should I think be `- {String}       requestMode  The HTTP request mode.`

n1k0: > Personally, I'd prefer the API change
> I don't consider a change to the API of Api really a Kinto API change

Let's update this `Api` API then!

src/http.js
```diff
@@ -33,7 +32,7 @@ export default class HTTP {
      * The eventt emitter instance.
      * @type {EventEmitter}
      */
-    this.events = options.events || new EventEmitter();
+    this.events = options.events;
```
n1k0: Note that the same as for `Api` applies here, `events` should become a required argument.

src/KintoBase.js
mozmark: I've added a kintobase-test.js file to test/ which covers this.

mozmark:
r=? @n1k0

mozmark:
The 'how do you trigger distinctly targeted builds?' question will be addressed in my next PR. We simply add a new dist-fx which uses a different entry point (and explicitly excludes the parts we don't want).

The FX specific implementations (the utils, moz-storage.jsm) can live outside of kinto.js - though the storage jsm can be included if we can have a way of omitting it from e.g. the coverage results.

n1k0:
Looks good, though there's a warning triggered from the test:

```
(node) warning: possible EventEmitter memory leak detected. 11 backoff listeners added. Use emitter.setMaxListeners() to increase limit.
Trace
    at EventEmitter.addListener (events.js:179:15)
    at Api._registerHTTPEvents (/Users/niko/Sites/kinto.js/src/api.js:114:17)
    at new Api (/Users/niko/Sites/kinto.js/src/api.js:93:10)
    at Context.<anonymous> (/Users/niko/Sites/kinto.js/test/api_test.js:26:11)
    at callFn (/Users/niko/Sites/kinto.js/node_modules/mocha/lib/runnable.js:286:21)
    at Hook.Runnable.run (/Users/niko/Sites/kinto.js/node_modules/mocha/lib/runnable.js:279:7)
    at next (/Users/niko/Sites/kinto.js/node_modules/mocha/lib/runner.js:297:10)
    at Immediate._onImmediate (/Users/niko/Sites/kinto.js/node_modules/mocha/lib/runner.js:319:5)
    at processImmediate [as _immediateCallback] (timers.js:367:17)
```

I'll investigate this on monday.

src/KintoBase.js
```diff
@@ -0,0 +1,99 @@
...
+    }
+    this._api = new Api(this._options.remote,
+                        this._options.events, {
+                          headers:     this._options.headers,
+                          requestMode: this._options.requestMode,
+                        });
```
n1k0: Nit: may I suggest

```js
const {remote, events, headers, requestMode} = this._options;
this._api = new Api(remote, events, {headers, requestMode});
```

n1k0:
Pending on build results, r+

Thanks!

PR #201, opened 2015-10-08
Moved fake indexedDB symbol imports to test logic.

n1k0:
r=? @mozmark


mozmark:
Looks good to me. r=me

PR #207, opened 2015-10-15
Fix snippet doc index (create instead of add)

leplatrem:
@n1k0 r?


ametaireau:
r+

n1k0:
Wow, such mistake, so great fix.

PR #209, opened 2015-10-16
[WIP] Force reupload when server was flushed (fixes #178)

### LINKED ISSUE #178: How does Kinto.js detect server was flushed and reupload the local database.
We have this use case for Syncto when a remote server could disappear and the new one we are assigned to is empty.

This is similar to what we also have with kinto.dev that is flushed every night.

This is related to Question #44 and this might be the first strategy we want to implement.

Detect that Kinto server was flushed and re-upload everything.

leplatrem:
As described in #178, if the client was already synced once, and the server timestamp was changed, but the pull for changes returns no result, then it means the remote server was flushed (or changed by another one).

In this case, as a first implementation, we recreate everything.

Thoughts @Natim @n1k0 ?

* [x] Basic implementation
* [ ] Confirm strategy/design (i.e. `serverEmpty` attribute)
* [x] Write test for `sync()`
* [ ] Unit test for `pullChanges`
* [ ] Unit test for `gatherLocalChanges`
* [ ] Fonctional test with call to `/__flush__` between two syncs

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -21,6 +21,7 @@ export class SyncResultObject {
     return {
       ok:           true,
       lastModified: null,
+      serverEmpty:  false,
       errors:       [],
       created:      [],
       updated:      [],
@@ -554,9 +555,17 @@ export default class Collection {
    *
    * @return {Object}
    */
-  gatherLocalChanges() {
+  gatherLocalChanges(options={forceAll: false}) {
+    var forceSync = Promise.resolve();
+    if (options.forceAll) {
+      forceSync = this.list({filters: {_status: "synced"}})
+        .then(synced => Promise.all(synced.data.map(r => {
+          return this.update(Object.assign(r, {lastModified: null}));
+        })));
+    }
     var _toDelete;
-    return this.list({}, {includeDeleted: true})
+    return forceSync
+      .then(_ => this.list({}, {includeDeleted: true}))
       .then(res => {
         return res.data.reduce((acc, record) => {
           if (record._status === "deleted" && !record.last_modified) {
@@ -600,6 +609,16 @@ export default class Collection {
       lastModified: options.lastModified,
       headers: options.headers
     })
+      // Check if server is empty (flushed or changed).
+      .then(changes => {
+        const firstSync =           (!this.lastModified);
+        const lastModifiedUpdated = (changes.lastModified !== this.lastModified);
+        const emptyCollection =     (changes.changes.length === 0);
+        syncResultObject.serverEmpty = (!firstSync &&
+                                        lastModifiedUpdated &&
+                                        emptyCollection);
+        return changes;
+      })
       // Reflect these changes locally
       .then(changes => this.importChanges(syncResultObject, changes))
       // Handle conflicts, if any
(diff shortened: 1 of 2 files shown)


src/collection.js
```diff
@@ -554,9 +555,20 @@ export default class Collection {
...
    */
-  gatherLocalChanges() {
+  gatherLocalChanges(options={forceAll: false}) {
+    var forceSync = Promise.resolve();
+    if (options.forceAll) {
+      forceSync = this.list({filter: {_status: "synced"}})
+        .then(synced => {
+          synced.map(r => {
+            r._status = "created";
+            delete r.lastModified;
```
Natim: I wish we could keep it but the Kinto server won't let us set it on creation anyway so it is probably better that way.

src/collection.js
```diff
@@ -554,9 +555,20 @@ export default class Collection {
...
+          });
```
n1k0: We should rather keep using `Object.assign` to update objects, for safety, and consistency.

Natim: Can you remove the value using assign null for instance?

src/collection.js
```diff
@@ -600,6 +612,16 @@ export default class Collection {
       lastModified: options.lastModified,
       headers: options.headers
     })
+      // Check if server is empty (flushed or changed).
+      .then(changes => {
+        const firstSync =           (!this.lastModified);
+        const lastModifiedUpdated = (changes.lastModified !== this.lastModified);
+        const emptyCollection =     (changes.changes.length === 0);
```
n1k0: Nit: no need to wrap these tests within parenthesis.

n1k0: From what I read, it seems we assume a collection is empty if no changes are retrieved from the server; might be worth a comment to avoid confusion.

src/collection.js
n1k0: I'm also confused with a `map` operation not being returned by the function. Please explain :)

src/collection.js
```diff
@@ -600,6 +612,16 @@ export default class Collection {
...
+        syncResultObject.serverEmpty = (!firstSync &&
+                                        lastModifiedUpdated &&
+                                        emptyCollection);
```
n1k0: How about a `flushed` Boolean property instead?

n1k0: Nit: Also, how about adding a `markFlushed` method to `SyncResultObject`? This would allow eventually performing more checks/updates at the result object level. (unsure about that)

michielbdejong:
In the FxOS Sync app we would like to be able to use Kinto.js as a read-only client for Syncto. This patch means Kinto.js might write to the server unpredictably, right? Although we will also configure the Syncto server to disallow writes, maybe we could define an option `Collection#sync({ readOnly: true })`?

n1k0:
Latest discussion with @leplatrem highlighted just that; server data overriding/refilling should be done explicitely.

That means now a flushed server detected will make `#sync()` to reject with an appropriatte error, and we'll provide a method users can call explicitely to reupload their local database to the server.

Would that work for you?

n1k0:
Superseded by #214.

michielbdejong:
> make #sync() to reject with an appropriatte error, and we'll provide a method users can call explicitely call to reupload their local database to the server.

Yes, that would work perfectly for our app. Thanks!

PR #210, opened 2015-10-16
fetchChangesSince should not throw if an empty 200 response is rece…

ferjm:
…ived

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -186,7 +186,7 @@ export default class Api {
         const etag = res.headers.get("ETag");  // e.g. '"42"'
         return {
           lastModified: etag ? parseInt(unquote(etag), 10) : options.lastModified,
-          changes: res.json.data
+          changes: res.json && res.json.data ? res.json.data : []
         };
       });
   }
--- a/test/api_test.js
+++ b/test/api_test.js
@@ -261,6 +261,17 @@ describe("Api", () => {
           });
       });
 
+      it("should resolve with no changes if HTTP 200 is received with no data", () => {
+        sandbox.stub(root, "fetch").returns(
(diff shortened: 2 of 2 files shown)


ferjm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1214661

n1k0:
Tthe patch looks good, though I'm puzzled as how this could ever happen (empty 200 response on fetch). Hints? @Natim @leplatrem @ametaireau

Natim:
@n1k0 this is related to #209

When the server was flushed, you can end-up with a 200 and an empty response with If-None-Match header.

Also it might be possible to get that with Syncto I guess.

n1k0:
> When the server was flushed, you can end-up with a 200 and an empty response

You mean no JSON *at all*? That sounds like a bug to me. Shouldn't a 200 should always return a JSON body?

Natim:
> You mean no JSON at all?

No it will be a ``{"data": []}`` @ferjm I would be interested in having the query that leads to this response. As @n1k0 said it is probably a bug.

n1k0:
So what's the status of this? Is it a server bug, or should we expect possibly empty 200 responses from the server?

leplatrem:
In the Kinto protocol, on the list of records, we will never return an empty dict with a 200 response.

However in SyncTo, there is this subtle condition that proves that it can actually happen when the Sync server returns no record : https://github.com/mozilla-services/syncto/blob/master/syncto/views/collection.py#L80-L81

This is thus a SyncTo server bug ! Good catch @ferjm !

n1k0:
> This is thus a SyncTo server bug

Hence closing, thanks.

PR #211, opened 2015-10-16
Fixes #190: Avoid exporting Buffer to dist files.

### LINKED ISSUE #190: Remove unneeded external dependencies

n1k0:
Refs #190. We should get rid of `deep-eql` which carries a dependency to `Buffer` we can't exclude from dist files; we should switch to using nodejs' own `assert#deepEqual` which doesn't have such heavy dependencies.


n1k0:
>  1,355 additions, 2,465 deletions not shown 

That's much better. r=? @mozmark

mozmark:
This looks good so far - I'm going to do some tests on using this in Firefox to ensure the assert library works there (no reason I can see that it shouldn't).

n1k0:
@mozmark ping, could you please confirm at patch works for you so I can land this?

mozmark:
I can confirm this works (my apologies for delaying you).

r=me

PR #217, opened 2015-10-20
Fixes #203: Removed dist files.

### LINKED ISSUE #203: Don't version dist files, publish & host them.
Building and versioning dist files is a real pain; we should rather build and host them for each release on github pages along the demo I think. Related to #202.

n1k0:
Refs #203. Current demo at http://kinto.github.io/kinto.js uses the new public asset links.

r=? @leplatrem


demo/index.html
```diff
@@ -37,7 +37,7 @@
       </li>
     </template>
   </div>
-  <script src="kinto.dev.js"></script>
+  <script src="http://mozilla-services.github.io/kinto.js/kinto.dev.js"></script>
```
Natim: Shall we use https for that?

Natim:
It looks good to me Also, I am in favour of the http/https switch.

demo/index.html
n1k0: > Shall we use https for that?

Good point; even better, using `//` for automatic current protocol conservation :)

n1k0:
Hmm for some reason the hosted demo doesn't seem to update with latest changes http://mozilla-services.github.io/kinto.js/; I'm holding down merging for now.

n1k0:
gh-pages branch has been properly updated, but hosted demo doesn't reflect it oO https://github.com/Kinto/kinto.js/blob/gh-pages/index.html

Cache issue? Delayed propagation?

Natim:
Yes it could be the cache. One way to make sure it is, is to load the js file url and force refresh it.

n1k0:
> One way to make sure it is, is to load the js file url

The problem isn't with the js file url but with the index.html one… purging browser cache doesn't fix it, and curl shows old version at its url. Weird.

n1k0:
Woookay I'm just realizing we've been using a wrong demo URL since we've moved the project under the Kinto org. The correct URL is http://kinto.github.io/kinto.js/. Gonna update all the links to it.

n1k0:
Requesting further validation with latest change; r=? @Natim @leplatrem

leplatrem:
Gave a quick look at everything and it looks sane to me

Natim:
r+

PR #218, opened 2015-10-21
Fixes #215: Added internal links to esdoc.

### LINKED ISSUE #215: Investigate how to link to foreign symbols from within esdoc
Something like this:

```
/**
 * A link to {#sync()}
 */
```

n1k0:
r=? @leplatrem

### DIFF
--- a/src/KintoBase.js
+++ b/src/KintoBase.js
@@ -12,8 +12,8 @@ const DEFAULT_REMOTE = "http://localhost:8888/v1";
  */
 export default class KintoBase {
   /**
-   * Provides a public access to the base adapter classes. Users can create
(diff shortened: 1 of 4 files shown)


Natim:
lgtm.

leplatrem:
We don't want them in the narrative docs ? `#justWondering`

r+ !

n1k0:
> We don't want them in the narrative docs ?

What do you mean, [these](http://kintojs.readthedocs.org/en/latest/)? Public symbols are already linked to the API docs there, already.

n1k0:
Note that if we move to host the narrative docs to esdoc, I suspect we'll be able to leverage their link syntax directly.

n1k0:
I'm gonna land this.

Issue #177, opened 2015-09-24
Expose transactions from Adapters


michielbdejong:
In order to make better use of IndexedDB transactions (one transaction per operation is superslow), a first step would be exposing transactions from Adapters. Severall syntax ideas are possible, @n1k0 and I were already brainstorming on irc. Here is just one example syntax (sorry for the raw braindump):

Instead of:
````js
Promise.all([
  adapter.create({ id: 0, data: 'foo' }),
  adapter.create({ id: 1, data: 'bar' })
]).then(results => console.log(results), err => console.error(err));
````
it could be:
````js
adapter.batch([
  'create', { id: 0, data: 'foo' },
  'create', { id: 1, data: 'bar' }
]).then(results => console.log(results), err => console.error(err));
````

The way to implement it in current IDB adapter could be something like:
* call `this.open`,
* call `prepare`,
* execute all the operations, using the transaction and store objects from `prepare`,
* Pass the results through `Promise.all`.

We have to look at what happens if different arguments need to be passed to `prepare`, maybe operations have to be grouped per `store`, and then `Promise.all` all the results from all the `store`s.

Since the current operations (`create`, `update`, etc.) in the IDB adapter already call open & prepare, they could be proxies to `_create`, `_update`, etc. functions, that execute the actual actions. So a bit of refactoring there, probably.

The `BaseAdapters` can probably implement this easily using just `Promise.all`, to keep other adapters compatible.

**Comment by n1k0** [comment]
> We have to look at what happens if different arguments need to be passed to prepare

How about using an higher order function?:

```js
const batch = adapter.transaction("storeName");
batch([
  'create', { id: 0, data: 'foo' },
  'create', { id: 1, data: 'bar' }
]).then(results => console.log(results), err => console.error(err));
```

Still I'm not super easy with the `[action, data, action, data, …]` mixed array. How about `.batch()` being smart enough to alow something like this?

```js
const batch = adapter.batch("storeName");
batch([
  batch.create({ id: 0, data: 'foo' }),
  batch.create({ id: 1, data: 'bar' }),
  batch.delete(1),
  batch.update({ id: 0, data: 'baz' })
])
  .then(results => console.log(results))
  .catch(err => console.error(err));
```

Here `batch.create` would simply call `store.add` with the provided data, still using the initiated transaction. Am I talking crazy here? :)

n1k0:
[...]

PR #219, opened 2015-10-21
212 firefox entry point

mozmark:
This gives us everything we need to do to build the jsm from kinto.js.


package.json
```diff
@@ -9,6 +9,7 @@
     "dist": "npm run dist-dev && npm run dist-prod",
     "dist-dev": "browserify -s Kinto -d -e src/index.js -o demo/kinto.dev.js",
     "dist-prod": "browserify -s Kinto -g uglifyify -e src/index.js -o demo/kinto.min.js",
+    "dist-fx": "browserify -s loadKinto -d -e fx-src/index.js -o dist/temp.jsm && cp fx-src/jsm_prefix.js dist/moz-kinto-client.js && cat dist/temp.jsm | grep -v sourceMappingURL >> dist/moz-kinto-client.js && rm dist/temp.jsm",
```
mozmark: Disturbingly, this is less ugly than all of the other options. We'll want to revisit this once we're happy that the es6 super support that landed in https://bugzilla.mozilla.org/show_bug.cgi?id=1066239 works with what we've done, but for now this works.

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,481 @@
...
+              reject(new Error("query cancelled or aborted"));
+            } else {
+              debug("last modified result was "+result);
+              resolve(result);
+            }
+            complete();
+          }
+        });
+      });
+    }.bind(this));
```
n1k0: Nit: Using arrow function automatic binding might avoid us this.

fx-src/index.js
```diff
@@ -0,0 +1,36 @@
+"use strict";
+
+const Cu = Components.utils;
+
+import BaseAdapter from "../src/adapters/base";
+import KintoBase from "../src/KintoBase";
+import FirefoxAdapter from "./FirefoxStorage";
+
+export default function loadKinto() {
+  let { EventEmitter } = Cu.import("resource://gre/modules/devtools/shared/event-emitter.js", {});
```
n1k0: Nit: Why `let` and not `const`? This is unlikely to be overridden, ever.

mozmark: because I initially wrote it in a scratchpad (where const gets annoying) and forgot to change it! Will fix

package.json
n1k0: Several notes here:

- The `-d` option exports source maps; remove it and you don't need to `grep -v` them anymore;
- The `dist` folder has been removed along #217, so we shouldn't use it; or at least making sure to create the folder — or a temporary one — first;
- I can't where does the `jsm_prefix.jsm` file come from; hints? Shouldn't it be `temp.js`?

n1k0:
As per IRC discussion, can't really tell for the adapter part. Some comments to address regarding the build command line. Otherwise looks good.

mozmark:
I addressed feedback: r? @n1k0

mozmark:
And r? @rnewman

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,484 @@
+import BaseAdapter from "../src/adapters/base";
```
rnewman: License block.

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,484 @@
...
+
+const { interfaces: Ci, results: Cr, utils: Cu } = Components;
+
+Cu.import("resource://gre/modules/Services.jsm");
+Cu.import("resource://gre/modules/FileUtils.jsm");
+
+function debug(message) {
+  // TODO: Read a pref or something to work out whether to emit debug
+  // messages
+  dump("FirefoxStorage "+message+"\n\n");
```
rnewman: Nit: always spaces around operators (`"foo" + bar + "baz"`).

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,484 @@
+import BaseAdapter from "../src/adapters/base";
+
+const { interfaces: Ci, results: Cr, utils: Cu } = Components;
+
+Cu.import("resource://gre/modules/Services.jsm");
+Cu.import("resource://gre/modules/FileUtils.jsm");
+
+function debug(message) {
```
rnewman: Consider what you want to do here other than use `dump`. There's `Log.jsm`, there's plain ol' `console.log`, …

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,484 @@
...
+  }
+
+  getName() {
+    return this.dbname;
+  }
+
+  getConnection() {
+    if (!this._dbconn) {
+      try {
+        this._dbconn = Services.storage.openDatabase(this.file);
```
rnewman: Unless you have a really, really good reason, you should be using Sqlite.jsm instead of Storage. In this review I won't look for Storage footguns.

https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/Sqlite.jsm
http://gregoryszorc.com/blog/2013/04/14/sqlite.jsm---sqlite-done-betterer/

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,484 @@
...
+
+// A way to handle our storage connection and storage statements
+class KintoStorage {
+  constructor(dbname) {
+    this.dbname = dbname;
+    this.file = FileUtils.getFile("ProfD", ["kinto.sqlite"]);
+    this.statements = {};
+  }
+
+  getStatement(statementString) {
```
rnewman: You can rip this out entirely if you use Sqlite.jsm.

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,484 @@
...
+export default class FirefoxAdapter extends BaseAdapter {
+  constructor(dbname) {
+    super();
+    this.operations = [];
+    this.busy = false;
+
+
+    // attempt creation - we don't need to use a transaction here since our
+    // operation queue means these will happen before other operations.
+    const statements = ["CREATE TABLE IF NOT EXISTS collection_metadata (collection_name TEXT PRIMARY KEY, last_modified INTEGER) WITHOUT ROWID;",
```
rnewman: You should be using `getSchemaVersion` and versioning your schema.

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,484 @@
...
+    debug("executeOperation");
+    if (!this.busy) {
+      debug("Executing operation now");
+      this.busy = true;
+      function executeNextOperation() {
+        debug("executeNextOperation");
+        const next = this.operations.shift();
+        if (next) {
+          debug("executing queued operation");
+          next(this.kintoStorage, executeNextOperation.bind(this));
```
rnewman: Can you explain why you're manually serializing operations here?

You should also think really hard about whether to use promises and Tasks, a la the second example here:

https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/Sqlite.jsm#Open_Execute_Close

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,484 @@
...
+        const statement = kintoStorage.getStatement("DELETE FROM collection_data WHERE collection_name = :collection_name;");
+        statement.params.collection_name = kintoStorage.getName();
+
+        // execute the statement
+        statement.executeAsync({
+          handleResult: function(aResultSet) {
+            debug("A result set was not expected");
+          },
+
+          handleError: function(aError) {
```
rnewman: This all gets much less painful when you use Sqlite.jsm, which returns promises.

fx-src/index.js
```diff
@@ -0,0 +1,36 @@
+"use strict";
```
rnewman: License block.

fx-src/index.js
```diff
@@ -0,0 +1,36 @@
...
+
+const Cu = Components.utils;
+
+import BaseAdapter from "../src/adapters/base";
+import KintoBase from "../src/KintoBase";
+import FirefoxAdapter from "./FirefoxStorage";
+
+export default function loadKinto() {
```
rnewman: Man, ES6 is crazy.

fx-src/jsm_prefix.js
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2015 - Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
rnewman: Really, we're Apache licensing rather than MPL2.0? Did someone in legal, or someone who knows more than me (Gerv?) already comment on this?

rnewman:
It looks like about 95% of the storage code will just go away if you use Sqlite.jsm — it does all the sane stuff like WALs, it's all safe and async, and it gives you promises which Kinto expects. So…

fx-src/index.js
mozmark: @rnewman Apache2 is fine: https://www.mozilla.org/en-US/MPL/2.0/FAQ/ - see Q13

MPL2 is my default, but kinto was Apache2 before I showed up..

mozmark:
@rnewman I addressed your feedback. r?

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,213 @@
...
+Components.utils.import("resource://gre/modules/Sqlite.jsm");
+Components.utils.import("resource://gre/modules/Task.jsm");
+
+const statements = {
+  "createCollectionData": "\
+    CREATE TABLE collection_data (\
+      collection_name TEXT, \
+      record_id TEXT, \
+      record TEXT\
+    );",
```
n1k0: Nit: You may want to use ES6 template string for these.

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,213 @@
...
+      finally {
+        // ensure the connection is closed
+        connection.close();
+      }
+    });
+  }
+
+  clear() {
+    const params = {collection_name: this.dbname};
+    return this._doTransaction(statements["clearData"], params);
```
n1k0: Nit: Why not `statements.clearData`?

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,213 @@
...
+
+      });
+    }
+
+  delete(id) {
+    const params = {
+      collection_name: this.dbname,
+      record_id: id,
+    };
+    return this._doTransaction(statements["deleteData"], params);
```
n1k0: The Collection API expects `delete(id)` to resolve with the id.

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,213 @@
...
+      const params = {
+        collection_name: this.dbname,
+        record_id: id,
+      };
+      return this._doTransaction(statements["getRecord"], params).then(result => {
+        if (result.length == 0) {
+          return;
+        }
+        return JSON.parse(result[0].getResultByName("record"));
+
```
n1k0: Nit: Empty line.

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,213 @@
...
+      return records;
+    });
+  }
+
+  saveLastModified(lastModified) {
+    const params = {
+      collection_name: this.dbname,
+      last_modified: lastModified,
+    };
+    return this._doTransaction(statements["saveLastModified"], params);
```
n1k0: The Collection API expects `saveLastModified` to resolve with saved `lastModified` value, as an int or `null`:

```js
var value = parseInt(lastModified, 10) || null;
```

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,213 @@
...
+  "getLastModified": "\
+    SELECT last_modified \
+      FROM collection_metadata \
+        WHERE collection_name = :collection_name",
+  "getRecord": "\
+    SELECT record \
+      FROM collection_data \
+        WHERE collection_name = :collection_name \
+        AND record_id = :record_id",
+  "listRecords": "\
```
n1k0: Nit: Maybe separating statements by an empty line would improve legibility a bit.

n1k0:
I must admit this looks much better :) Leaving @rnewman doing the real review ;)

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,227 @@
...
+    CREATE UNIQUE INDEX unique_collection_record
+      ON collection_data(collection_name, record_id);`,
+
+  "clearData": `
+    DELETE FROM collection_data
+      WHERE collection_name = :collection_name;`,
+
+  "createData": `
+    INSERT INTO collection_data (collection_name, record_id, record)
+      VALUES (:collection_name, :record_id, :record)`,
```
rnewman: You have trailing `;` in some of your queries but not others. Be consistent.

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,227 @@
...
+          let schema = yield (connection.getSchemaVersion());
+
+          if (schema == 0) {
+
+            for (let statementName of createStatements) {
+              yield (connection.execute(statements[statementName]));
+            }
+
+            yield (connection.setSchemaVersion(1));
+          } else if (schema != 1) {
```
rnewman: Extract `const currentSchemaVersion = 1;`.

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,227 @@
...
+      let connection = yield self._doOpenConnection();
+
+      // ensure we're initialized
+      yield self.init;
+
+      //execute the transaction
+      try {
+        return yield connection.executeCached(statement, params);
+      }
+      finally {
```
rnewman: Cuddle.

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,227 @@
...
+  constructor(dbname) {
+    super();
+    this.dbname = dbname;
+
+    // make a promise to initialize
+    const self = this;
+    this.init = Task.spawn(function* () {
+      let connection;
+      try {
+        connection = yield self._doOpenConnection();
```
rnewman: If you move this outside the `try`, you don't need to check whether it exists inside the `finally`.

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,227 @@
...
+
+      // ensure we're initialized
+      yield self.init;
+
+      //execute the transaction
+      try {
+        return yield connection.executeCached(statement, params);
+      }
+      finally {
+        // ensure the connection is closed
```
rnewman: No need for this comment.

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,227 @@
...
+    return this._doTransaction(statements.clearData, params);
+  }
+
+  create(record) {
+    const params = {
+      collection_name: this.dbname,
+      record_id: record.id,
+      record: JSON.stringify(record)
+    };
+    return this._doTransaction(statements.createData, params).then(() => {
```
rnewman: IIRC you can just do `() => record`, no?

rnewman: Question to think about: do you really want to force every individual record change to occur within an opened and closed connection?

There might be a big mismatch here between the API Kinto exposes and efficient sqlite programming. You ideally want to open the connection *once*, execute changes in *one* transaction, then close the connection.

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,227 @@
...
+      }
+    });
+  }
+
+  _doOpenConnection() {
+    const opts = { path: "kinto.sqlite", sharedMemoryCache: false }
+    return Sqlite.openConnection(opts);
+  }
+
+  _doTransaction(statement, params){
```
rnewman: This has nothing to do with transactions, no, except that sqlite implicitly creates one for the operation you run?

This should probably be called `withOpenConnection`.

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,227 @@
...
+    const self = this;
+    return Task.spawn(function* (){
+      // get a connection
+      let connection = yield self._doOpenConnection();
+
+      // ensure we're initialized
+      yield self.init;
```
rnewman: This can't be right. `self.init` itself opens a connection! I think the connection should be an argument to `self.init`.

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,227 @@
...
+            for (let statementName of createStatements) {
+              yield (connection.execute(statements[statementName]));
+            }
+
+            yield (connection.setSchemaVersion(1));
+          } else if (schema != 1) {
+            throw new Error("Unknown database schema: " + schema);
+          }
+        });
+        self.initialized = true;
```
rnewman: This is only accurate if you open the connection with the same parameters, and the same file is opened!

If you call `self.init`, then drop tables, we'll still show as initialized…

rnewman: You also never used `self.initialized` in this file. That's probably a bad sign.

fx-src/FirefoxStorage.js
rnewman: Paging @n1k0 for that. I think the adapter API needs explicit `open` and `close` that are called at the start and end of the expected lifecycle of this object.

The indexeddb version of this code just opens once in the constructor and leaves the connection open until the browser kills it.

fx-src/FirefoxStorage.js
mozmark: Actually, a remnant from old code. Will remove - but leaving for the line comments.

mozmark:
Gah; push killed some important line comments: From @rnewman 

Question to think about: do you really want to force every individual record change to occur within an opened and closed connection?

There might be a big mismatch here between the API Kinto exposes and efficient sqlite programming. You ideally want to open the connection once, execute changes in one transaction, then close the connection.

Paging @n1k0 for that. I think the adapter API needs explicit open and close that are called at the start and end of the expected lifecycle of this object.

The indexeddb version of this code just opens once in the constructor and leaves the connection open until the browser kills it.

fx-src/FirefoxStorage.js
n1k0: We're pretty much aware of that; the limitation is [documented](http://kintojs.readthedocs.org/en/latest/limitations/#transactions), we have an [active issue](https://github.com/Kinto/kinto.js/issues/177) tagged for the very next release, we've discussed the implementation details and strategy on a [lengthy thread](https://mail.mozilla.org/pipermail/kinto/2015-October/000019.html) on the Kinto development mailing-list… so I think we're covered here; just, it's a super tricky topic when it comes to IndexedDB, proper implementation should take some time to be done *well* and we badly needed to tag a first stable working version :)

mozmark:
In response; yes, it's undoubtedly the case that there's currently a mismatch between Kinto.js storage interface and efficient sqlite programming. Niko is aware of this and some work is being done to address this (see branch 172-bulk-operations - though there has been more recent thinking).

As a stop-gap - I could reintroduce manual serialization via an operation queue so we can re-use connections from any current operations - this would fix the perf. overhead of flapping the db for each operation. thoughts on this @rnewman?

fx-src/FirefoxStorage.js
mozmark: question on your first line comment here, @rnewman; I'm not sure what you mean? Should I also be manually checking the schema each time I initialize? Surely, since for a given adapter instance it will only open a connection for a single file with a similar set of params, we should be OK here?

rnewman: My point is that `self.initialized` is scoped to this object, but the pattern you use is to open the connection again each time in the method formerly known as `doTransaction`. So you init on construction, close that connection, then you might get a different DB state later!

In particular, if `dbname` here isn't `kinto.sqlite`, you *definitely* will have problems — see line 116!

rnewman: You can boil down my concern here to: you should check the schema version every time you open the connection. The real issue is that we open the connection more than once.

fx-src/FirefoxStorage.js
rnewman: > We're pretty much aware of that; the limitation is documented <http://kintojs.readthedocs.org/en/latest/limitations/#transactions>, we have an active issue <https://github.com/Kinto/kinto.js/issues/177> tagged for the very next release, we've discussed the implementation details and strategy on a lengthy thread <https://mail.mozilla.org/pipermail/kinto/2015-October/000019.html> on the Kinto development mailing-list… so I think we're covered here; just, it's a super tricky topic when it comes to IndexedDB, proper implementation should take some time to be done well and we badly needed to tag a first stable working version :)
> 

I’m actually not super concerned about transactions — we definitely want them, but not yet. I’m much more concerned about open/close semantics.

rnewman:
Serialization ~= opening one connection works as a stopgap… so long as we have a way of closing the database connection when we're done with it. That's the critical issue for me: we can't turn this on in Firefox if it flaps the DB connection, but we can't use a single long-lived connection without having a place to close it.

To reiterate: transactions are a nice-to-have, but connection management is a must-have.

n1k0:
> we can't use a single long-lived connection without having a place to close it

I can imagine how a performance killer opening and closing the sqlite db is if we do that for each atomic operation… It makes sense to expose a `close` method to the `BaseAdapter` interface for drivers needing it, I agree. And for IDB, it would do basically nothing I guess. Too bad we've just tagged 1.0 :/

rnewman:
Sounds like a 1.1!

n1k0:
> Sounds like a 1.1!

Might come sooner than expected ;) #227

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,222 @@
...
+export default class FirefoxAdapter extends BaseAdapter {
+  constructor(dbname) {
+    super();
+    this.dbname = dbname;
+  }
+
+  _init(connection) {
+    return Task.spawn(function* () {
+      yield connection.executeTransaction(function* doSetup() {
+        let schema = yield (connection.getSchemaVersion());
```
rnewman: Nit: surplus parens.

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,222 @@
...
+
+        if (schema == 0) {
+
+          for (let statementName of createStatements) {
+            yield (connection.execute(statements[statementName]));
```
rnewman: … throughout!

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,222 @@
...
+    return this._connection.executeCached(statement, params);
+  }
+
+
+  open() {
+    const self = this;
+    return Task.spawn(function* (){
+      const opts = { path: "kinto.sqlite", sharedMemoryCache: false }
+      if (!self._connection) {
+        self._connection = yield Sqlite.openConnection(opts).then(connection => self._init(connection));
```
rnewman: `.then(self._init)` is equivalent.

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,222 @@
...
+};
+
+const createStatements = ["createCollectionData",
+                          "createCollectionMetadata",
+                          "createCollectionDataRecordIdIndex"];
+
+const currentSchemaVersion = 1;
+
+export default class FirefoxAdapter extends BaseAdapter {
+  constructor(dbname) {
```
rnewman: Can we call this `collection` instead of `dbname`, given that `dbname` is exactly what it isn't? :D

fx-src/FirefoxStorage.js
```diff
@@ -0,0 +1,222 @@
...
+      });
+    }
+
+  delete(id) {
+    const params = {
+      collection_name: this.dbname,
+      record_id: id,
+    };
+    return this._executeStatement(statements.deleteData, params)
+           .then(() => id);
```
rnewman: Nit: I prefer stacked dots (throughout):

```
return this._execute…
           .then(() => id);
```

rnewman:
LGTM, @mozmark. Just nits.

mozmark:
@n1k0 tests are green. We're good to go.

PR #220, opened 2015-10-22
Fixes #176: Consistent sync result error reporting.

### LINKED ISSUE #176: Consistent reported errors formatting in sync result object
Quoting @michielbdejong in #174:

> One a side node (maybe for a follow-up): it's a pity that the error format [for outgoing](https://github.com/Kinto/kinto.js/blob/165-expose-per-record-import-errors/src/api.js#L207-L211) is so different from the errors format [for incoming](https://github.com/Kinto/kinto.js/blob/165-expose-per-record-import-errors/src/collection.js#L440), but with this patch  at least all incoming errors have the same format, regardless of whether they come from e.g. `RemoteTransformer#decode` or from e.g. `IdSchema#verify`.

n1k0:
refs #176; r=? @michielbdejong

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -330,7 +330,7 @@ The synchronization result object exposes the following properties:
 - `ok`: The boolean status of the synchronization operation; `true` if no unresolved conflicts and no errors were encountered.
 - `lastModified`: The timestamp of the latest known successful synchronization operation (no error and no conflict encountered).
 - `conflicts`: The list of unresolved conflicts encountered during both import and export operations (see *[Resolving conflicts manually](#resolving-conflicts-manually)*);
-- `errors`:    The list of encountered errors, if any.
+- `errors`:    The list of encountered errors, if any. Each error has a `type` property, which value can either be `incoming` or `outgoing` depending on which part of the sync flow it's been caught from;
 - `created`:   The list of remote records which have been successfully imported into the local database.
 - `updated`:   The list of remote record updates which have been successfully reflected into the local database.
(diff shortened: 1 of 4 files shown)


docs/api.md
```diff
@@ -330,7 +330,7 @@ The synchronization result object exposes the following properties:
 - `ok`: The boolean status of the synchronization operation; `true` if no unresolved conflicts and no errors were encountered.
 - `lastModified`: The timestamp of the latest known successful synchronization operation (no error and no conflict encountered).
 - `conflicts`: The list of unresolved conflicts encountered during both import and export operations (see *[Resolving conflicts manually](#resolving-conflicts-manually)*);
-- `errors`:    The list of encountered errors, if any.
+- `errors`:    The list of encountered errors, if any. Each error has a `type` property, which can be set to `incoming` or `outgoing` depending on which part of the sync flow it's been caught from;
```
michielbdejong: Nit: Remove 'set to' from the phrase, because this property is something that as a developer you read, not something you write to.

test/api_test.js
```diff
@@ -511,7 +511,7 @@ describe("Api", () => {
                 {
                   path: `/${SPV}/buckets/blog/collections/articles/records/1`,
                   sent: published[0],
-                  error: { 500: true }
+                  error: { 500: true },
```
michielbdejong: Nit

michielbdejong: Oh right, I guess this is on purpose, sorry.

michielbdejong:
r+

PR #221, opened 2015-10-22
Fixes #114: Drop collection metas on #clear().

### LINKED ISSUE #114: Clarify if 'clearing the collection' affects only client, or client+server
Sometimes I will want to remove all the items from the collection on the server.
Other times (actually, probably only when debugging, or maybe when offline and offering the user something like `git checkout -- .` to discard unpushed changes), I only want to flush the cache, to make sync start from zero.
It is not clear to me which one of these two is done by http://kintojs.readthedocs.org/en/latest/api/#clearing-the-collection

n1k0:
refs #114; r=? @michielbdejong

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -230,16 +230,15 @@ export default class Collection {
   }
 
   /**
-   * Deletes every records in the current collection.
-   *
-   * XXX: refs #114, collection metas should be cleared.
+   * Deletes every records in the current collection and marks the collection as
+   * never synced.
    *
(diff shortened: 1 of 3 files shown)


n1k0:
> The command "git checkout -qf f45fbe28e6ca0472e7ca10e61194a3b59320f44a" failed and exited with 128 during .

Github seems to be [under DDos](https://status.github.com/).

michielbdejong:
r+

PR #223, opened 2015-10-22
Fixes #99: Added docs for authorization.

### LINKED ISSUE #99: Add an Authentication section to the docs
Even if really simple, it's especially important to document the FxA authentication process.

n1k0:
Refs #99. r=? @leplatrem @Natim

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -240,10 +240,11 @@ Result:
 
 ## Authentication
 
-Authenticating against a Kinto server can be achieved in [two distinct ways](http://kinto.readthedocs.org/en/latest/api/cliquet/authentication.html):
+Authenticating against a Kinto server can be achieved in [three distinct ways](http://kinto.readthedocs.org/en/latest/api/cliquet/authentication.html):
 
 - Using Basic Auth
 - Using an OAuth Bearer Token
+- Using Firefox Account
 
 ### Using Basic Auth
 
@@ -283,6 +284,10 @@ const kinto = new Kinto({
   }
 });
 ```
+
+### Using Firefox Account
+
+If your Kinto instance supports it, you can also use a [Firefox Account](https://accounts.firefox.com/) Bearer Token. Once obtained, the procedure is the same as with a standard Bearer Token, as documented above.
 
 ## Fetching and publishing changes
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -240,10 +240,9 @@ Result:
 
 ## Authentication
 
-Authenticating against a Kinto server can be achieved in [two main ways](http://kinto.readthedocs.org/en/latest/api/cliquet/authentication.html):
+Authenticating against a Kinto server can be achieved by adding an `Authorization` header to the request.
 
-- Using Basic Auth
-- Using an OAuth Bearer Token (eg. Firefox Account)
+By default Kinto server supports Basic Auth authentication, but others mechanisms can be activated such as OAuth (eg. [Firefox Account](https://accounts.firefox.com/))
 
 ### Using Basic Auth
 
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -234,8 +234,70 @@ Result:
...
+> - Clearing the local collection will mark the collection as never synchronized;
 > - Detailed API documentation for `Collection#clear()` is available [here](https://doc.esdoc.org/github.com/Kinto/kinto.js/class/src/collection.js~Collection.html#instance-method-clear).
 
+## Authentication
+
+Authenticating against a Kinto server can be achieved in [three distinct ways](http://kinto.readthedocs.org/en/latest/api/cliquet/authentication.html):
+
+- Using Basic Auth
+- Using an OAuth Bearer Token
+- Using Firefox Account
```
Natim: Firefox Account and OAuth Bearer Token is basically the same thing.
Also note that it is possible to plug any pyramid auth backend with kinto

docs/api.md
```diff
@@ -234,8 +234,70 @@ Result:
...
+
+Simply provide an `Authorization` header option to the `Kinto` constructor:
+
+```js
+const username = "my_username";
+const password = "my_password";
+const kinto = new Kinto({
+  remote: "https://my.server.tld/v1",
+  headers: {
+    Authorization: "Basic " + btoa(`${username}:${password}`)
```
Natim: You can also use anything you want here.

```Authorization: "Basic " + btoa('Some random Token')```

**Comment by Natim** [comment]
It looks good to me, I wouldn't go this specific because telling people how to provide specific Authorization headers should be enough. Also it is good to have working example in the documentation

File: docs/api.md (hunk 1 above)
Comment (n1k0) [inline docs/api.md]: > Firefox Account and OAuth Bearer Token is basically the same thing.

Yeah I know, but I know FA doesn't strictly implement the spec, so… Your call.

>  Also note that it is possible to plug any pyramid auth backend with kinto

Maybe that's not really something the JavaScript client shouldn't be too much concerned about?

File: docs/api.md (hunk 2 above)
Comment (n1k0) [inline docs/api.md]: I used the username/password couple because it's a classic, but I'll add a note about that.

File: docs/api.md [hunk 3]
```diff
@@ -234,8 +234,74 @@ Result:
 ```
 
 > #### Notes
+>
+> - Clearing the local collection will mark the collection as never synchronized;
 > - Detailed API documentation for `Collection#clear()` is available [here](https://doc.esdoc.org/github.com/Kinto/kinto.js/class/src/collection.js~Collection.html#instance-method-clear).
 
+## Authentication
+
+Authenticating against a Kinto server can be achieved in [two main ways](http://kinto.readthedocs.org/en/latest/api/cliquet/authentication.html):
```
Natim: Authenticating against a Kinto server can be achieved by adding an Authorization header to the request.

By default Kinto server supports Basic Auth authentication, but other mecanism can be activated such as OAuth (eg. Firefox Account)

docs/api.md
```diff
@@ -234,8 +234,74 @@ Result:
...
+```js
+kinto.collection("articles")
+  .sync({
+    headers: {Authorization: "Basic " + btoa(`${username}:${password}`)}
+  }).then(…);
+```
+
+> #### Notes
+>
+> - This authentication method should never been used in production, and passwords in clear should never be part of the source code;
```
Natim: I don't really agree, requests are send through HTTPS so using Basic Auth is safe here. The only limitation is that it is impossible to change the user password because each couple of (username:password) for a unique identifier for the user.

Natim: Also within the scope of Kinto.js authorization credentials should be read from either the user input or the URL hash but not in the source code.

Natim:
r+

PR #226, opened 2015-10-27
Fixed links to public assets in the docs.

n1k0:
r=? @michielbdejong

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -14,14 +14,8 @@ $ npm install kinto --save
 
 ## Static assets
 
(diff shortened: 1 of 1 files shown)


michielbdejong:
I would keep the separation between 'stable' and 'latest master', e.g. linking to https://github.com/Kinto/kinto.js/tree/3813157 (because the version tags are in the master branch, you have to use the commit id to link to v1.0.0 in the gh-pages branch). Otherwise, r+!

michielbdejong:
Another option is maybe to create a `/release` folder in the gh-pages branch, and keep the built versions of all tagged releases there.

PR #227, opened 2015-10-27
Exposed open() and close() methods to the BaseAdapter interface.

n1k0:
Refs https://github.com/Kinto/kinto.js/pull/219#issuecomment-151587615

Basically we now add `open` and `close` methods to the `BaseAdapter` base class, so implementers must implement them.

Also is now exposed a `Collection#close` method so apps using adapters needing explicit closing can call that (opening is always done implicitly when a first operation is scheduled for execution).

cc @rnewman 

@mozmark: would this work for you?

### DIFF
--- a/src/adapters/IDB.js
+++ b/src/adapters/IDB.js
@@ -33,6 +33,7 @@ export default class IDB extends BaseAdapter {
   /**
    * Ensures a connection to the IndexedDB database has been opened.
    *
+   * @override
    * @return {Promise}
    */
   open() {
@@ -70,6 +71,20 @@ export default class IDB extends BaseAdapter {
   }
 
   /**
+   * Closes current connection to the database.
+   *
+   * @override
+   * @return {Promise}
+   */
+  close() {
+    if (this._db) {
+      this._db.close(); // indexedDB.close is synchronous
+      this._db = null;
+    }
+    return super.close();
+  }
+
+  /**
    * Returns a transaction and a store objects for this collection.
    *
    * To determine if a transaction has completed successfully, we should rather
@@ -93,6 +108,7 @@ export default class IDB extends BaseAdapter {
   /**
    * Deletes every records in the current collection.
    *
+   * @override
    * @return {Promise}
    */
   clear() {
@@ -111,6 +127,7 @@ export default class IDB extends BaseAdapter {
    *
    * Note: An id value is required.
    *
+   * @override
    * @param  {Object} record The record object, including an id.
    * @return {Promise}
    */
@@ -128,6 +145,7 @@ export default class IDB extends BaseAdapter {
   /**
    * Updates a record from the IndexedDB database.
    *
+   * @override
    * @param  {Object} record
    * @return {Promise}
    */
@@ -145,6 +163,7 @@ export default class IDB extends BaseAdapter {
   /**
    * Retrieve a record by its primary key from the IndexedDB database.
    *
+   * @override
    * @param  {String} id The record id.
    * @return {Promise}
    */
@@ -162,6 +181,7 @@ export default class IDB extends BaseAdapter {
   /**
    * Deletes a record from the IndexedDB database.
    *
+   * @override
    * @param  {String} id The record id.
    * @return {Promise}
    */
@@ -179,6 +199,7 @@ export default class IDB extends BaseAdapter {
   /**
    * Lists all records from the IndexedDB database.
    *
+   * @override
    * @return {Promise}
    */
   list() {
@@ -203,6 +224,7 @@ export default class IDB extends BaseAdapter {
   /**
    * Store the lastModified value into metadata store.
    *
+   * @override
    * @param  {Number}  lastModified
    * @return {Promise}
    */
@@ -221,6 +243,7 @@ export default class IDB extends BaseAdapter {
   /**
    * Retrieve saved lastModified value.
    *
+   * @override
    * @return {Promise}
    */
   getLastModified() {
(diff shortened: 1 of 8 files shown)


src/adapters/IDB.js
```diff
@@ -70,6 +70,16 @@ export default class IDB extends BaseAdapter {
   }
 
   /**
+   * Closes current connection to the database. Doesn't do anything as IndexedDB
+   * database can't be closed.
+   *
+   * @return {Promise}
+   */
+  close() {
+    return Promise.resolve();
```
mozmark: https://developer.mozilla.org/en-US/docs/Web/API/IDBDatabase/close ?

n1k0: **TIL**

mozmark:
So the key thing for me is when these get called. Is this a case of 'open' and 'close' around operations in the API?  For my use-case, it only really matters that we share a connection for the duration of a sync (since I only really need sync and list) but I can see for others that will differ - if we're not sharing connections between a bunch of create, update or deletions then we'll still have a problem.

n1k0:
There's `Collection#close` you can call when you need it, eg. after a sync():

```js
coll.sync().then(_ => coll.close());
```

We could also do this automatically after the `sync()` flow is completed, in every case. Thoughts?

rnewman:
I'm generally in favor of making callers think about what they're doing and notice errors, so my opinion is:

* Don't implicitly open. You can hide or cause bugs this way — you might reopen a connection that you just closed, causing it to leak, or you might cause flapping in the presence of a mis-paired set of open/close calls.
* Don't implicitly close. You can cause bugs or performance problems this way.
* Return a rejected promise, or throw, if the connection isn't open when you try to perform an operation that needs it. You'll find bugs this way.

These should be the documented behaviors of the API, so that *all* consumers of *any* adapter look like this:

```
let coll = SomeAdapter(...);
yield coll.open();
coll.isOpen;       // true
try {
    yield coll.insert(...);
    yield coll.delete(...);
} finally {
    coll.close();
}
coll.isOpen;      // false
yield coll.query(...);      // Throws or rejects: "collection not open"
```

n1k0:
In this case as opening a connection is an async operation, that means `KintoBase#collection()` should now return a promise, so the collection instance has the necessary time to ask the adapter to provide that opened connection. This breaks the API, so that means v2.

mozmark:
No, KintoBase#collection can stay async - it's the open that needs to return a promise - which it does. If we're explicit in the docs that, for implementations where we need the connection management, we throw if open hasn't been called then this should be fine.

mozmark:
I think this works for now.

src/collection.js
```diff
@@ -242,6 +242,15 @@ export default class Collection {
...
   /**
+   * Closes any current connection to the database for this collection.
+   *
+   * @return {Promise}
+   */
+  close() {
+    return this.db.close();
+  }
+
+  /**
```
mozmark: We need an 'open' and an 'isOpen' too.

open - returns a promise that resolves when adapter.open does
isOpen - that returns true once the collection is open

src/adapters/base.js
```diff
@@ -16,7 +16,25 @@ export default class BaseAdapter {
   }
 
   /**
-   * Adds a record to the IndexedDB database.
+   * Opens a connection o the database.
+   *
+   * @return {Promise}
+   */
+  open() {
+    throw new Error("Not Implemented.");
```
mozmark: I would argue that base could do what you've currently done in LocalStorage - that way, other adapters that don't need to explicitly manage connections don't have to worry about implementing them. If we throw when these are unimplemented, this is a breaking change.

src/collection.js
n1k0: Is it really necessary for now? I can't see much issue with implicit opening, you already get a rejection from the adapter when trying to execute a query with no opened db…

src/adapters/base.js
n1k0: That's a good point.

n1k0: Hmm actually, scratch my latest comment; nobody is actually calling `#close()` as it's a new method introduced by this patch :)

I think we can live with forcing implementers to explicitly provide their `#close()` implementation.

src/collection.js
n1k0: Also, forcing "opening" a collection would break existing apps using Kinto.js 1.0, because you'd then need to do:

```js
const coll = kinto.collection("foo"); 
coll.open().then(_ => coll.create());
```

n1k0:
So after discussing it a little over IRC, here's the plan:

```
[11:28:11] mgoodwin 1) add open and close to collection
[11:28:23] mgoodwin 2) Add open and close to all adapters - no-op when it's not really needed
[11:29:15] mgoodwin 3) have base implement the no-op open and close so that other potential implementations inherit that
[11:29:30] mgoodwin 4) have old adapters not throw if not open - since they don't care anyway
[11:29:16] NiKo` yeah so old adapters would keep using the API the same way
```

mozmark:
OK. R+ from me; this will allow sensible explicit connection management.

n1k0:
I'm now thinking removing `Collection#open` and `#close` are they're just proxy to `collection.db.open()` and `collecttion.db.close()`. That allows managing specific connection opening/closing needs specifically in the target project, not in Kinto.js itself — which still allows it as soon as the BaseAdapter interface require these methods to be always available anyway. Thoughts?

n1k0:
After IRC discussion, Collection#open and #close are now gone. Explicit db opening/closing should be achieved leveraging the adapetr api exposed through the `db` property of the `Collection` instance.

PR #230, opened 2015-10-29
Updated dist command to add current version to asset names.

n1k0:
Refs https://github.com/Kinto/kinto.js/pull/226#issuecomment-151615723

This will basically allow to generate dist assets with the current kinto.js version in their filename, eg. `kinto-1.0.0.dev.js` before pushing them to the `gh-pages` branch.

Notes: 

- We'll need to update the `gh-pages` branch by hand to put the 1.0.0 version, as code have been updated on master since latest release (**edit: done**);
- I've also removed the remaining versioned dist assets from master, as we really want to host them and make them available through gh-pages only.

cc @michielbdejong


docs/hacking.md
```diff
@@ -66,15 +66,15 @@ Of course, as for `npm test`, you can explictely pass the pass to the Kinto serv
 
     $ npm run dist
 
-This should have created the following assets:
+This should have created the following assets, with the current version in the filename (here, `1.0.0`):
```
michielbdejong: Maybe rephrase 'here' to 'for instance', so that we don't have to update this file with each release.

docs/hacking.md
```diff
@@ -66,15 +66,15 @@ Of course, as for `npm test`, you can explictely pass the pass to the Kinto serv
...
 
--  `demo/kinto.dev.js`: Developement version, unminified, embedding source maps;
--  `demo/kinto.min.js`: Production version, minified, no source maps.
+-  `demo/kinto-1.0.0.dev.js`: Developement version, unminified, embedding source maps;
```
michielbdejong: Nit: Development

michielbdejong:
r+

PR #234, opened 2015-10-29
[WiP] Use bucket in demo

michielbdejong:
Haven't tested this yet.

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -1,7 +1,8 @@
 function main() {
   var db = new Kinto({
     remote: "https://kinto.dev.mozaws.net/v1/",
-    headers: {Authorization: "Basic " + btoa("user:pass")}
+    headers: {Authorization: "Basic " + btoa("user:pass")},
+    bucket: 'user'
   });
   var tasks = db.collection("tasks");
 


n1k0:
Could you explain the rationale behind this change? Demo users are all sharing the same auth credentials on purpose...

michielbdejong:
It's just about best practices. We're pretty sure now that all known problems with the default bucket have been fixed, and so both systems (default bucket or one bucket per user) should work correctly now. But in case there are unknown unknowns there, keeping buckets separate feels like a nice extra safety measure that doesn't cost the developer anything, right?

Natim:
> keeping buckets separate feels like a nice extra safety measure that doesn't cost the developer anything

Not exactly because the default bucket implicitely creates collections when first use which is not the case with other buckets.

n1k0:
Because of implicit default bucket creation, we cannot simply specify a custom bucket that way; a better approach would probably be to properly document the procedure when one's willing to use custom buckets (with explicit creation operations).

Natim:
The steps are:

```
    const createBucket = (server, bucketId, collectionId, authorization) => {
       return fetch(`${server}buckets/${bucketId}`, {
         method: 'PUT',
         headers: {
           Authorization: authorization
         }
       }).then(() => fetch(`${server}buckets/${bucketId}/collections/${collectionId}`, {
         method: 'PUT',
         headers: {
           Authorization: authorization
         }
       }));
    };
```

michielbdejong:
ok!

PR #243, opened 2015-10-30
Fixes #241: Preserving old files in gh-pages branch.

### LINKED ISSUE #241: Preserve old files in gh-pages branch, so older assets versions aren't overriden
That's easy now https://github.com/tschaub/gh-pages/pull/26 has landed, and #237 is merged.

n1k0:
Refs #241.

We also need to ensure we're not publishing previously versioned assets, when reusing the current `package.json` version tag if it's not been updated on master while code has changed.


n1k0:
This is obsolete, already fixed by the new version of the publish command.

PR #247, opened 2015-11-03
Upgrade to babel 6 (fixes #242)

### LINKED ISSUE #242: Upgrade to Babel 6
Notes: 

- #240 shows upgrading to 6.0.12 currently breaks;
- An upgrading guide is available [there](https://medium.com/@malyw/how-to-update-babel-5-x-6-x-d828c230ec53#.wa4cvxvhg).

leplatrem:
* [x] Upgrade dependencies
* [x] Upgrade command line and config
* [ ] ~~Get rid of workaround for `export default class`~~
* [ ] ~~Fix behaviour with default parameters~~
* [x] Fix firefox dist


src/adapters/IDB.js
```diff
@@ -259,3 +259,5 @@ export default class IDB extends BaseAdapter {
     });
   }
 }
+
+export default IDB;
```
n1k0: Any specific rationale behind these changes? :)

n1k0:
Hmm, it seems default argument value are suddenly not supported anymore. Maybe we need to set a flag for it?

n1k0:
After messing around the branch, investigating the loss of default arg support, I couldn't find a fix. I suspect this is related to our test environment, as the feature works just fine [on the babel website](http://babeljs.io/repl/#?experimental=false&evaluate=true&loose=false&spec=false&code=class%20Api%20{%0A%20%20constructor%28remote%2C%20events%2C%20options%3D{}%29%20{%0A%20%20%20%20console.log%28%22options%22%2C%20options%29%0A%20%20}%0A}%0A%0Anew%20Api%28%22a%22%2C%20%22b%22%29%3B).

I suggest we park this for now and just wait for the ecosystem to catch up with this new version of babel.

n1k0:
Note: [this](http://www.jisaacks.com/writing-mocha-tests-in-es6-with-babel-6/) guy seems to have successfully integrated mocha and babel 6 sharing a very similar configuration, might be worth creating a minimal test case to investigate the loss of default arguments support.

leplatrem:
@n1k0 I did some progress on this.

The last missing part is the Firefox build. 

Before the transpilation of arrow and generator functions were disabled using the `--blacklist` parameter.

But now that Babel uses presets, this CLI option has been disabled (error `Unknown option: base.blacklist while parsing file`).

And I believe we should define our own preset which would be something like `firefox = es2015 - generator - arrow`.
@mozmark what do you think ?

mozmark:
I may be that support for these features are working well enough that we can do without this part of transpilation altogether. We were actually close to being able to do this when the Firefox storage support initially landed. My suggestion would be to see whether this works first (a try-push on a few platforms should illuminate any issues) and deal with what doesn't afterwards.

That said, this seems like a sensible approach should the support for these features in fx be lacking still.

leplatrem:
A quick status update on this.

leplatrem:
A quick status update on this:

Using es2015 preset (current branch status), `moz-kinto-client.js` is not loaded correctly : 

```
$ cd mozilla-central/
$ hg head
changeset:   280132:02398f2be72b
date:        Fri Jan 15 13:52:19 2016 +0100

$ cp ../kinto.js/dist/moz-kinto-client.js services/common
$ ./mach xpcshell-test services/common/tests/unit/test_kinto.js
...
...
LOG: Thread-1 ERROR TypeError: loadKinto is not a function at /home/mathieu/Code/Mozilla/mozilla-central/obj-x86_64-unknown-linux-gnu/_tests/xpcshell/services/common/tests/unit/test_kinto.js:13
```

I first thought it could be a sort of interpreter error that could lead Fx to give up on loading the file.
But I couldn't see any related log/output. Also while I was playing with plugins, I generated another version and could get an explicit syntax error:

```
 0:00.38 LOG: Thread-1 ERROR SyntaxError: invalid for/in left-hand side at resource://services-common/moz-kinto-client.js:2762
```

So this has to be something else.

### Minimal set of transform

The minimal set of transformations that does not bring up any syntax error is this one (transform-object-rest-spread + transform-es2015-modules-commonjs + transform-es2015-for-of) :

```
browserify -s loadKinto -e fx-src/index.js -o temp.jsm -t [ babelify --sourceMapRelative . --plugins [ transform-object-rest-spread transform-es2015-modules-commonjs transform-es2015-for-of ] ] && mkdir -p dist && cp fx-src/jsm_prefix.js dist/moz-kinto-client.js && cat temp.jsm >> dist/moz-kinto-client.js && rm temp.jsm
```

It ends to the same `TypeError: loadKinto` problem ([full log](https://pastebin.mozilla.org/8856918)). I don't know where to look so far. 

### Notes

* The babel 5 (kinto.js master) works with the same changeset in Mozilla Central
* I cannot see much difference in the exported symbols and function definition between 5 and 6 (I know there is to be one)

Stay tuned!

leplatrem:
Ok we could fix the problem.

It was due to the way `export default function` is transpiled for commonjs (required by browserify).

Now, for Firefox we transpile:
* `for of` loops
* commonjs for browserify

And for the rest, we use the `es2015` explicitly everywhere. 

> I could not find a way to enable `es2015` everywhere and disable it (like blacklist) just for Firefox

n1k0:
There's an issue with the `npm run demo` command:

```
ReferenceError: Unknown plugin "transform-es2015-modules-commonjs" specified in "base" at 0, attempted to resolve relative to "fx-src" while parsing file: /home/niko/work/kinto.js/fx-src/index.js
    at /home/niko/work/kinto.js/node_modules/babel-core/lib/transformation/file/options/option-manager.js:193:17
    at Array.map (native)
    at Function.normalisePlugins (/home/niko/work/kinto.js/node_modules/babel-core/lib/transformation/file/options/option-manager.js:173:20)
    at OptionManager.mergeOptions (/home/niko/work/kinto.js/node_modules/babel-core/lib/transformation/file/options/option-manager.js:271:36)
    at OptionManager.init (/home/niko/work/kinto.js/node_modules/babel-core/lib/transformation/file/options/option-manager.js:416:10)
    at File.initOptions (/home/niko/work/kinto.js/node_modules/babel-core/lib/transformation/file/index.js:191:75)
    at new File (/home/niko/work/kinto.js/node_modules/babel-core/lib/transformation/file/index.js:122:22)
    at Pipeline.transform (/home/niko/work/kinto.js/node_modules/babel-core/lib/transformation/pipeline.js:42:16)
    at Babelify._flush (/home/niko/work/kinto.js/node_modules/babelify/index.js:28:24)
    at Babelify.<anonymous> (_stream_transform.js:118:12)
```

I'm on it.

n1k0:
re previous comment, false alarm, i was missing an npm install.

On the other hand, the demo was broken because we needed to apply the same fix as for the FFx exports. Could you please have a look at my [latest commit](https://github.com/Kinto/kinto.js/commit/ae64e20dab1f24ce1dad0796ced4dd86eb063274) @leplatrem? I was wondering if we should properly test for the presence of `module`, or if we should create an intermediary file instead... Thoughts?

demo/index.html
```diff
@@ -37,7 +37,7 @@
       </li>
     </template>
   </div>
-  <script src="//npmcdn.com/kinto@1.1.2/dist/kinto-1.1.2.js"></script>
+  <script src="kinto.js"></script>
```
leplatrem: in publish demo we have `cp dist/kinto-$npm_package_version.js demo/`

package.json
```diff
@@ -5,19 +5,20 @@
...
-    "publish-demo": "gh-pages -d demo",
+    "dist-dev": "browserify -s Kinto -d -e src/index.js -o dist/kinto-$npm_package_version.js -t [ babelify --sourceMapRelative . --presets [ es2015 ] ]",
+    "dist-noshim": "browserify -s Kinto -g uglifyify --ignore isomorphic-fetch --ignore babel-core/polyfill -e src/index.js -o dist/kinto-$npm_package_version.noshim.js -t [ babelify --sourceMapRelative . --presets [ es2015 ] ]",
+    "dist-prod": "browserify -s Kinto -g uglifyify -e src/index.js -o dist/kinto-$npm_package_version.min.js -t [ babelify --sourceMapRelative . --presets [ es2015 ] ]",
+    "dist-fx": "browserify -s loadKinto -e fx-src/index.js -o temp.jsm -t [ babelify --sourceMapRelative . --plugins [ transform-es2015-modules-commonjs transform-es2015-for-of ] ] && mkdir -p dist && cp fx-src/jsm_prefix.js dist/moz-kinto-client.js && cat temp.jsm >> dist/moz-kinto-client.js && rm temp.jsm",
+    "publish-demo": "npm run dist-prod && cp dist/kinto-$npm_package_version.js demo/ && gh-pages -d demo",
```
leplatrem: see https://github.com/Kinto/kinto.js/commit/572511a0e640ef27a9e76d53ef8c6b7990fb5ffb

demo/index.html
n1k0: ah right for publish, but for run demo it's another issue :/

package.json
leplatrem: this fixes #298 by the way

n1k0: Nice. Found a fix, waiting for a green badge.

n1k0:
I'll have to update the releasing docs as well.

n1k0:
Tested latest moz-kinto-client.js generated using latest commits, works fine.

@leplatrem final r=?

leplatrem:
\o/ r+ with stash :)

PR #264, opened 2015-11-16
[Parked] Fixes #263: Handle network request timeouts.

### LINKED ISSUE #263: Handle network request timeouts.
Right now when a server is down/unreachable, a first `OPTIONS` request sent will hang forever. 

We need to support request timeouts with a configurable value. Possible pattern to use:

```js
var timeoutPromise = new Promise(function(resolve) {
  setTimeout(resolve, 10000);
})

var load = fetch(‘/some-resource’);

Promise.race([timeoutPromise, load]).then(function(res) {
  if (res instanceof Response) {
    // Network replied in time.
  } else {
    // Timeout
  }
})
```

n1k0:
refs #263. r=? @leplatrem

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -19,15 +19,25 @@ export default class HTTP {
   }
 
   /**
+   * Default options.
+   *
+   * @type {Object}
+   */
+  static get defaultOptions() {
+    return {timeout: 5000, requestMode: "cors"};
+  }
+
+  /**
    * Constructor.
    *
    * Options:
+   * - {Number} timeout      The request timeout in ms (default: `5000`).
    * - {String} requestMode  The HTTP request mode (default: `"cors"`).
    *
    * @param {EventEmitter} events  The event handler.
    * @param {Object}       options The options object.
    */
-  constructor(events, options={requestMode: "cors"}) {
+  constructor(events, options={}) {
     // public properties
     /**
      * The event emitter instance.
@@ -38,12 +48,20 @@ export default class HTTP {
     }
     this.events = events;
 
+    options = Object.assign({}, HTTP.defaultOptions, options);
+
     /**
      * The request mode.
      * @see  https://fetch.spec.whatwg.org/#requestmode
      * @type {String}
      */
     this.requestMode = options.requestMode;
+
+    /**
+     * The request timeout.
+     * @type {Number}
+     */
+    this.timeout = options.timeout;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


n1k0:
Also cc @michielbdejong for feedback

Natim:
LGTM.

src/http.js
```diff
@@ -108,6 +131,11 @@ export default class HTTP {
         }
         return {status, json, headers};
       });
+    return Promise.race([timeoutPromise, requestPromise])
```
leplatrem: nit: add a small comment to specify that the quickest one is resolved

src/http.js
```diff
@@ -62,11 +80,16 @@ export default class HTTP {
...
    */
   request(url, options={headers:{}}) {
-    let response, status, statusText, headers;
+    let response, status, statusText, headers, _timeout;
     // Ensure default request headers are always set
     options.headers = Object.assign({}, HTTP.DEFAULT_REQUEST_HEADERS, options.headers);
     options.mode = this.requestMode;
-    return fetch(url, options)
+    const timeoutPromise = new Promise((_, reject) => {
+      _timeout = setTimeout(() => {
```
leplatrem: `_timeoutId` ? (*not sure if relevant*)

leplatrem:
r+!

n1k0:
I'm realizing there's no way to guarantee the request is aborted once the timeout value is reached, see https://github.com/whatwg/fetch/issues/27.

That's annoying, as the user may believe the request was never actually performed, while it totally may, resulting in a possibly messy situation. 

I don't have any solution in mind right now.

n1k0:
Issue is partially confirmed by changing the test like this:

```js
    describe("Request timeout", () => {
      it("should timeout the request", () => {
        sandbox.stub(root, "fetch").returns(
          new Promise(resolve => {
            console.log("executed :(");
            setTimeout(resolve, 20000);
          }));
        return http.request("/")
          .should.eventually.be.rejectedWith(Error, /timeout/);
      });
    });
``` 

Where "executed :(" is printed out to the console regardless.

n1k0:
The only option to properly support this would be to switch back to XHR, which supports timeouts natively. I'm just gonna park this for now out of depression ;)

michielbdejong:
Ah, could well be the fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1224203 :)

michielbdejong:
Can't we just set a variable 'hasTimedOut' and check for this before returning success?

n1k0:
Fixed by #273

PR #266, opened 2015-11-20
Explicit which protocol_version is supported by Kinto,js

Natim:
r? @n1k0


n1k0:
I'd really prefer all the updates related to 1.9.0 and protocol v2 (code, config and docs) in a single PR.

Natim:
Well we can update to Kinto 1.9.0 without this taking care of protocol_version yet. But for the documentation I agree with you.

leplatrem:
Dropping this, since this is being worked on in `kinto-client` directly :)

PR #265, opened 2015-11-20
Add support for db.import

Natim:
For some of the Mozilla platform use-cases, it's desirable to have the ability to import initial data. Since we only need it here, for the time being, we can prototype the idea in the Firefox storage adapter and apply what we learn to the other adapters.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -214,6 +214,29 @@ articles.list({sort: "-title"})
 > - By default, the records are sorted on `last_modified` in descending order.
 > - As mentioned in the [limitations](limitations.md) section, the sort is performed in memory.
 
+## Importing a data dump locally
+
+You may want to preload a dump of records from the server, before
+actually starting the first sync with it.
+
+The list of imported records is returned.
+
+```js
+articles.loadDump([
+  {
+    id: "2dcd0e65-468c-4655-8015-30c8b3a1c8f8",
+    title: "baz",
+    last_modified: 1432222889337
+  }
+])
+  .then(records => console.log(records));
+```
+
+> #### Notes
+>
+> - Existing records are replaced if they do not have more recent modifications.
+> - Imported records won't be synced with the server.
+> - The importation is optimized in a single database transaction.
 
 ## Clearing the collection
 
--- a/fx-src/FirefoxStorage.js
+++ b/fx-src/FirefoxStorage.js
@@ -72,7 +72,12 @@ const statements = {
   "listRecords": `
     SELECT record
       FROM collection_data
-        WHERE collection_name = :collection_name;`
+        WHERE collection_name = :collection_name;`,
+
+  "importData": `
+    REPLACE INTO collection_data (collection_name, record_id, record)
+      VALUES (:collection_name, :record_id, :record);`
+
 };
 
 const createStatements = ["createCollectionData",
@@ -193,6 +198,39 @@ export default class FirefoxAdapter extends BaseAdapter {
         let row = result[k];
         records.push(JSON.parse(row.getResultByName("record")));
       }
+      return records;
+    });
+  }
+
+  loadDump(records) {
+    let connection = this._connection;
+    let collection_name = this.collection;
+    return Task.spawn(function* () {
+      yield connection.executeTransaction(function* doImport() {
+        for (let record of records) {
+          const params = {
+            collection_name: collection_name,
+            record_id: record.id,
+            record: JSON.stringify(record)
+          };
+          yield connection.execute(statements.importData, params);
+        }
+        const lastModified = Math.max(...records.map(record => record.last_modified));
+        const params = {
+          collection_name: collection_name
+        };
+        const previousLastModified = yield connection.execute(
+          statements.getLastModified, params).then(result => {
+            return result ? result[0].getResultByName('last_modified') : -1;
+          });
+        if (lastModified > previousLastModified) {
+          const params = {
+            collection_name: collection_name,
+            last_modified: lastModified
+          };
+          yield connection.execute(statements.saveLastModified, params);
+        }
+      });
       return records;
     });
   }
--- a/src/collection.js
+++ b/src/collection.js
@@ -787,4 +787,64 @@ export default class Collection {
         return this.pullChanges(result, options);
       });
   }
+
+  /**
+   * Load a list of records already synced with the remote server.
+   *
+   * @param  {Array} records.
+   * @param  {Object} options Options.
+   * @return {Promise}
+   */
+  loadDump(records) {
+    const reject = msg => Promise.reject(new Error(msg));
+    if (!(records instanceof Array)) {
+      return reject("Records is not an array.");
+    }
+
+    for(const record of records) {
+      if (!record.id || !this.idSchema.validate(record.id)) {
+        return reject("Record has invalid ID: " +
+                      JSON.stringify(record));
+      }
+
+      if (!record.last_modified) {
+        return reject("Record has no last_modified value: " +
+                      JSON.stringify(record));
+      }
+    }
+
+    // Fetch all existing records from local database,
+    // and skip those who are newer or not marked as synced.
+    return this.list({}, {includeDeleted: true})
+      .then(res => {
+        return res.data.reduce((acc, record) => {
+          acc[record.id] = record;
+          return acc;
+        }, {});
+      })
+    .then(existingById => {
+      return records.filter(record => {
+        const localRecord = existingById[record.id];
+        const shouldKeep = (
+          // No local record with this id.
+          localRecord === undefined ||
+          // Or local record is synced
+          localRecord._status === "synced" &&
+          // And was synced from server
+          localRecord.last_modified !== undefined &&
+          // And is older than imported one.
+          record.last_modified > localRecord.last_modified
+        );
+        return shouldKeep;
+      });
+    })
+    .then(newRecords => {
+      return newRecords.map(record => {
+        return Object.assign({}, record, {
+          _status: "synced"
+        });
+      });
+    })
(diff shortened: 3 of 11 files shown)


Natim:
We've added adapter import method for LocalStorage and IndexDB with @leplatrem.

leplatrem:
Remaining tasks:

* [ ] Documentation
* [ ] Move of updated last modified to adapter tests
* [ ] Update collection last modified only if superior

Natim:
In order to update the FirefoxStorage adapter, we need to add tests in Firefox.

Here is the flow to run the FirefoxStorage tests:

 1. kinto.js> ``npm run dist-fx``
 2. ``cp ~/kinto.js/dist/moz-kinto-client.js ~/moz-central/services/common/``
 3. moz-central> ``./mach build ``
 4. Update ``services/common/tests/unit/test_storage_adapter.js`` and ``services/common/tests/unit/test_kinto.js``
 5. Run the tests with ``./mach xpcshell-test services/common/tests/unit/test_storage_adapter.js``

Natim:
Ok so we are almost there I guess.

The remaining problem to handle is the If-Match header.

Because the import method is overridding, we do not want to override already synced records so we need a way to tell, import these records only if the lastModified value is older or something like that.

src/collection.js
```diff
@@ -787,4 +787,45 @@ export default class Collection {
...
+   */
+  import(records) {
+    const reject = msg => Promise.reject(new Error(msg));
+    if (!(records instanceof Array)) {
+      return reject("Records is not an array.");
+    }
+
+    if (!records.every(record => {
+      return record.id && this.idSchema.validate(record.id);
+    })) {
```
n1k0: How about rejecting at first invalid record encountered instead?

Natim:
An even better solution would be to import only records of the dump with a last_modified value newer than the current collection_timestamp.

src/collection.js
```diff
@@ -787,4 +787,45 @@ export default class Collection {
...
+      return reject("Record has invalid ID.");
+    }
+
+    if (!records.every(record => record.last_modified)) {
+      return reject("Record has no last_modified value.");
```
n1k0: Ditto

src/collection.js
leplatrem: with `some()` ?

test/adapters/IDB_test.js
```diff
@@ -149,5 +149,22 @@ describe("adapter.IDB", () => {
...
+        sandbox.stub(db, "prepare").returns({
+          store: {add() {}},
+          transaction: {
+            get onerror() {},
+            set onerror(onerror) {
+              onerror({target: {error: "transaction error"}});
+            }
+          }
+        });
+        return db.import([{foo: "bar"}])
```
n1k0: Hmm, what if I want to import records and have ids autogenerated for me?

Natim: That's actually not part of the feature we are implementing.

mozmark:
The difficulty with only importing items with lastModified later than the newest value is that you will end up keeping things that should be removed. I'd suggest either 1) Only allow import of a new collection or 2) Allow a collection to be completely replaced if (and only if) the imported data has a last modified value newer than the old collection.

Natim:
Yes that's the idea.

Also you raise an important point, we should probably handle tombstones in our import in order to remove deleted records.

Natim:
- [ ] Rename import to loadDump
 - [ ] Skip records that were already updated (local record ``status: synced`` and ``last_modified`` value more recent than the one in the dump or ``status: updated``).
 - [ ] Result of promise should return the list of modified records.

mozmark:
This looks good to me. I guess we r? @n1k0 when ready.

docs/api.md
```diff
@@ -214,6 +214,29 @@ articles.list({sort: "-title"})
 > - By default, the records are sorted on `last_modified` in descending order.
 > - As mentioned in the [limitations](limitations.md) section, the sort is performed in memory.
 
+## Importing initial records locally
```
n1k0: Nit: Importing a data dump locally

n1k0:
LGTM, r+

docs/api.md
```diff
@@ -214,6 +214,29 @@ articles.list({sort: "-title"})
...
+    title: "baz",
+    last_modified: 1432222889337
+  }
+])
+  .then(records => console.log(records));
+```
+
+> #### Notes
+>
+> - Existing records are replaced if they do not have more recent modifications.
```
rnewman: This…

fx-src/FirefoxStorage.js
```diff
@@ -72,7 +72,12 @@ const statements = {
   "listRecords": `
     SELECT record
       FROM collection_data
-        WHERE collection_name = :collection_name;`
+        WHERE collection_name = :collection_name;`,
+
+  "importData": `
+    REPLACE INTO collection_data (collection_name, record_id, record)
```
rnewman: … is at odds with this. You iterate over the incoming records in a transaction, calling `INSERT OR REPLACE` with each. That will unconditionally discard the existing rows that violate the uniqueness constraint on `record_id`:

```
When a UNIQUE or PRIMARY KEY constraint violation occurs, the REPLACE algorithm deletes pre-existing rows that are causing the constraint violation prior to inserting or updating the current row and the command continues executing normally.
```

fx-src/FirefoxStorage.js
```diff
@@ -197,6 +202,39 @@ export default class FirefoxAdapter extends BaseAdapter {
...
+    let collection_name = this.collection;
+    return Task.spawn(function* () {
+      yield connection.executeTransaction(function* doImport() {
+        for (let record of records) {
+          const params = {
+            collection_name: collection_name,
+            record_id: record.id,
+            record: JSON.stringify(record)
+          };
+          yield connection.execute(statements.importData, params);
```
rnewman: At this point, all conflicting records have been deleted and the incoming records inserted.

fx-src/FirefoxStorage.js
```diff
@@ -197,6 +202,39 @@ export default class FirefoxAdapter extends BaseAdapter {
...
+        }
+        const lastModified = Math.max(...records.map(record => record.last_modified));
```
rnewman: You should short-circuit out of this method if there are no incoming records so that you don't need to handle the empty case here.

fx-src/FirefoxStorage.js
```diff
@@ -197,6 +202,39 @@ export default class FirefoxAdapter extends BaseAdapter {
...
+        const previousLastModified = yield connection.execute(
+          statements.getLastModified, params).then(result => {
+            return result ? result[0].getResultByName('last_modified') : -1;
+          });
+        if (lastModified > previousLastModified) {
+          const params = {
+            collection_name: collection_name,
+            last_modified: lastModified
+          };
+          yield connection.execute(statements.saveLastModified, params);
```
rnewman: So you calculate the max of the new records' `last_modified`, query the existing metadata table to find its stored `last_modified`, compare that to the default max, and then do `INSERT OR REPLACE`.

After doing this bulk insert, the collection is guaranteed to contain data. (You short-circuited when the incoming set of records is empty, didn't you?)

So why not just do

```
UPDATE metadata SET last_modified = (SELECT MAX(last_modified) FROM records WHERE collection = :collection) WHERE collection = :collection
```

— in other words, "make sure my last_modified time matches the data I'm storing". Right?

fx-src/FirefoxStorage.js
rnewman: I think you probably want a two-phase operation here instead. Either

```
SELECT last_modified, record_id FROM records WHERE collection_name = :collection_name AND record_id IN (?, ?, ?, …)
```

then merge in memory and do `INSERT OR REPLACE` with just the records that are newer or not present, or

```
UPDATE … WHERE last_modified < :last_modified
```

for each incoming record, checking the number of modified rows; if none modified, then

```
INSERT OR IGNORE …
```

to insert missing records.

fx-src/FirefoxStorage.js
Natim: > make sure my ``last_modified`` time matches the data I'm storing

That's not exactly what we are doing there.
We are keeping the ``collection_timestamp`` from the last sync except if the records we are importing are more recent that the last sync.

fx-src/FirefoxStorage.js
leplatrem: Hi Richard,
Thank you for this thorough review!

Regarding this point, we can explain the inconsistency between the doc up there and this piece of code.

Indeed, the high-level (and two-phase) checking of existing records versus imported records is done in the `Kinto/collection.js` class, so that it does not have to be duplicated in every adapter (IndexedDB, localstorage and SQLite). Therefore the code is the adapter is rather naive and brutal because its input comes prepared and filtered already.

In this repo we run the collection tests against indexeddb and localstorage. Thanks to your feedback we added the same tests cases in mozilla-central (see https://github.com/Natim/gecko-dev/commit/dc9c6a80e51738c7d4246073fade6879068e6570, @Natim is submitting a patch in BZ https://bugzilla.mozilla.org/show_bug.cgi?id=1227877)

fx-src/FirefoxStorage.js
leplatrem: True, it would save us a hit on SQLite to get the current last modified. But otherwise `Math.max([])` returns `-Infinity`...

fx-src/FirefoxStorage.js
rnewman: > Therefore the code in the adapter is rather naive and brutal because its input comes prepared and filtered already.

Ah, gotcha. Could you clarify the docs/comments for the adapter to note that they don't need to do the timestamp filtering themselves?

PR #269, opened 2015-11-24
Relaxed UUID validation.

n1k0:
We're now accepting uuid formats matching the cliquet spec. But we're still creating UUID4 by default.

r=? @natim @leplatrem @almet

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -2,7 +2,7 @@
 
 import { deepEqual } from "assert";
 
-const RE_UUID = /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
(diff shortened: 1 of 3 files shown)


Natim:
Yep, looks great :+1:

src/utils.js
```diff
@@ -123,15 +124,26 @@ export function partition(array, n) {
...
+/**
+ * Checks if a string is an UUID4, according to RFC4122.
+ *
+ * @param  {String} uuid The uuid to validate.
+ * @return {Boolean}
+ */
+export function isUUID4(uuid) {
+  return RE_UUID4.test(uuid);
+}
+
```
leplatrem: Do we want to keep this since it's now unused?

n1k0: Yeah, will remove.

leplatrem:
r+

PR #271, opened 2015-11-25
[Adaptation of #264] Only resolve from fetch if not timed out

michielbdejong:
This seems to work, based on what I described in https://github.com/Kinto/kinto.js/pull/264#issuecomment-159206686

PR #273, opened 2015-12-01
Fixes #263: Handle network request timeouts. r=@leplatrem

### LINKED ISSUE #263: Handle network request timeouts.

michielbdejong:
This is @n1k0's commit from #264 with my adaptation from #271 squashed into it.

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -19,15 +19,25 @@ export default class HTTP {
   }
 
   /**
+   * Default options.
+   *
+   * @type {Object}
+   */
+  static get defaultOptions() {
+    return {timeout: 5000, requestMode: "cors"};
+  }
+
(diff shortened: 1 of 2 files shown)


n1k0:
LGTM.

PR #274, opened 2015-12-02
Fix JSM compatibility for Firefox adapter

leplatrem:
When implementing more tests cases for the `loadDump()` feature, we faced a couple of problems releated to JSM (ref #273)

This diff appears to solve them properly.

We take any feedback! :)

@mozmark r?

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -797,19 +797,17 @@ export default class Collection {
    */
   loadDump(records) {
     const reject = msg => Promise.reject(new Error(msg));
-    if (!(typeof records.sort == 'function')) {
+    if (!Array.isArray(records)) {
       return reject("Records is not an array.");
     }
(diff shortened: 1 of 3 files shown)


src/collection.js
```diff
@@ -797,7 +797,7 @@ export default class Collection {
    */
   loadDump(records) {
     const reject = msg => Promise.reject(new Error(msg));
-    if (!(records instanceof Array)) {
+    if (!(typeof records.sort == 'function')) {
```
n1k0: `Array.isArray(records)` maybe? Or are you suggesting passed arg is possibly not a *true* array, like `Arguments` or `NodeList`?

leplatrem: I will check, but JSM the first version wasn't working.

Issue #257, opened 2015-11-07
Pass remote in sync() method


leplatrem:
When I built [this small demo](https://github.com/leplatrem/kinto-demo-leaflet), I felt puzzled because authorization `headers` can be passed to `sync()` but not the `remote` server. I would really have expected it there :) 

I looked at the code, and realized it was not an obvious change (because of api instantiation).

Would you find relevant if I spend some time on this?

n1k0:
I hear you. I don't have a sense how critical is it though, but yeah, we could probably create the `Api` instance at `sync()` call time only. On the other hand, it's also cool being able to just call `sync()` without having to pass it the whole server configuration each time, but that's probably not that a big deal.

I'd love seeing a PR for that, but one more time, I'm not sure it's P1.

leplatrem:
>  it's also cool being able to just call sync() without having to pass it

Oh yes of course! What I had in mind is to treat `remote` just like `headers`, i.e. being able to pass them in both places

n1k0:
The more I'm working on the [amo-blocklist-ui](https://github.com/mozilla-services/amo-blocklist-ui), the more I realize we badly need to decouple anything server settings related from the `KintoBase` constructor, and only handle it from `#sync()`.

We should treat a Kinto instance as a local db driver, and collections an abstraction on top of tables. I'd go as far as extracting `#sync()` to its own separated module:

```js
import sync from "kinto/lib/sync";

const kinto = new Kinto();
const articles = kinto.collection("articles");

sync(articles, {
  remote: "http://localhost:9999",
  headers: {...}
});
```

We could also imagine having a `liveSync` version of it at some point:

```js
import liveSync from "kinto/lib/liveSync";

sync(liveSync, settings);
```

Having these things in their own module will greatly help avoiding bloating the `Collection` class, and will probably ease testing by a lot.

It will also help building a more powerful admin UI where server settings could be distincly provided for each collection.

Thoughts?

leplatrem:
It reminds me an idea of @michielbdejong to be able to sync databases of existing applications that already store data in indexeddb.

Otherwise, as a quick feedback I am ok with this change. And in a first we are not obliged to break the current public API of the collection class.

> We could also imagine having a liveSync version of it at some point:

I didn't get that part, but I see the idea: depending of the nature of the first parameter, being able to handle different sync scenarios.

PouchDb has a slightly different approach but with the same idea. You instantiate a client on local or remote, and sync both instances:

```js
[...]

PR #276, opened 2015-12-03
refs #257: Allowing defining the remote endpoint in #sync().

n1k0:
Took the easiest way to allow passing the remote to `#sync()`. Works, but ugly, as I'm mutating the api instance during the time of the sync flow execution :/ That's the price to pay for strict BC.

Later we'll move the whole sync flow to its own module, so that's a temporary situation (famous last words).

r=? @leplatrem

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -46,25 +46,12 @@ For now, our `demo.js` file content is simply:
 
 ```js
 function main() {
-  var db = new Kinto({remote: "https://kinto.dev.mozaws.net/v1"});
+  var db = new Kinto();
   var tasks = db.collection("tasks");
 }
 
 window.addEventListener("DOMContentLoaded", main);
-
 ```
-
-> #### Notes
->
-> - For convenience, we're using a mozilla-hosted version of Kinto at `https://kinto.dev.mozaws.net/v1`; **data pushed to this instance are reset everyday**;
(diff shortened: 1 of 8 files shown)


src/collection.js
```diff
@@ -764,18 +764,31 @@ export default class Collection {
...
-  sync(options={strategy: Collection.strategy.MANUAL, headers: {}, ignoreBackoff: false}) {
+  sync(options={
+    strategy: Collection.strategy.MANUAL,
+    headers: {},
+    ignoreBackoff: false,
+    remote: null,
+  }) {
+    const previousRemote = this.api.remote;
+    if (options.remote) {
+      console.log("using custom remote", options.remote);
```
leplatrem: Add a bit of context like `#sync()` prefix ?

leplatrem:
It looks good to me but I'm sorry, I didn't take the time to review it seriously. Be back next week!

leplatrem:
r+ :)

src/collection.js
michielbdejong: Should kinto.js contain console.log statements? Just noting that this seems to be the first/only console.log statement so far.

n1k0: No it shouldn't. It was probably a debug statement :) I'll drop it.

docs/tutorial.md
```diff
@@ -76,7 +63,7 @@ We want to listen to form submission events to add tasks into our local database
 
 ```js
 function main() {
-  var db = new Kinto({remote: "https://kinto.dev.mozaws.net/v1"});
```
almet: It's still possible, right?

n1k0: Yeah but no more encouraged

PR #278, opened 2015-12-04
Add a way to limit the number of records retrieved by Kinto.js in one…

ferjm:
… sync run


michielbdejong:
@n1k0 as discussed on Vidyo, we want to do this with the next page token. Cool! :)

n1k0:
Superseded by #279.

PR #279, opened 2015-12-04
Fixes #277: Limit number of changes to read from a sync() operation.

### LINKED ISSUE #277: Add a way to limit the number of records retrieved by Kinto.js in one sync run
We can probably use the `_limit` param https://github.com/mozilla-services/syncto/blob/master/syncto/views/collection.py#L37

n1k0:
Supersedes #278, WiP:

Reference: [Kinto server docs about paginated changes retrieval](http://kinto.readthedocs.org/en/latest/api/1.x/synchronisation.html#paginated-changes)

- [x] Support the `_limit` query string parameter when fetching changes from the server
- [x] Check for the `Next-Page` header value in HTTP responses
- [x] Recursively retrieve all next pages until there's none, aggregate the results
- [x] Add support for a `maxPages` option so you can limit the number of pages to retrieve
- [x] Expose methods to store and retrieve any collection metadata to `BaseAdapter`; this is required to persist the next page url along the `lastModified` value
- [ ] Store the last next page url in collection db metas, in case the  `maxPages` option has been used and the whole synchronization flow is therefore not fully completed
- [ ] When a next page url is stored, we should *not* bump the collection `lastModified` value; we should do so as soon as there's no `Next-Page` header sent by the server
- [ ] Add a check in `#sync()` if a next page url to crawl is remaining, and start the change retrieval process from there
- [ ] If a `412` status code is received during the paginated changes retrieval flow, restart it from the beginning
- [ ] Ensure any stored next page url is cleared when `#clear()` and `#resetSyncStatus()` are called
- [ ]  Update docs
 
/cc @michielbdejong @Natim @ferjm

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -191,9 +191,15 @@ export default class Api {
         if (res.status === 304) {
           return {
             lastModified: options.lastModified,
+            token: null,
             changes: []
           };
         }
+
+        // Extract pagination token, if any
+        const nextPage = res.headers.get("Next-Page");
+        const token = nextPage ? urlParse(nextPage, true).query._token : null;
+
         // XXX: ETag are supposed to be opaque and stored «as-is».
         // Extract response data
         let etag = res.headers.get("ETag");  // e.g. '"42"'
@@ -208,7 +214,7 @@ export default class Api {
           throw Error("Server has been flushed.");
         }
 
-        return {lastModified: etag, changes: records};
+        return {lastModified: etag, token, changes: records};
       });
   }
 
--- a/src/api.js
+++ b/src/api.js
@@ -156,6 +156,48 @@ export default class Api {
       });
   }
 
+  _fetchChangeUrl(results, url, headers, options) {
+    return this.http.request(url, {headers})
+      .then(res => {
+        // If HTTP 304, nothing has changed
+        if (res.status === 304) {
+          return results;
+        } else if (res.status=== 412) {
+          // XXX what now?
+          throw new Error("Changed meanwhile.");
+        }
+
+        // Extract pagination token, if any
+        const nextPage = res.headers.get("Next-Page") || null;
+
+        // XXX: ETag are supposed to be opaque and stored «as-is».
+        // Extract response data
+        let etag = res.headers.get("ETag");  // e.g. '"42"'
+        etag = etag ? parseInt(unquote(etag), 10) : options.lastModified;
+        const records = res.json.data;
+
+        // Check if server was flushed
+        const localSynced = options.lastModified;
+        const serverChanged = etag > options.lastModified;
+        const emptyCollection = records ? records.length === 0 : true;
+        if (localSynced && serverChanged && emptyCollection) {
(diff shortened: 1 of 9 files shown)


src/api.js
```diff
@@ -162,26 +164,42 @@ export default class Api {
...
           return {
             lastModified: options.lastModified,
+            token: null,
             changes: []
           };
         }
+
+        // Extract pagination token, if any
+        const nextPage = res.headers.get("Next-Page");
+        const token = nextPage ? urlParse(nextPage, true).query._token : null;
```
Natim: You should rather save the full URL (including other query parameters)

n1k0: Ah. Okay, feels a little odd but I get your point. Will update.

n1k0:
Ok it's a little late, but I'd still like feedback from @michielbdejong @ferjm @Natim on this when you get time :)

michielbdejong:
I think it's simpler to force `maxPages` to 1. It's easy enough to call `.sync` multiple times. And then we get the advantage of reduced memory consumption.

Without a `maxPages`, I'm not sure what the use case for `fetchLimit` would be? If I want 150 records then I can just do `fetchLimit: 150` instead of `fetchLimit: 50, maxPages: 3`, right?

IMO, if the goal is to limit memory consumption, then all data should at least be written to IDB after each page.

n1k0:
Alternative approach as discussed on IRC:

```js
const coll = kinto.collection("foo");
coll.sync()
 .then(result => {
    if (result.next) {
      return coll.syncNext(result);
    } else {
     return result;
    }
  });
```

Crawling result pages would be always explicitly performed by users until the list is exhausted or as soon as they feel they have enough information. They could store somewhere the `result` value and call `syncNext` when they feel it's a better time to.

That's really nice on the paper, now need to figure out the possible limitations of such a simple approach. Feel free to provide feedback on this idea.

Natim:
> If I want 150 records then I can just do fetchLimit: 150 instead of fetchLimit: 50, maxPages: 3, right?

It doesn't do the same thing. ``fetchLimit: 150`` equals ``fetchLimit: 150, maxPages: 1`` one may want to limit 50 items per pages.

michielbdejong:
> one may want to limit 50 items per pages.

But the app developer doesn't get to see the pages, only kinto.js internally sees them.

michielbdejong:
Is this PR superseded by https://github.com/Kinto/kinto.js/pull/299 ? I don't remember, I'm just seeing that I started that one after the last change was made to this one.

leplatrem:
It's tempting to close this PR (since Github now has a way to reopen).

Indeed, a great amount of the work done here was implemented in kinto-client : https://github.com/Kinto/kinto-client/blob/v0.6.0/src/collection.js#L297-L333 : `listRecords()` accepts a `pages` parameter.

Some code may still be relevant, but the merge of the current state requires a lot of rework. 

What do you think if we close and reopen when revamped ?

n1k0:
wfm

PR #286, opened 2015-12-09
Extend initial record information on update().

n1k0:
WiP


Natim:
We need to add a test that explain how we do when we want to remove an attribute that is present remotely.

n1k0:
Now I think about it, we should probably rather introduce a new `#patch()` method we would just do this, instead of mofifying the behaviour of `#update()`.

PR #291, opened 2015-12-15
Add dist with minimalist version, without polyfills

leplatrem:
The build weights 59K compared to 790K for unminified with polyfills

I could not come out without a good name, and ended up with «`noshim`».

What do you think ?


n1k0:
Thanks! `noshim` sounds right. r+

Would you mind pushing a new release so we can benefit from a CDN url asap? :)

PR #294, opened 2015-12-16
Fix parsing of ES7 code in esdoc

leplatrem:
Used to fail with this:

```
error: could not parse the following code. if you want to use ES7, see esdoc-es7-plugin(https://github.com/esdoc/esdoc-es7-plugin)
/home/mathieu/Code/Mozilla/kinto.js/src/collection.js
676|         // Merge outgoing conflicts into sync result object
677|         syncResultObject.add("conflicts", conflicts);
678|         // Reflect publication results localy
679|         const missingRemotely = skipped.map(r => ({...r, deleted: true}));
680|         const toApplyLocally = published.concat(missingRemotely);
681|         return Promise.all(toApplyLocally.map(record => {
682|           if (record.deleted) {

```

@n1k0 r?

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -31,8 +31,8 @@ In the generated `CHANGELOG.md`, replace `HEAD` by `vX.Y.Z` in full changelog an
 
 ### Publish on NPM
 
-    $ npm build
-    $ npm dist
+    $ npm run build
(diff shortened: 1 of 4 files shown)


n1k0:
r+ with conflicts resolved.

PR #299, opened 2015-12-29
Fixes #277: Add fetchLimit and fetchToken support to Collection#sync

### LINKED ISSUE #277: Add a way to limit the number of records retrieved by Kinto.js in one sync run

michielbdejong:
To do:

* [x] tests for token param
* [x] docs
* [x] add pagination
* [x] test for parsing of `Next-Page` header
* [x] test for not saving `last_modified` if `Next-Page` header is present

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -181,12 +181,20 @@ export default class Api {
    */
   fetchChangesSince(bucketName, collName, options={lastModified: null, headers: {}}) {
     const recordsUrl = this.endpoints().records(bucketName, collName);
-    let queryString = "";
+    let queryString = "?_sort=last_modified";
     const headers = Object.assign({}, this.optionHeaders, options.headers);
 
     if (options.lastModified) {
-      queryString = "?_since=" + options.lastModified;
+      queryString += "&_since=" + options.lastModified;
       headers["If-None-Match"] = quote(options.lastModified);
+    }
+
+    if (options.limit) {
+      queryString += "&_limit=" + options.limit;
+    }
+
+    if (options.token) {
+      queryString += "&_token=" + encodeURIComponent(options.token);
     }
 
     return this.fetchServerSettings()
@@ -199,6 +207,28 @@ export default class Api {
             changes: []
           };
         }
+
+        // Example response header:
+        // Next-Page:"https://syncto.dev.mozaws.net/v1/buckets/d83134db8e5cf927\
+        // 624742e63e5e9f8f/collections/history/records?_sort=last_modified&_li\
+        // mit=50&_token=1451832797410%3A50"
+        function getQueryParam(url, key) {
+          const queryStartPos = url.indexOf("?");
+          if (queryStartPos === -1) {
+            return;
+          }
+          const params = url.substring(queryStartPos + 1).split("#")[0].split("&");
+          for (let i = 0; i < params.length; i++) {
+            const pairs = params[i].split("=");
+            if (decodeURIComponent(pairs.shift()) == key) {
+              return decodeURIComponent(pairs.join("="));
+            }
+          }
+        }
+
+        const nextPageHeader = res.headers.get("Next-Page");
+        const fetchToken = (nextPageHeader ? getQueryParam(nextPageHeader, "_token") : null);
+
         // XXX: ETag are supposed to be opaque and stored «as-is».
         // Extract response data
         let etag = res.headers.get("ETag");  // e.g. '"42"'
@@ -213,7 +243,7 @@ export default class Api {
           throw Error("Server has been flushed.");
         }
 
-        return {lastModified: etag, changes: records};
+        return {lastModified: etag, changes: records, fetchToken};
       });
   }
 
--- a/src/collection.js
+++ b/src/collection.js
@@ -552,7 +552,8 @@ export default class Collection {
       .then(syncResultObject => {
         syncResultObject.lastModified = changeObject.lastModified;
         // Don't persist lastModified value if any conflict or error occured
-        if (!syncResultObject.ok) {
+        // or if pagination is still in progress.
+        if (!syncResultObject.ok || syncResultObject.fetchToken) {
           return syncResultObject;
         }
         // No conflict occured, persist collection's lastModified value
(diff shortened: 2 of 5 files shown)


Natim:
Don't you want to save the ``Next-Page`` url somewhere to let the user go through all the pages if they want to?

michielbdejong:
@Natim this is taken care of by sorting the records by last-modified, and using `_since` in the next query.

Natim:
Oh great :+1:

michielbdejong:
Found a problem with this PR: when many records have the same `last_modified` record, some are skipped. In the case of fetching records from Syncto/FxSync, all records have the timestamp of when they were uploaded, so that's a problem - we still need to pagination. Will work on that tomorrow.

michielbdejong:
Problem from previous comment fixed and tests added.

r? @n1k0 @Natim

src/api.js
```diff
@@ -182,6 +190,28 @@ export default class Api {
...
+            return;
+          }
+          const params = url.substring(queryStartPos + 1).split("#")[0].split("&");
+          for (let i = 0; i < params.length; i++) {
+            const pairs = params[i].split("=");
+            if (decodeURIComponent(pairs.shift()) == key) {
+              return decodeURIComponent(pairs.join("="));
+            }
+          }
+        }
```
n1k0: We should probably leverage nodejs' `url` module here: https://nodejs.org/api/url.html#url_url_parse_urlstr_parsequerystring_slashesdenotehost

src/api.js
```diff
@@ -182,6 +190,28 @@ export default class Api {
...
+
+        const nextPageHeader = res.headers.get("Next-Page");
+        const fetchToken = (nextPageHeader ? getQueryParam(nextPageHeader, "_token") : null);
```
n1k0: I remember @Natim claiming we should not store the token but the whole url instead, though I can't remember the reasoning. Hint @Natim?

src/collection.js
```diff
@@ -626,13 +629,27 @@ export default class Collection {
...
       headers: options.headers
     })
+      // Deal with pagination
+      .then(changes => {
+        if (changes.fetchToken) {
+          syncResultObject.fetchToken = changes.fetchToken;
+        }
+        // TODO: deal with the case where records disappear from
+        // pages already fetched, due to updates to the server
+        // while the paginating query series is in progress.
```
n1k0: As far as I can remember we should get a 412 from the server whenever this happens. @Natim can you confirm this?

src/api.js
Natim: Because that's the protocol, you cannot suppose that the other query string parameters won't be altered as well.

n1k0:
It's been a long time I didn't work on this feature, but I'm surprised this implementation proposal differs from what I recall from past discussions about it.

I remember we decided to treat each page results as a whole chunk of changes where once processed, we could take the highest last_modified value from the records in that page and mark the local collection as synced using that value. That would obviously imply ensuring the changes are listed by last_modified ASC. 

Am I confused with my memories, or did you encounter a flaw in that reasoning so you needed to proceed otherwise?

src/collection.js
Natim: There are two strategies here.
The first one is to send the ``If-Match`` header and expect a 412 in case the page fails. The problem with this strategy is that you cannot really now how long you will have to restart the process in case the collection change a lot.

The second strategy (and the one I think we should take for Kinto.js) is not to send the If-Match header and to follow the pagination til this end.

And then ask the server all the changes that happened since the beginning of the fetching (taking the Etag from the first page of the fetching)

michielbdejong:
Thanks for your comments!

@n1k0 yeah, the approach of only looking at Last-Modified [didn't work out](https://github.com/Kinto/kinto.js/pull/299#issuecomment-168518138), so using `_token` is necessary.

@Natim ok, if we need to use the exact URL form the `Next-Page` header then that's a bigger change in api.js; it would make sense then to pass something like `forceURL` to `fetchChangesSince`, to override other options and defaults.

I will revise the patch.

leplatrem:
I suggest (and can help to do so) to write a couple of functional tests. Especially because of the assumptions made about the `Next-Page` header which was supposed to be opaque.

michielbdejong:
Yes, I should continue work on this PR. I just rebased it to see how Travis would react.

About ordering, I used oldest-first because that way, if you download half the pending records and then stop for whatever reason, you can already increase the lastModified value of the collection, which might give better performance on resumes.

Note that the download order is unrelated to the default order of `Collection#list` (where we use newest-first).

leplatrem:
> I used oldest-first because that way, if you download half the pending records and then stop for whatever reason, you can already increase the lastModified value of the collection

:+1:

src/api.js
michielbdejong: Right, so when retrieving a subsequent page, we should not allow any options to be passed to [`fetchChangesSince`](https://github.com/michielbdejong/kinto.js/blob/fetch-limit/src/api.js#L182), we should just use the whole URL from the previous response, including its exact query string. In any case, that code now lives in [kinto-client](https://github.com/kinto/kinto-client), so should move/rebase this PR.

leplatrem:
As I said in #277: 

It's tempting to close this PR (since Github now has a way to reopen).

Indeed, a great amount of the work done here was implemented in kinto-client : https://github.com/Kinto/kinto-client/blob/v0.6.0/src/collection.js#L297-L333 : `listRecords()` accepts a `pages` parameter.

Some code may still be relevant, but the merge of the current state requires a lot of rework. 

What do you think if we close and reopen when revamped ?

n1k0:
I'll let @michielbdejong tell us :)

michielbdejong:
I remember Sean Lee was considering whether this work is still needed for Smart TV 2.6 or not, and haven't heard what the conclusion was. @weilonge, do you know?

If the TV team still need this, it can be reopened/rebased; let's close it now to clean up outdated PRs from the list.

weilonge:
Our partner has known the record limitation and how to change the preferences, but we don't get their strong requirement to fix the limitation.

Issue #153, opened 2015-09-09
Add collection hooks.


almet:
To be able to sign the collection before sending it to the kinto server, a hook mechanism would help.

For instance:

```js
var db = new Kinto(...);
var mycollection = db.collection('mycollection');
mycollection.on('sync-local-changes', callback);
```

I'm not sure about the name of the event, but that's the idea :)

leplatrem:
Duplicate of #154

PR #302, opened 2016-01-20
Add collection transformers.

almet:
This adds the concept of collection transformers.

I haven't updated the documentation because I'm not sure about the concepts we want to expose, as mentioned in #154 & #153.

The documentation states that transformers should be able to encode and decode, but I'm not sure that should be the case for what's implemented here as "collection transformers". I'm happy to have any feedback on the naming (and of course the implementation).

If we chose to go with the "transformers" approach (and still not have decode / encode), then the documentation will need to be updated to explain what a collection should be.

For what it's worth, the first version of this code was implemented with "decode" and "encode" functions, but I couldn't see why we were needing them, so I removed them and passed a function instead.

I'm very interested by any feedback on this, so if anyone has time, don't hesitate :) cc @n1k0  @michielbdejong

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -570,7 +570,7 @@ kinto.events.on("deprecated", function(event) {
 
 Transformers are basically hooks for encoding and decoding records, which can work synchronously or asynchronously.
 
-For now, only *remote transformers* have been implemented, but there are plans to implement local transformers and collection transformers in a next version.
+For now, only *remote transformers* and *collection transformers* have been implemented, but there are plans to implement local transformers in a next version.
 
 ### Remote transformers
 
@@ -623,7 +623,8 @@ coll.create({title: "foo"}).then(_ => coll.sync())
 
 ### Local transformers
 
-In a near future, Kinto.js will provide transfomers aimed at providing facilities to encode and decode records when persisted locally.
+In a near future, Kinto.js will provide transfomers aimed at providing facilities to encode and decode records when persisted locally. If you are interested into
+making this happen, [don't hesitate to connect with us](https://kiwiirc.com/client/irc.freenode.net/?#kinto).
 
 ### Async transformers
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -570,7 +570,7 @@ kinto.events.on("deprecated", function(event) {
 
 Transformers are basically hooks for encoding and decoding records, which can work synchronously or asynchronously.
 
-For now, only *remote transformers* and *collection transformers* have been implemented, but there are plans to implement local transformers in a next version.
+For now, only *remote transformers* and *hooks* have been implemented, but there are plans to implement local transformers in a next version.
 
 ### Remote transformers
 
@@ -623,8 +623,7 @@ coll.create({title: "foo"}).then(_ => coll.sync())
 
 ### Local transformers
 
-In a near future, Kinto.js will provide transfomers aimed at providing facilities to encode and decode records when persisted locally. If you are interested into
-making this happen, [don't hesitate to connect with us](https://kiwiirc.com/client/irc.freenode.net/?#kinto).
+In a near future, Kinto.js will provide transfomers aimed at providing facilities to encode and decode records when persisted locally.
 
 ### Async transformers
 
--- a/docs/extending.md
+++ b/docs/extending.md
@@ -46,7 +46,6 @@ Then create the Kinto object passing a reference to your adapter class:
 const kinto = new Kinto({adapter: MyAdapter});
 ```
 
-<<<<<<< bb86b694d73101136b4c6c39b47bcf8e27afd4f0
 Read the `BaseAdapter` class [source code](https://github.com/Kinto/kinto.js/blob/master/src/adapters/base.js) to figure out what needs to be implemented exactly. The [IDB](https://github.com/Kinto/kinto.js/blob/master/src/adapters/IDB.js) adapter is also worth a read if you need guidance writing your own.
 
 ## Supporting transactions
@@ -97,13 +96,13 @@ Hooks can be called to extend the behaviour of kinto. So far it is only possible
 To install a hook, you need to pass it to the collection:
 
 ```
-function doSomething(payload) {
+function doSomething(payload, collection) {
   // Do something with the payload here.
 };
 
 let collection = db.collection(collectionName, {
-    "hooks": {
-      "incoming-changes": doSomething
-    }
+  "hooks": {
+    "incoming-changes": doSomething
+  }
 });
 ```
--- a/docs/extending.md
+++ b/docs/extending.md
@@ -46,6 +46,7 @@ Then create the Kinto object passing a reference to your adapter class:
 const kinto = new Kinto({adapter: MyAdapter});
 ```
 
+<<<<<<< bb86b694d73101136b4c6c39b47bcf8e27afd4f0
 Read the `BaseAdapter` class [source code](https://github.com/Kinto/kinto.js/blob/master/src/adapters/base.js) to figure out what needs to be implemented exactly. The [IDB](https://github.com/Kinto/kinto.js/blob/master/src/adapters/IDB.js) adapter is also worth a read if you need guidance writing your own.
 
 ## Supporting transactions
@@ -84,4 +85,25 @@ db.list()
       transaction.update(Object.assign({}, existing, {foo: "bar"}));
     }, {preload});
   });
+```
+
+## Hooks
+
+Hooks can be called to extend the behaviour of kinto. So far it is only possible to hook when incoming changes are to be applied.
+
+- `incoming-changes` hooks are called just after new changes are retrieved, and
+  before these changes are reflected locally.
+
+To install a hook, you need to pass it to the collection:
+
+```
+function doSomething(payload) {
+  // Do something with the payload here.
+};
+
+let collection = db.collection(collectionName, {
+    "hooks": {
+      "incoming-changes": doSomething
+    }
+});
 ```
--- a/src/collection.js
+++ b/src/collection.js
@@ -6,6 +6,7 @@ import { waterfall } from "./utils";
 import { v4 as uuid4 } from "uuid";
 import { deepEquals, isUUID, pFinally } from "./utils";
 
+<<<<<<< e030e48ddf01fc99d081510a097865cdf55af165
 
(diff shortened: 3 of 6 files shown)


almet:
@n1k0 can you have a look at the code and tell me if that looks okay? (docs needs an update). Thanks!

docs/api.md
```diff
@@ -545,7 +545,7 @@ kinto.events.on("deprecated", function(event) {
 
 Transformers are basically hooks for encoding and decoding records, which can work synchronously or asynchronously.
 
-For now, only *remote transformers* have been implemented, but there are plans to implement local transformers and collection transformers in a next version.
+For now, only *remote transformers* and *collection transformers* have been implemented, but there are plans to implement local transformers in a next version.
```
n1k0: Nit: we should reuse the term *hooks*

src/collection.js
```diff
@@ -230,6 +237,52 @@ export default class Collection {
...
+   *
+   * @param  {Object|undefined} hooks
+   * @return {Object}
+   */
+  _validateHooks(hooks) {
+    if (typeof hooks === "undefined") {
+      return {};
+    }
+    if (Array.isArray(hooks)) {
+      throw new Error("hooks should be an object, not an array.");
```
n1k0: How could this ever happen as this is a private function?

docs/api.md
almet: yes, as I said, this doc is completely out of date. I plan to use hooks here.

src/collection.js
almet: this function is called at initialization, with the data entered by the user in collection.hooks.

src/collection.js
```diff
@@ -537,6 +590,7 @@ export default class Collection {
         if (!syncResultObject.ok) {
           return syncResultObject;
         }
+        // XXX import of data should be done after conflict handling.
```
n1k0: I don't get this comment

src/collection.js
```diff
@@ -633,12 +687,22 @@ export default class Collection {
       lastModified: options.lastModified,
       headers: options.headers
     })
+      .then(changes => this.applyHook("incoming-changes", {changes}))
       // Reflect these changes locally
-      .then(changes => this.importChanges(syncResultObject, changes))
+      .then(payload => this.importChanges(syncResultObject, payload.changes))
```
n1k0: Nit: can be written that way

```js
.then(({changes}) => this.importChanges(syncResultObject, changes))
```

src/collection.js
```diff
@@ -633,12 +687,22 @@ export default class Collection {
...
       // Handle conflicts, if any
       .then(result => this._handleConflicts(result, options.strategy));
   }
 
+  applyHook(hookName, payload) {
+    if (typeof this.hooks[hookName] == "undefined") {
+      return Promise.resolve(payload);
+    }
+    return waterfall(this.hooks[hookName].map(hook => {
+      return record => hook(payload, this);
```
n1k0: [not finished reading the whole patch, parking this thought for now]: could we avoid passing `this` as a `hook` dependency here?

test/collection_test.js
```diff
@@ -900,6 +941,42 @@ describe("Collection", () => {
         }));
       });
 
+      describe("incoming changes hook", () => {
+        it("should be called", () => {
+          let hookCalled = false;
+          articles = testCollection({
+            hooks: {
+              "incoming-changes": [
+                function(payload, collection) {
```
n1k0: Ok, so I suggest we don't pass collection as an argument, because most often the user will have an instance at hand already.

test/collection_test.js
```diff
@@ -900,6 +941,42 @@ describe("Collection", () => {
...
+                  return payload;
+                }
+              ]
+            }
+          });
+
+          return articles.pullChanges(result)
+            .then(_ => {
+              expect(hookCalled).to.be.true;
+            });
```
n1k0: Nit: can be written on a single line

test/collection_test.js
```diff
@@ -900,6 +941,42 @@ describe("Collection", () => {
...
+        });
+
+        it("should reject the promise if the hook raises", () => {
```
n1k0: nit: s/raises/throws ;)

n1k0:
Looks good so far. As hooks seem to support returning promises (because you use `waterfall`), we should probably add some tests for it.

src/collection.js
almet: Currently, we import the data *first* and then we check if we have conflicts. I would have expected to retrieve everything, check for conflicts (and solve them) and finally import.

src/collection.js
almet: it's done on purpose: the hook actually need a reference to the collection to work properly (with my use case in mind, where we need to retrieve the list of records in the database)

test/collection_test.js
almet: Yeah, that could also work without passing it it seems.

src/collection.js
n1k0: I see

src/collection.js
n1k0: That makes sense, but is probably broader than the scope of this patch; could you please file an issue about it?

docs/api.md
```diff
@@ -598,7 +598,8 @@ coll.create({title: "foo"}).then(_ => coll.sync())
 
 ### Local transformers
 
-In a near future, Kinto.js will provide transfomers aimed at providing facilities to encode and decode records when persisted locally.
+In a near future, Kinto.js will provide transfomers aimed at providing facilities to encode and decode records when persisted locally. If you are interested into
+making this happen, [don't hesitate to connect with us](https://kiwiirc.com/client/irc.freenode.net/?#kinto).
```
leplatrem: nit: I would rather point to contributing page instead

docs/api.md
leplatrem: Also, I think it could be worth mentioning this feature in the «extending kinto.js» docs, no?

docs/api.md
n1k0: +1

almet: Then I need to update the contributing page to have a "how to contact us" link :)

src/collection.js
almet: I've updated the code to act like you say, but it means that when registering the collection hooks I need to now register a "factory" to access the collection. Something like this:

```js
function signatureVerifierFactory(collection) {
  return function validateCollectionSignature(payload) {
    // Access collection here.
  };
};
```

That's doable, but I wonder if it's the API we want to expose?

almet:
rebased.

almet:
@n1k0 last r?

docs/extending.md
```diff
@@ -46,6 +46,7 @@ Then create the Kinto object passing a reference to your adapter class:
 const kinto = new Kinto({adapter: MyAdapter});
 ```
 
+<<<<<<< bb86b694d73101136b4c6c39b47bcf8e27afd4f0
```
n1k0: ?

docs/extending.md
```diff
@@ -85,3 +86,24 @@ db.list()
...
+```
+function doSomething(payload) {
+  // Do something with the payload here.
+};
+
+let collection = db.collection(collectionName, {
+    "hooks": {
+      "incoming-changes": doSomething
+    }
+});
```
n1k0: Nit: 2 spaces indentation

src/collection.js
```diff
@@ -310,6 +315,52 @@ export default class Collection {
...
+   * Validate the passed hook is correct.
+   *
+   * @param {Array|undefined} hook.
+   * @return {Array}
+   **/
+  _validateHook(hook) {
+    if (!Array.isArray(hook)) {
+      throw new Error("A hook definition should be an array of functions.");
+    }
+    return hook.map(fun => {
```
n1k0: Ubernit: `fn` is more common in js land, and in this project.

src/collection.js
n1k0: Sorry, I'm missing previous context it seems. If you want to pass the collection instance as a second arg to avoid creating factories, that's a very good reason for doing so :)

test/collection_test.js
```diff
@@ -157,6 +159,45 @@ describe("Collection", () => {
       });
     });
 
+    describe("hooks registration", () => {
+      function registerHooks(hooks) {
+        return new Collection(TEST_BUCKET_NAME, TEST_COLLECTION_NAME, api, {
+          hooks,
+          adapter: IDB
```
n1k0: Nit: Isn't this the default?

test/collection_test.js
```diff
@@ -995,6 +1036,40 @@ describe("Collection", () => {
...
+                function(payload) {
+                  hookCalled = true;
+                  return payload;
+                }
+              ]
+            }
+          });
+
+          return articles.pullChanges(result)
+            .then(_ => {expect(hookCalled).to.be.true;});
```
n1k0: Nit: unnecessary curly braces. Also, always prefer `eql(true)` for boolean asserts.

n1k0:
I couldn't find tests for using hooks returning a Promise. Did I miss something?

docs/extending.md
almet: Ah, this looks like a trace of a rebase :)

src/collection.js
almet: Okay, going back in the other direction!

test/collection_test.js
almet: This is how it's done in the tests at other locations, I just copied it over. Could be addressed as a separate issue I guess.

n1k0: There are existing tests for async transformers [here](https://github.com/Kinto/kinto.js/blob/master/test/collection_test.js#L930-L944).

almet:
I've updated the pull request:
- [x] addressed your nits;
- [x] added new tests ensuring that the hooks are chainable;
- [x] added new tests ensuring that the return value of the hooks is used;
- [x] pass the current collection as the second argument of the hooks + tests.

Let me know if there is anything else needed in order to land this.

test/collection_test.js
almet: It actually breaks with `No adapter provided`.

n1k0: I suspect that's because we're never expected to use Collection directly, only through the Kinto instance... Ok leave it that way, no big deal.

n1k0:
r+

PR #303, opened 2016-01-21
Add support for transactions at the adapter level (fixes #16, fixes #93, fixes #172, fixes #177)

### LINKED ISSUE #16: Perform batch local updates within a single transaction

### LINKED ISSUE #93: Set up a .babelrc
https://babeljs.io/docs/usage/babelrc/

This also might ease building dist files (concat + minification)

### LINKED ISSUE #172: Bulk operations
* When data comes in from the server, instead of doing one IDB transaction per change (see https://github.com/Kinto/kinto.js/blob/master/src/adapters/IDB.js#L110), do only one IDB transaction, containing all the updates (e.g. importing 1000 records into an empty database)
* For `Collection#create`, `Collection#update` and `Collection#delete`, if an Array is passed instead of an object, apply the operation to each element of this array, but in one single IDB transaction.

Background:
This code:
````js
var kinto = new Kinto();
var coll = kinto.collection('foo');
var requests = [];
for (var i=0; i<1000; i++) {
  requests.push(coll.create({ num: i }));
}
var startTime = new Date().getTime();
console.log("Starting");
Promise.all(requests).then(results => {
  console.log("Finished, ms:", new Date().getTime() - startTime)
}, err => console.error(err));
````
Outputs:
````
Object { Symbol(record)_3.999wfjax45: Object }
Starting
Finished, ms: 2144
````
It takes more than 2000 milliseconds on B2G-Desktop on my Macbook to store 1000 items in Kinto.js, so that's a lot, especially if we translate that to low-end devices on which our FxOs Sync app will have to run.

### LINKED ISSUE #177: Expose transactions from Adapters

### PR DESCRIPTION
**Pull request by n1k0**
Currently we have ~50% percent of performance improvement :)

In order to be merged, this PR would require:
* [x] Migration of Firefox adapter to new API
* [x] Get rid of localstorage
* [x] Documentation update
* [x] File an issue to get rid of `Object.assign` using rest-object spread plugin - #304
* [x] Docstrings for `execute()`
* [x] Check the demo :)
* [x] Add a new section with breaking changes (Upgrade/changelog)
* [x] Check that documentation does not mention anymore anything like «*kinto.js does not handle transactions*»
* [x] Push to try
* [x] Add a test in firefox storage to assert rollbacks
* [ ] Review of mark or/and richard
* [x] Get rid of atomic operations in adapters
* [x] Clarify how transactions errors should be presented in sync result
* [x] Turn the `TransactionProxy` class into a function
* [x] Check all package npm commands keep running properly

### CODE REVIEW

File: src/collection.js [hunk 1]
```diff
@@ -521,8 +504,36 @@ export default class Collection {
...
+      return this._decodeRecord("remote", change);
     }))
+      .then(decodedChanges => {
+        const changeIds = decodedChanges.map(change => change.id);
+        // XXX: this is not efficient at all.
+        return this.list({}, {includeDeleted: true})
+          .then(res => {
+            const existingRecords = res.data.filter(record => {
+              return changeIds.indexOf(record.id) !== -1;
+            });
```
leplatrem: Since we later build a dictionnary based on ids anyway, we don't need this filter this. I had significant performance improvement with removing this filtering.

src/collection.js
```diff
@@ -521,8 +504,36 @@ export default class Collection {
...
-      return this._importChange(change);
+      if (change.deleted) {
+        return Promise.resolve(change);
+      }
+      return this._decodeRecord("remote", change);
     }))
+      .then(decodedChanges => {
+        const changeIds = decodedChanges.map(change => change.id);
+        // XXX: this is not efficient at all.
+        return this.list({}, {includeDeleted: true})
```
leplatrem: I had significant performance improvements by specifiying `{order: null}` as list params (with appropriate safety checks in `utils.js`)

src/collection.js
n1k0: Yes, we'll reintroduce it when we'll have native idb list() filtering capabilities.

src/collection.js
n1k0: Okay, could you please commit your changes?

docs/upgrading.md
```diff
@@ -0,0 +1,30 @@
+# Upgrading
+
+This page lists the breaking API changes between major Kinto.js version, as well as upgrade guides.
+
+## 1.x to 2.x
+
+Kinto.js 2.x introduces transactions, and the Adapter base interface has been updated accordingly:
```
leplatrem: I would introduce this part with the comment you put at the end «*This is internal adapter stuff*».

docs/upgrading.md
```diff
@@ -0,0 +1,30 @@
+# Upgrading
+
+This page lists the breaking API changes between major Kinto.js version, as well as upgrade guides.
+
+## 1.x to 2.x
+
```
leplatrem: Reminder: Local storage adapter removed!

leplatrem:
Reminder: `LocalStorage` is mentionned in Kinto constructor docstring

fx-src/FirefoxStorage.js
```diff
@@ -144,47 +144,58 @@ export default class FirefoxAdapter extends BaseAdapter {
...
   create(record) {
-    const params = {
-      collection_name: this.collection,
-      record_id: record.id,
-      record: JSON.stringify(record)
-    };
-    return this._executeStatement(statements.createData, params)
-               .then(() => record);
+    return this.execute(transaction => transaction.create(record))
+      .then(() => record);
```
leplatrem: reminder: remove those atomic methods as well

leplatrem:
Pushed to try: https://treeherder.mozilla.org/#/jobs?repo=try&revision=2b44eb54d3ce

docs/upgrading.md
```diff
@@ -0,0 +1,34 @@
+# Upgrading
+
+This page lists the breaking API changes between major versions of Kinto.js, as well as upgrade tips.
+
+## 1.x to 2.x
+
+Kinto.js 2.x introduces general usage of transactions in database adapters. This changes doesn't impact the `Collection` API, so most users shouldn't be impacted by this change.
```
leplatrem: nit: this change doesn't

leplatrem:
Try tests are green!

r+ :)

n1k0:
@mozmark we're now about to land this patch, but would really appreciate your opinions on it, especially the modifications we made to the FirefoxAdapter.

FYI we submitted a try build with the changes and everything works as expected, though we have missed something. Also, we are fully aware this is a BC breaks so next released version would be a 2.0.0. Your feedback would be highly valuable, but we could also land this as is and iterate later when you're about to actually update the FFx client; your call :)

n1k0:
I'm gonna land this because it gets in the way with other stuff we have in the pipe. @mozmark feel free to give it a look and provide feedback anyway after it's merged.

michielbdejong:
\o/

docs/limitations.md
```diff
@@ -14,11 +14,11 @@ You can read more about Kinto features [here](http://kinto.readthedocs.org).
...
 Kinto.js won't try to outsmart you as a conflict resolver. Instead, it provides very explicit ways to know about conflicts, and [simple helpers to handle them](api.md#resolving-conflicts).
 
-## Transactions
+## Transactions & rollback
 
-Ideally, we should wrap the whole synchronization flow related operations within a single transaction. While that's a goal and [part of our roadmap](https://github.com/Kinto/kinto.js/issues/16), right now this isn't implemented.
+Ideally, we should wrap the whole synchronization flow related operations within a single transaction. While that's a goal and [part of our roadmap](https://github.com/Kinto/kinto.js/issues/16), right now this isn't fully implemented.
 
-That means if anything fails during the sync flow while some records have already been processed locally, there won't be any rollback performed — and you'll have to handle the situation by hand.
+That means if anything fails during a given step of the sync flow while some records have already been processed locally already in the previous one, the database should turn into an intermediate inconsistent state.
```
rnewman: This sentence needs a little bit of English fixing. Not urgent.

docs/limitations.md
```diff
@@ -26,6 +26,7 @@ Fortunately, as in theory records don't carry any relations information, usually
 
 > *If you don't emulate relations in your data schemas, you're safe.*
```
rnewman: So… if one does so, you can end up in an inconsistent state.

This is the case with something like Sync's bookmark model. There are two relations in every record: a parent pointer and a child list.

What changes need to occur in Kinto for this limitation to be lifted? This seems like the right doc to make a note of those.

n1k0: I think Kinto isn't supporting constrained relations on purpose, mainly to keep things simple. This decision could be revisited though, we'll probably talk about this at the upcoming London ww.

PR #307, opened 2016-01-27
Fixes #305: better err message when trying to recreate a virtually deleted record

### LINKED ISSUE #305: Cannot create; delete; create a record with same id
To reproduce:

````js
var k = new Kinto();
k.collection('foo').create({ id: '7e22bc3a-4515-462b-976d-351d9285002f' },
                           { useRecordId: true }).then(() => {
  return k.collection('foo').delete('7e22bc3a-4515-462b-976d-351d9285002f');
}).then(() => {
  return k.collection('foo').create({ id: '7e22bc3a-4515-462b-976d-351d9285002f' },
                                    { useRecordId: true });
});
````

Expected: should work
Actual: ConstraintError

### PR DESCRIPTION
**Pull request by n1k0**
r=? @michielbdejong

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -314,7 +314,14 @@ export default class Collection {
     }
     return this.db.create(newRecord).then(record => {
       return {data: record, permissions: {}};
-    });
+    })
+      .catch(err => {
+        if (options.useRecordId) {
+          throw new Error(
+            "Couldn't create record. It may have been virtually deleted.");
+        }
+        throw err;
+      });
   }
 
   /**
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/collection.js [hunk 1]
```diff
@@ -314,7 +314,14 @@ export default class Collection {
     }
     return this.db.create(newRecord).then(record => {
       return {data: record, permissions: {}};
-    });
+    })
+      .catch(err => {
+        if (options.useRecordId) {
```
michielbdejong: why not check for `record._status === "deleted"` instead?

n1k0: That could be another check to add, yes; though we may want to allow this in tests, so I'm not sure - after all, should we forbid creating virtually deleted records?

michielbdejong: Ah ok, then let's do it like you did it here.

michielbdejong:
r+

PR #308, opened 2016-01-28
Add babel-cli dependency for npm run build

michielbdejong:
See http://stackoverflow.com/a/33445787/680454


n1k0:
Yeah, we fixed this in #303 which is about to land anytime soon.

n1k0:
Now #303 has landed, we can safely close this.

Issue #306, opened 2016-01-27
304 not detected properly


michielbdejong:
Even if you see a 304 in the Network tab, `res.status === 200` [here](https://github.com/Kinto/kinto.js/blob/b0edc3edd0caf33c0001968982e1b57faa7398f2/src/api.js#L196), and the changes from the last 200 response are returned.

This means that if you try to change a record that was fetched in the previous sync run, this will give a conflict, but shouldn't.

michielbdejong:
I think the thing to do would be to check the `ETag` response header instead of checking for `res.status === 304`, but that's blocked by https://github.com/mozilla-services/cliquet/issues/631

michielbdejong:
Can no longer reproduce the problem after applying https://github.com/Kinto/kinto.js/compare/master...michielbdejong:copy-lastmodified?expand=1

n1k0:
And that makes sense; if the database knows about a last_modified for this record, we should keep it of not provided.

I've already have similar patch in #286, which has strangely never landed. Maybe we should do so :)

michielbdejong:
So even though it's no longer overwriting the last_modified value, we should look at why it's trying to create conflicts with `remote: null` - I think the first problem is that it's repeating a request that returned records before, which I think should never happen, and the second problem is that it's trying to read the body of a 304 response.

michielbdejong:
This seems to be fixed now by a combination of various fixes; can't reproduce anymore.

PR #309, opened 2016-01-28
Set last_modified on updated record if missing

michielbdejong:
Here's what happens without this patch:
* Start an empty Kinto server
* Sync (collection is empty)
* Create record '12345678-1234-1234-1234-1234567890ab' as `{ value: 'a' }`, and sync (record is created remotely)
* Update record '12345678-1234-1234-1234-1234567890ab' to `{ value: 'b' }`, and sync

Expected: the record is updated remotely
Actual:
* First, data is fetched, and (due to https://github.com/Kinto/kinto.js/issues/306 I think?) record  '12345678-1234-1234-1234-1234567890ab' is updated to `{ value: 'a' }`, but that's probably unrelated here.
* Now `gatherLocalChanges` notes the update to `{ value: 'b' }` and adds it to `toSync`, but without `last_modified`.
* Since there is not `last_modified` on `updatedRecord`, the `If-None-Match: *` header is set
* This request fails, of course.
* The record is reverted to  `{ value: 'a' }`

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -422,6 +422,9 @@ export default class Collection {
           const source = options.patch ? Object.assign({}, existing, record)
                                        : record;
           const updated = markStatus(source, newStatus);
+          if (existing.last_modified && !updated.last_modified) {
+            updated.last_modified = existing.last_modified;
+          }
           transaction.update(updated);
           return {data: updated, permissions: {}};
         });
--- a/test/collection_test.js
+++ b/test/collection_test.js
@@ -461,6 +461,36 @@ describe("Collection", () => {
           _status: "updated"
         });
     });
+
+    it("should remove previous record fields", () => {
+      return articles.create(article)
+        .then(res => articles.get(res.data.id))
+        .then(res => {
+          return articles.update(
+            Object.assign({}, {id: res.data.id}, {
+              title: "new title",
+            }));
+        })
+        .then(res => res.data)
+          .should.eventually.not.have.property("url");
+    });
+
+    it("should preserve record.last_modified", () => {
+      return articles.create({
+        title: "foo",
+        url: "http://foo",
+        last_modified: 123456789012
+      })
+        .then(res => articles.get(res.data.id))
+        .then(res => {
+          return articles.update(
+            Object.assign({}, {id: res.data.id}, {
+              title: "new title",
+            }));
+        })
+        .then(res => res.data)
+          .should.eventually.have.property("last_modified").eql(123456789012);
+    });
   });
 
   /** @test {Collection#resolve} */
--- a/test/integration_test.js
+++ b/test/integration_test.js
@@ -636,6 +636,7 @@ describe("Integration tests", () => {
         let syncResult;
 
         beforeEach(() => {
+          let recordId;
           return fetch(`${TEST_KINTO_SERVER}/buckets/default/collections/tasks/records`, {
             method: "POST",
             headers: {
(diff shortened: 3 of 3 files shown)


michielbdejong:
See also https://github.com/Kinto/kinto.js/pull/286 (although that changes the behavior of `Collection#update`).

n1k0:
> See also #286 (although that changes the behavior of Collection#update).

Note that I'm now suggesting we introduce a new `#patch()` method instead of changing `#update()`'s behaviour.

michielbdejong:
Rebased. Got some interesting failures in the integration tests.

src/collection.js
```diff
@@ -407,8 +407,12 @@ export default class Collection {
     if (!this.idSchema.validate(record.id)) {
       return Promise.reject(new Error(`Invalid Id: ${record.id}`));
     }
+
     return this.get(record.id)
-      .then(_ => {
+      .then(res => {
+        if (res.data.last_modified && !record.last_modified) {
```
Natim: Good catch. I wouldn't think of this case but know that you talk about it, it seems totally a good catch :+1:

michielbdejong:
I'm having some trouble understanding the [integration tests for outgoing conflicts](https://github.com/Kinto/kinto.js/blob/master/test/integration_test.js#L633-L651). As I understand it, an outgoing conflict would only occur if the data changes after `pullChanges` but before `pushChanges`, and these tests seem to be testing something else?

michielbdejong:
Ah, I thought of a situation in which an outgoing conflict would occur:
* create a record
* sync
* delete the record with `virtual: false`, so the remote record is forgotten about
* create a new one with the same `id`
* sync (since it forgot about the remote record, it will try to create a new record)

leplatrem:
As far as I understand the setup of this test, it does the following:
- create a record
- synchronize (fetch it from server)
- update it locally as if it would never have been sync (`last_modified` undefined)

The next call to sync will generate a `PUT` request for this record with a `If-None-Match: *` header (cause it has no last_modified, it is recognized as local). Since the record already exists on the server (same id) a `412` response is returned, and an outgoing conflict is created.

michielbdejong:
OK, as the trick of adding `last_modified: undefined` from outside kinto.js no longer works with this PR, I improved the test so that it actually deletes and creates the record the proper way, and made it expext `_status: created` instead of `_status: updated`.

test/collection_test.js
```diff
@@ -446,6 +446,36 @@ describe("Collection", () => {
...
+        .then(res => articles.get(res.data.id))
+        .then(res => {
+          return articles.update(
+            Object.assign({}, {id: res.data.id}, {
+              title: "new title",
+            }));
+        })
+        .then(res => res.data)
+          .should.eventually.not.have.property("url");
+    });
```
michielbdejong: Adding this extra test is maybe more related to #310 than to this PR itself, but guess it can't hurt.

michielbdejong:
r=r? @leplatrem

test/collection_test.js
leplatrem: :+1:

src/collection.js
```diff
@@ -408,7 +408,10 @@ export default class Collection {
       return Promise.reject(new Error(`Invalid Id: ${record.id}`));
     }
     return this.get(record.id)
-      .then(_ => {
+      .then(res => {
+        if (res.data.last_modified && !record.last_modified) {
+          record.last_modified = res.data.last_modified;
+        }
```
leplatrem: I wonder if this is still valid with changes of #310.
Also shouldn't we avoid `record` mutation and create a copy instead?

michielbdejong: OK, will rebase on the other PR and use `Object.assign`.

michielbdejong:
@leplatrem Rebased. The record is copied already by `markStatus` before this new if-statement sets the `last_modified` value.

leplatrem:
r+ then!
(travis tests fail because of some weird eslint error...)

michielbdejong:
Looks like tests were failing intermittently on Travis, merging quickly before they fail again ;)

n1k0:
Huh, what was the error? npm? or something with our code?

michielbdejong:
@n1k0 first time [lint failed](https://travis-ci.org/Kinto/kinto.js/builds/105895159), second time [shutting down the server timed out](https://travis-ci.org/Kinto/kinto.js/builds/106185286), third time [was green](https://travis-ci.org/Kinto/kinto.js/builds/106191597)

n1k0:
Hmm okay, nothing too worrying then. Thanks.

PR #310, opened 2016-01-28
Allow patching a record using Collection#update().

n1k0:
WiP; refs #286, #309.

This adds support for a new `patch` option to `Collection#update()`, so one can simply augment an existing record with supplementary properties or replace the values of existing ones.

Feedback=? @michielbdejong

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -120,6 +120,7 @@ Result is:
 > #### Notes
 >
 > - An ID is required, otherwise the promise will be rejected;
+> - The `patch` option allows amending the existing record with passed data. By default this option is set to `false`, so existing records are overriden with passed data;
 > - Detailed API documentation for `Collection#update()` is available [here](https://doc.esdoc.org/github.com/Kinto/kinto.js/class/src/collection.js~Collection.html#instance-method-update).
 
 ## Deleting records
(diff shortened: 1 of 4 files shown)


michielbdejong:
Sounds useful, to avoid race conditions if you're only updating one field of a record.

n1k0:
Added docs; r=? @michielbdejong

michielbdejong:
LGTM

leplatrem:
Just one comment for the sake of the ancient art of bikeshedding: shouldn't it be `replace: true/false`? Or is *patch* clear enough in common english?

n1k0:
I'm easy both ways. I'm finding *patch* cooler, but *replace* is probably more obvious. Side note, `replace` should then be the opposite of *patch* so the default option value should be `true`.

I don't know what to do, and this is my contribution to the even more ancient art of indecision ;)

n1k0:
Actually patch can also *replace* a *property*:

```js
const existing = {id: 1, title: "foo", rank: 1};
coll.update({id: 1, rank: 2}) // {id: 1, title: "foo", rank: 2}
```

"Patch" conveys a little more the idea of merging the provided values with the initial existing ones I think. Thoughts?

n1k0:
I'm gonna land this.

michielbdejong:
Yeah, I also like `patch: true` better than `replace: false`, for the same reason.

PR #315, opened 2016-01-29
Use IndexedDB indexes on gatherLocalChanges()

leplatrem:
* [x] Take advantage of index if available for filtered field
* [x] Optimized gatherLocalChanges()
* [x] Move reduceRecords to IDB and Firefox storage (long term: replace it with generated sql)

Small benchmark: 4500 records in local store. None of them changed.

Before (1000+):
capture d ecran de 2016-01-29 14-38-28

After (<50):

capture d ecran de 2016-02-01 11-32-58

### DIFF
--- a/src/adapters/IDB.js
+++ b/src/adapters/IDB.js
@@ -1,6 +1,7 @@
 "use strict";
 
 import BaseAdapter from "./base.js";
+import { reduceRecords } from "../utils";
 
 /**
  * IndexedDB adapter.
@@ -205,12 +206,33 @@ export default class IDB extends BaseAdapter {
    * @override
    * @return {Promise}
    */
-  list() {
+  list(params={filters: {}, order: ""}) {
+    // Extract from `params.filters` the fields that are indexed.
+    const filteredFields = Object.keys(params.filters);
+    const indexedFields = filteredFields.filter(f => {
+      return ["id", "_status", "last_modified"].indexOf(f) !== -1;
+    });
+    // Since indices were created on single-columns, use the first one only.
+    const indexField = indexedFields[0];
+
     return this.open().then(() => {
       return new Promise((resolve, reject) => {
         const results = [];
         const {transaction, store} = this.prepare();
-        const request = store.openCursor();
+
+        let request;
+        if (indexField) {
+          // Filter using index.
+          const value = params.filters[indexField];
+          const range = IDBKeyRange.only(value);
+          const index = store.index(indexField);
+          request = index.openCursor(range);
+        }
+        else {
+          // Get all records.
+          request = store.openCursor();
+        }
+
         request.onsuccess = function(event) {
           const cursor = event.target.result;
           if (cursor) {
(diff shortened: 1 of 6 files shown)


src/adapters/IDB.js
```diff
@@ -205,12 +205,30 @@ export default class IDB extends BaseAdapter {
    * @override
    * @return {Promise}
    */
-  list() {
+  list(params={}) {
```
n1k0: we should provide default property values for the params object; eg `Object.keys(params.filters)` can fail when no `filters` object is provided.

leplatrem:
Ok, I just realized we were interating on the whole database even when no change was fetched from server!

Sync() is now 40 msec when no change is fetched :)

capture d ecran de 2016-02-01 11-32-58

src/collection.js
```diff
@@ -571,22 +574,23 @@ export default class Collection {
    */
   resetSyncStatus() {
     let _count;
-    // XXX filter by status
-    return this.list({}, {includeDeleted: true})
-      .then(result => {
+    return Promise.all([
+      this.list({order: "", filters: {_status: "deleted"}}, {includeDeleted: true}),
+      this.list({order: "", filters: {_status: "synced"}}, {includeDeleted: true}),
+    ])
```
n1k0: Nit: refactoring suggestion:

```js
const listStatus = (status) => this.list({
  order: "", 
  filters: {_status: status}
}, {includeDeleted: true})
return Promise.all([listStatus("deleted"), listStatus("synced")])
```

src/collection.js
```diff
@@ -604,22 +608,16 @@ export default class Collection {
...
-        }, {toDelete: [], toSync: []});
-      })
-      .then(({toDelete, toSync}) => {
-        _toDelete = toDelete;
-        return Promise.all(toSync.map(this._encodeRecord.bind(this, "remote")));
+    return Promise.all([
+      this.list({order: "", filters: {_status: "created"}}, {includeDeleted: true}),
+      this.list({order: "", filters: {_status: "updated"}}, {includeDeleted: true}),
+      this.list({order: "", filters: {_status: "deleted"}}, {includeDeleted: true})
+    ])
```
n1k0: Nit: reusing the listStatus fn suggested above would be great here

src/collection.js
n1k0: As we're using the same query type later, we could even extract the function as something reusable:

```js
function listStatus(coll, status, options={includeDeleted: true}) {
  return coll.list({order: "", filters: {_status: status}}, options);
};
```

Or if we want to pick all records matching provided statuses:

```js
function listStatuses(coll, statuses, options={includeDeleted: true}) {
  return Promise.all(statuses.map(status => {
    return coll.list({order: "", filters: {_status: status}}, options);
  })).then((...results) => [].concat.apply([], results));
};
```

leplatrem:
Ready for review/merge.

src/adapters/IDB.js
```diff
@@ -205,12 +206,33 @@ export default class IDB extends BaseAdapter {
    * @override
    * @return {Promise}
    */
-  list() {
+  list(params={filters: {}, order: ""}) {
```
n1k0: Can't we simply avoid passing `order` here?

src/collection.js
```diff
@@ -96,6 +96,13 @@ function markSynced(record) {
   return markStatus(record, "synced");
 }
 
+function listStatuses(collection, statuses, options={includeDeleted: true}) {
+  return Promise.all(statuses.map(status => {
+    return collection.list({order: "", filters: {_status: status}}, options);
```
n1k0: Now I think a little more about it, we could probably directly use the adapter api here. Will experiment.

n1k0: Done in 87690a506d9fd87c487d67dca5bd729be18db1e4

leplatrem:
r+

PR #318, opened 2016-02-02
Add support for enum filters with indexed fields in DB adapter list()

n1k0:
WiP /cc @leplatrem

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -166,20 +166,47 @@ Result is:
   data: [
     {
       id: "705b17be-e957-4c14-8f4c-86f8eaac29c0",
-      title: "foo"
+      title: "foo",
+      rank: 2,
+      _status: "created"
     },
     {
       id: "68e63131-3859-40cc-a4f7-b237ca179329",
       last_modified: 1432222889336,
-      title: "Web page"
+      title: "Web page",
+      rank: 3,
+      _status: "synced"
+    },
+    {
+      id: "86f8baac-4d12-4957-805c-8f4c17bc29c0",
+      title: "Another page",
+      rank: 1,
+      _status: "created"
     },
   ]
 }
 ```
 
+### List filtering and ordering
+
+The `#list()` method accepts an object argument allowing to define filters and ordering:
+
+```js
+articles.list({filters: {_status: "created"}, order: "rank"})
+  .then(console.log.bind(console));
+```
+
+Filters accepts an object where a key is the column name and the property value the pattern to filter the column with. For now this pattern can be either a single value or an array of values; in the latter case, results will contain all records having the filtered column value containing any of the provided ones:
+
+```js
+articles.list({filters: {_status: ["created", "updated"]}})
+  .then(console.log.bind(console));
+```
+
 > #### Notes
 >
-> - Records with `last_modified` attribute were sync'ed on a server.
+> - Records with `last_modified` attribute denote they've already been synced on a server.
+> - By default, results are ordered by `last_modified` DESC.
 > - Detailed API documentation for `Collection#list()` is available [here](https://doc.esdoc.org/github.com/Kinto/kinto.js/class/src/collection.js~Collection.html#instance-method-list).
 
 ### Filtering
--- a/src/adapters/IDB.js
+++ b/src/adapters/IDB.js
@@ -3,6 +3,104 @@
 import BaseAdapter from "./base.js";
 import { reduceRecords } from "../utils";
 
+const INDEXED_FIELDS = ["id", "_status", "last_modified"];
+
+
+/**
+ * IDB cursor handlers.
+ * @type {Object}
+ */
+const cursorHandlers = {
+  all(done) {
+    const results = [];
+    return function(event) {
+      const cursor = event.target.result;
+      if (cursor) {
+        results.push(cursor.value);
+        cursor.continue();
+      } else {
+        done(results);
+      }
+    };
+  },
+
+  in(values, done) {
+    const sortedValues = [].slice.call(values).sort();
+    const results = [];
+    return function(event) {
+      const cursor = event.target.result;
+      if (!cursor) {
+        done(results);
+        return;
+      }
+      const {key, value} = cursor;
+      let i = 0;
+      while (key > sortedValues[i]) {
+        // The cursor has passed beyond this key. Check next.
+        ++i;
+        if (i === sortedValues.length) {
+          done(results); // There is no next. Stop searching.
+          return;
+        }
+      }
+      if (key === sortedValues[i]) {
+        results.push(value);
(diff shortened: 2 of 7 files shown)


src/adapters/IDB.js
```diff
@@ -3,6 +3,80 @@
...
+  // Extract from `filters` the fields that are indexed.
+  const filteredFields = Object.keys(filters);
+  const indexedFields = filteredFields.filter(field => {
+    return INDEXED_FIELDS.indexOf(field) !== -1;
+  });
+  // Since indices were created on single-columns, use the first one only.
+  return indexedFields[0];
+}
+
+function defaultListCursor(done) {
```
leplatrem: could be called `getAll()`

leplatrem:
Small benchmark with 4500 records in local database and 1 change:

before:
capture d ecran de 2016-02-02 18-28-41


after:
capture d ecran de 2016-02-02 18-27-30

Looks like an improvement \o/

docs/api.md
```diff
@@ -166,20 +166,47 @@ Result is:
...
+      rank: 1,
+      _status: "created"
     },
   ]
 }
 ```
 
+### List filtering and ordering
+
+The `#list()` method accepts a `params` object argument allowing to define filters and ordering:
```
leplatrem: Do you think it's worth mentionning the name `params` here?

src/adapters/IDB.js
```diff
@@ -3,6 +3,105 @@
...
+
+/**
+ * Extract from filters definition the first indexed field. Since indexes were
+ * created on single-columns, extracting a single one makes sense.
+ *
+ * @param  {Object} filters The filters object.
+ * @return {String|undefined}
+ */
+function findIndexedField(filters) {
+  //
```
leplatrem: nit: empty comment

src/adapters/IDB.js
```diff
@@ -207,49 +306,28 @@ export default class IDB extends BaseAdapter {
...
-    const filteredFields = Object.keys(params.filters);
-    const indexedFields = filteredFields.filter(f => {
-      return ["id", "_status", "last_modified"].indexOf(f) !== -1;
-    });
-    // Since indices were created on single-columns, use the first one only.
-    const indexField = indexedFields[0];
-
+    const {filters} = params;
+    const indexField = findIndexedField(filters);
+    const value = filters[indexField];
```
leplatrem: Since choosing the first filtered field only is due to a «limitation» inside `createListRequest()`, I would move these lines up there and pass `params` directly. What do you think?

src/adapters/IDB.js
```diff
@@ -207,49 +306,28 @@ export default class IDB extends BaseAdapter {
...
     return this.open().then(() => {
       return new Promise((resolve, reject) => {
-        const results = [];
+        let results = [];
```
leplatrem: nit: initialization value could be undefined.

src/adapters/IDB.js
```diff
@@ -3,6 +3,105 @@
 import BaseAdapter from "./base.js";
 import { reduceRecords } from "../utils";
 
+const INDEXED_FIELDS = ["id", "_status", "last_modified"];
```
leplatrem: This is out of scope for this PR, but why do we index `last_modified`?

src/collection.js
```diff
@@ -616,11 +615,12 @@ export default class Collection {
...
-    return listStatuses(this.db, ["created", "updated", "deleted"])
-      .then(([created, updated, deleted]) => {
-        _toDelete = deleted;
+    return this.list({filters: {_status: ["created", "updated", "deleted"]}},
+                     {includeDeleted: true})
+      .then((results) => {
+        _toDelete = results.data.filter(r => r._status === "deleted");
         // Encode unsynced records.
-        const unsyncRecords = created.concat(updated);
+        const unsyncRecords = results.data.filter(r => r._status !== "deleted");
```
leplatrem: This is probably nitpicking, but I think this particular case I would prefer the version with two calls to `list()`. 

```js
return Promise.all([
   this.list({filters: {_status: ["created", "updated"]}}),
   this.list({filters: {_status: "deleted"}}),
)
.then(([unsynced, deleted]) => {
   ...
```
```

File: test/adapters/IDB_test.js [hunk 8]
```diff
@@ -248,17 +248,16 @@ describe("adapter.IDB", () => {
   describe("#list", () => {
     beforeEach(() => {
       return db.execute(transaction => {
-        transaction.create({id: 1, foo: "bar"});
-        transaction.create({id: 2, foo: "baz"});
+        for (let id=1; id<=100; id++) {
```
leplatrem: Why not 10? :)

leplatrem:
@n1k0 I was wondering how to test cleanly (stubbing list?)

```diff
diff --git a/src/collection.js b/src/collection.js
index 62d439e..64b8988 100644
--- a/src/collection.js
+++ b/src/collection.js
@@ -528,8 +528,9 @@ export default class Collection {
         if (decodedChanges.length === 0) {
           return Promise.resolve(syncResultObject);
         }
-        // XXX: list() should filter only ids in changes.
-        return this.list({order: ""}, {includeDeleted: true})
+        const remoteIds = decodedChanges.map((r) => r.id);
+        return this.list({filters: {id: remoteIds}, order: ""},
+                         {includeDeleted: true})
           .then(res => {
             return {decodedChanges, existingRecords: res.data};
           })
(END)
```

test/adapters/IDB_test.js
n1k0: Because?

src/collection.js
n1k0: True that, will update.

src/adapters/IDB.js
n1k0: Because we'll eventually order using the defined index for performance purpose.

src/adapters/IDB.js
n1k0: I think conveying the expected type is good to have.

src/adapters/IDB.js
n1k0: Good idea, will update.

n1k0:
Sorry GMail being unable to refresh a thread with latest comments live it seems, I didn't fully address your review comments. Will amend.

src/adapters/IDB.js
n1k0: Now I remember why I didn't do that; `indexField` is used in the next `.then()` callback so we can drop it from filters to be applied in-memory.

n1k0:
Ok review comments are addressed in 512a21a011d2e31d31ff5be8df6663a104c8e462

Final r=? @leplatrem

leplatrem:
r+ \o/

PR #321, opened 2016-02-04
stop transpilation of for...of in dist-fx

mozmark:


mozmark:
So it turns out we're already not transpiling classes. I need to do some work to figure out how to get a subclass-able KintoBase into a jsm - but that's a separate issue, I think.

r? @n1k0

n1k0:
r+

PR #322, opened 2016-02-04
Functional test when no change on server (ref #306)

leplatrem:
@michielbdejong I tried to reproduce #306 but can't find any problem.

I updated the kinto server to 1.11.2 which includes the fix about `ETag` in 304 responses.

Please take over the branch if you know how to highlight the problem :)

Otherwise it's not a bad idea to merge it IMO.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_install:
 - export PATH=$HOME/.local/bin:$PATH
 - pip install --user `whoami` virtualenv
 - virtualenv .env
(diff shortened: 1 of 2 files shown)


n1k0:
The test looks good. Side note, as integration tests don't use a real browser (we're requesting from node), we may miss a weird implementation behavior with http cache or something.

michielbdejong:
ok, will have a look!

leplatrem:
Bug was not reproduced, but this test does not hurt.

PR #325, opened 2016-02-05
Remove useless dependency for "for .. of" transpilation

leplatrem:
We forgot to do it in https://github.com/Kinto/kinto.js/pull/321


n1k0:
r+

Issue #341, opened 2016-02-19
import of data fails (with FirefoxStorage) when a collection has no lastModified value


mozmark:

leplatrem:
Fixed in #342

PR #342, opened 2016-02-19
fix failure in loadDump with no lastModified value

mozmark:
This is a fix for issue #341


leplatrem:
r+

This is consistent with what is done in `geLastModified()`

PR #359, opened 2016-03-08
Add Kinto master as a TravisCI target (fixes #334)

### LINKED ISSUE #334: Pin Kinto version to use in travis, add Kinto master as secondary target
This has been done for kinto-client already.

leplatrem:
@n1k0 r?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,17 +5,19 @@ node_js:
 - "4"
 python:
 - "2.7"
+env:
+- SERVER=1.11.2 ACTION=test
+- SERVER=development ACTION=test
+- ACTION="run lint"
+- ACTION="run dist"       # ensures building dist files doesn't break
+- ACTION="run build-demo" # ensures building the demo doesn't break
 before_install:
 - export PATH=$HOME/.local/bin:$PATH
 - pip install --user `whoami` virtualenv
 - virtualenv .env
-- .env/bin/pip install kinto==1.11.2
+- if [[ $SERVER && $SERVER != development ]]; then .env/bin/pip install kinto==$SERVER; fi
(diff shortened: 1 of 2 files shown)


n1k0:
Only on master... Also has 2000ms timeouts with master.

leplatrem:
> 2000ms timeouts

Failing tests or slow server startup ?

n1k0:
That's much probably slow server startup. I thought setting the higher scope global timeout to `0` (no limit) would affect children tests as well, but it's visibly not. It never fails that way on my box, so hard to investigate.

I can dump the server logs on such a failure though, which would possibly help if anything went wrong with the server itself during startup.

n1k0:
Re a8a2a8af7fe8b4b679c73e12e85df035527ab0d1, does this specifically solves the issues we had with 403, or is this just cosmetics for Kinto master compat?

leplatrem:
It solves a breaking change of kinto master (2.0) where the `default` bucket is not enabled by default.

n1k0:
> It solves a breaking change of kinto master (2.0) where the default bucket is not enabled by default.

Ok. Meaning Kinto.js should probably issue its creation when such configuration is detected, right?

leplatrem:
> Ok. Meaning Kinto.js should probably issue its creation when such configuration is detected, right?

I don't think we should automatize too much. Sometimes we'll be behind a CDN where things can't be created etc.

I'd rather document how to use the underlying `api` object to create bucket or collections.

n1k0:
> I'd rather document how to use the underlying api object to create bucket or collections.

Yeah, that would work. I'm concerned with this is slightly defeating the pure offline-first approach I wanted to have with this project, where you create buckets/collection locally and then synchronize these when back online. Though it's probably okay to require connectivity when performing these operations, I guess.

PR #363, opened 2016-03-14
Output Kinto client version in Moz build for Firefox

leplatrem:
I realized that currently there was no easy way to obtain the version of the client embedded in Firefox.

I added a `echo $npm_package_version` in the `run dist-fx`. But maybe we could generalize it to every builds?

@mozmark r?
@n1k0 feedback?

The output looks like so:
```js
/*
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
  [...]
 * limitations under the License.
 */

/*
 * This file is generated from kinto.js - do not modify directly.
 */

this.EXPORTED_SYMBOLS = ["loadKinto"];

/*
 * Version 1.2.0
 */

(function(f){if(typeof exports==="object"&&typeof module!=="undefined"){module.e
```


n1k0:
We could also `Kinto.version = require("package.json").version`, though I suspect this will embeds the whole package.json contents in dist files.

n1k0:
OK, it does. So a comment is probably good enough. Could you please take care of adding this information to other builds as well? Thanks!

leplatrem:
> adding this information to other builds as well?

Should I duplicate the ugly echo command in every target?

PR #367, opened 2016-03-17
Remove one level of object in hooks payloads (fixes #357)

### LINKED ISSUE #357: Hooks don't need a nested payload object
Right now, hooks expect a change payload object formatted that way:

```js
const payload = {changes: {lastModified: unquoted, changes: data}};
```

We could pass either two arguments (changes array and lastModified value), revisit the names or more directly remove the unnecessary level here.

feedback=? @almet

leplatrem:
@almet r?
@n1k0 feeback?

### DIFF
--- a/docs/extending.md
+++ b/docs/extending.md
@@ -88,16 +88,19 @@ db.list()
 
 ## Hooks
 
-Hooks can be called to extend the behaviour of kinto. So far it is only possible to hook when incoming changes are to be applied.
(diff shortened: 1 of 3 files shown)


n1k0:
Feedback+++

almet:
Yes! Let's land this :)

almet:
Oh, actually, I believe the docs need to be updated as well

PR #379, opened 2016-04-05
[1.2] Ensure a single instance of babel-polyfill is imported.

n1k0:
When having kinto.js as a dependency in another project using the Babel polyfill as well, we get an `only one instance of babel/polyfill is allowed` error, because that polyfill can [only be imported once](https://phabricator.babeljs.io/T1019).

This maintenance patch for the 1.2-maintenance branch fixes this, and we'll publish a 1.2.1 patch release when this is landed onto that branch.

Note that we should issue the fix for master as well, though in a next PR.

r=? @leplatrem


package.json
```diff
@@ -64,13 +64,14 @@
   "dependencies": {
     "babel": "^5.8.19",
     "btoa": "^1.1.2",
-    "fake-indexeddb": "^1.0.3",
+    "fake-indexeddb": "1.0.3",
```
n1k0: This is because latest 1.0.4 introduces a failure with one of its dependencies, which I'm about to file an issue for.

leplatrem:
r+

PR #384, opened 2016-04-07
Added hook to provide a custom kinto api class.

n1k0:
This is total WiP, PoC and DoNotLandOrYouGetFired.

This is an attempt at considering the opportunity to specify the class to use for constructing the Kinto api http client, so in Firefox we'd provide our own, which would be a nearly-pure ES6 version of the current `kinto-client` bundled version currently bundled itself in the `dist-fx` bundle.

There are certainly many other things to check and consider, so this patch is mainly an attempt at engaging a conversation around the idea.

Feedback=? @leplatrem @mozmark


fx-src/index.js
```diff
@@ -24,6 +24,10 @@ import FirefoxAdapter from "./FirefoxStorage";
 export default function loadKinto() {
   const { EventEmitter } = Cu.import("resource://devtools/shared/event-emitter.js", {});
 
+  // XXX import the standalone kinto-client module landed in FFx
+  // Lets assume this import exposes a KintoHttpClient symbol:
+  Cu.import("resource://gre/modules/kinto-http-client.jsm");
```
leplatrem: ``"resource://services-common/kinto-http-client.js"`` instead

n1k0: Yeah that was an example so you get the picture. You think this will be the final location eventually?

leplatrem:
Some context: 

* I updated `moz-kinto-client.js` in Gecko with current master ([see PR](https://github.com/leplatrem/gecko-dev/pull/8/files)), and the resulting bundle embeds new 3000 lines of transpiled code.
* This comes from the fact that we splitted the http client part into a new lib  `kinto-client.js`. When required by kinto.js and browersified to build `moz-kinto-client.js`, its transpiled version is shipped as-is.
* A related task with this issue consists in shipping `kinto-client.js` into a dedicated file so that we don't embed a transpiled version but pure ES6 code (as supported by firefox).
* Plus it will allow raw API access to performed with some dedicated code (e.g. polling `/changes`) instead of vanilla `fetch()`

leplatrem:
r+ :)

PR #385, opened 2016-04-07
Switch from assert to deeper

leplatrem:
Use deeper instead of assert, which is shipped entirely by browserify.
(as an example, it reduces the gecko build by 25Ko)

I had to blacklist `buffertools` because of this: https://github.com/othiym23/node-deeper/pull/4

@n1k0 r?

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,10 +8,10 @@
     "build-demo": "mkdir -p dist && npm run dist-prod && cp dist/kinto-$npm_package_version.min.js demo/kinto.js",
     "demo": "npm run build-demo && http-server demo",
     "dist": "mkdir -p dist && rm -f dist/*.* && npm run dist-dev && npm run dist-prod && npm run dist-noshim && npm run dist-fx",
-    "dist-dev": "browserify -s Kinto -d -e src/index.js -o dist/kinto-$npm_package_version.js -t [ babelify --sourceMapRelative . ]",
-    "dist-noshim": "browserify -s Kinto -g uglifyify --ignore isomorphic-fetch --ignore babel-polyfill -e src/index.js -o dist/kinto-$npm_package_version.noshim.js -t [ babelify --sourceMapRelative . ]",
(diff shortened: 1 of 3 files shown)


n1k0:
Test failures are worrying...

package.json
```diff
@@ -8,10 +8,10 @@
...
+    "dist-dev": "browserify -s Kinto -d --ignore buffertools -e src/index.js -o dist/kinto-$npm_package_version.js -t [ babelify --sourceMapRelative . ]",
+    "dist-noshim": "browserify -s Kinto -g uglifyify --ignore buffertools --ignore isomorphic-fetch --ignore babel-polyfill -e src/index.js -o dist/kinto-$npm_package_version.noshim.js -t [ babelify --sourceMapRelative . ]",
+    "dist-prod": "browserify -s Kinto -g uglifyify --ignore buffertools -e src/index.js -o dist/kinto-$npm_package_version.min.js -t [ babelify --sourceMapRelative . ]",
+    "dist-fx": "BABEL_ENV=firefox browserify -s loadKinto --ignore buffertools --ignore isomorphic-fetch -e fx-src/index.js -o temp.jsm -t [ babelify --sourceMapRelative . ] && mkdir -p dist && cp fx-src/jsm_prefix.js dist/moz-kinto-client.js && echo \"\n/*\n * Version $npm_package_version - $(git rev-parse --short HEAD)\n */\n\" >> dist/moz-kinto-client.js && cat temp.jsm >> dist/moz-kinto-client.js && rm temp.jsm",
```
n1k0: At some point we should investigate if browserify common options could be configured in JSON in a dedicated section of this file...

src/utils.js
```diff
@@ -1,6 +1,6 @@
 "use strict";
 
-import { deepEqual } from "assert";
+import { deepEqual } from "deeper";
```
n1k0: I think it's `import deepEqual from "deeper"` ;)

leplatrem:
> Test failures are worrying...

My fault. Contrary to assert it does not throw if false :)

package.json
leplatrem: could not find any doc about that

n1k0: that's sad... oh well

n1k0:
r+

PR #386, opened 2016-04-07
Get rid of uuid in Gecko dist file

leplatrem:
Tests pass in Gecko.

@mozmark had done something similar here: https://dxr.mozilla.org/mozilla-central/source/services/common/KintoCertificateBlocklist.js#26-40

This allows to get rid of `uuid` in the Gecko bundle.

@n1k0 r?


fx-src/index.js
```diff
@@ -15,18 +15,30 @@
...
 export default function loadKinto() {
   const { EventEmitter } = Cu.import("resource://devtools/shared/event-emitter.js", {});
+  const { generateUUID } = Cc["@mozilla.org/uuid-generator;1"].getService(Ci.nsIUUIDGenerator);
 
   Cu.import("resource://gre/modules/Timer.jsm");
   Cu.importGlobalProperties(['fetch']);
 
+  // Leverage Gecko service to generate UUIDs.
+  function makeIDSchema() {
+    const RE_UUID = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
```
n1k0: We should probably export and reuse the one defined there https://github.com/Kinto/kinto.js/blob/master/src/utils.js#L5

n1k0:
r+wc

Issue #402, opened 2016-05-02
Get rid of deeper and its dependencies to Buffer and buffertools


leplatrem:
The `deepEqual()` function that we need is a lot simpler than the one provided, since we only want to compare incoming records with simple attributes.

* [x] Write simple recursive object comparison
* [x] Remove `--ignore buffertools`
* [x] Include EventEmitter (and possible others)
* [x] Add `--bare` to dist commands (?) This could reduce the size of dist files.
* [x] Remove hack with Buffer mock in `fx-src/jsm_prefix.js`

* Related #400

n1k0:
`deeper` has been removed in #406.

n1k0:
Fixed by #406 and #407.

PR #406, opened 2016-05-02
Simple object comparison function (ref #402)

leplatrem:
Partially addresses #402.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -110,6 +110,14 @@ export function pFinally(promise, fn) {
   );
 }
 
+/**
+ * Simple deep object comparison function. This only supports comparison of
+ * serializable JavaScript objects.
+ *
+ * @param  {Object} a The source object.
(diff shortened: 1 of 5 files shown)


src/utils.js
```diff
@@ -109,3 +109,32 @@ export function pFinally(promise, fn) {
...
+ * @param  {Object} b The compared object.
+ * @return {Boolean}
+ */
+export function deepEqual(a, b) {
+  if (a === b) {
+    return true;
+  }
+  if (typeof(a) !== typeof(a)) {
+    return false;
+  }
```
glasserc: The second one should probably be `typeof(b)`

n1k0: Good catch ;)

n1k0: Fixed in https://github.com/Kinto/kinto.js/commit/10b553d609fc8fd0e203ab14669b27550f774163

PR #407, opened 2016-05-02
Exclude process module from dist files.

n1k0:
I tried as much as I could to use `--bare` along `--require events` but this is visibly not how it's supposed to work.

Meanwhile I discovered the useful `browserify --list` command option which lists all detected and bundled packages; that way I could see that the only nodejs builtins we're actively using and bundling are `process` and `events`. While we're actually using `events`, we don't care about `process`, so I've just ignored it.

I'm still concerned with the size of the non-minified build (1MB), though it comes with a bunch of polyfills old browsers may actually need. Nowadays most users should use the `noshim` build anyway, which features a nice 68KB.

r=? @leplatrem


package.json
```diff
@@ -5,12 +5,12 @@
   "main": "lib/index.js",
   "scripts": {
     "build": "babel -d lib/ src/",
-    "build-demo": "mkdir -p dist && npm run dist-prod && cp dist/kinto-$npm_package_version.min.js demo/kinto.js",
+    "build-demo": "mkdir -p dist && npm run dist-dev && cp dist/kinto-$npm_package_version.js demo/kinto.js",
```
n1k0: I'm gonna revert that.

n1k0:
Filed #408 about the intermittent test server failure.

leplatrem:
Ok, let's do that then.

The alternative is to find a minimalist *polyfill* for EventEmitter

PR #416, opened 2016-05-05
add last_modified to all tests

glasserc:
This is something I encountered when I was working on the conflict
resolution thing.

Our stubs should return the same kinds of data that an actual server would.


leplatrem:
r+ (wonder if rebase is needed though)

glasserc:
OK, rebased (thanks!)

Issue #414, opened 2016-05-04
Allow marking an individual local record as "resolved"


n1k0:
We don't currently have any API to mark a record as *resolved* after a conflict has been handled locally (typically using the `Collection#resolve()` method). It would probably make sense to schedule these for being individually force-published (`CLIENT_WINS` strategy) at next `sync()` call.

What do you think? /cc @leplatrem @Natim @almet @glasserc

glasserc:
I don't like it. This really only works in the case where we have CLIENT_WINS, otherwise we could lose data.

I was able to reproduce it using Kinto/kinto-admin#54. At time t=t0, I created a record locally and synced it. At t=t1, I changed it in the browser. At t=t2, I changed it on the Kinto server using httpie. At t=t3, I tried to do another sync and got a conflict. I resolved the conflict and retried the sync and still got the conflict (as @n1k0 reported).

I think what we're seeing is more like a bug in Kinto.js. What happens is:

- The client initiates a sync by checking for any changes since time t0, which is when the last sync happened.
- The Kinto server reports that at t2, there was a change locally.
- When I resolve the change, it correctly updates the record's last_modified to be t2.
- However, when I do another sync, it begins by asking again for any changes since time t0.

The first thing I thought of was to make it save the new last modified on every conflict resolution, but of course that won't work because it can hide/cause us to "forget" other conflicts. I think the right fix here is that when the client does a sync, it should still check for changes since t0, but when it sees that one record was changed at t2 and that our local copy was "resolved" with the t2 version (according to its last_modified), it shouldn't consider it a conflict. It should still only update the collection's last_modified when no conflicts happen (or when the only conflicts have already been resolved).

glasserc:
I tried to write the above up as a test, etc. but I broke other tests and I'm not confident that I wrote the test correctly anyhow. You can see it at https://github.com/glasserc/kinto.js/tree/conflicts.

leplatrem:
I'm thinking out loud: when do we resolve a conflict, why don't we set the sync status to «updated»? It would be equivalent to having accepted the server version and applied changes on top of it.

Or not ? :)

n1k0:
I hate you.

n1k0:
@glasserc after discussing it with @leplatrem, we think you just nailed it on the head. Writing a test for this is no easy task, we'll help you.

leplatrem:
Something like:
[...]

PR #417, opened 2016-05-05
Conflicts

glasserc:
Try to address #414 by not considering it a conflict if we pull a change that has the same `last_modified` as the remote version. This represents a conflict that we've already resolved, so it's no longer a conflict, even if its data is different. Tests included.

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -151,7 +151,7 @@ function importChange(transaction, remote) {
       // one, this represents an object that corresponds to a resolved
       // conflict. Our local version represents the final output, so
       // we keep that one. (No transaction operation to do.)
-      // But if our last_modifieds is undefined,
+      // But if our last_modified is undefined,
       // that means we've created the same object locally as one on
       // the server, which *must* be a conflict.
       return {type: "void"};
--- a/src/collection.js
+++ b/src/collection.js
@@ -150,10 +150,12 @@ function importChange(transaction, remote) {
       // If our local version has the same last_modified as the remote
       // one, this represents an object that corresponds to a resolved
       // conflict. Our local version represents the final output, so
-      // we keep that one. But if our last_modifieds is undefined,
+      // we keep that one. (no transaction operation to do).
+      // But if our last_modifieds is undefined,
       // that means we've created the same object locally as one on
       // the server, which *must* be a conflict.
-      return {type: "updated", data: markSynced(local), previous: local};
+      const updated = markStatus(local, "updated");
+      return {type: "resolved", data: updated, previous: local};
     }
     return {
       type: "conflicts",
--- a/src/collection.js
+++ b/src/collection.js
@@ -150,12 +150,11 @@ function importChange(transaction, remote) {
       // If our local version has the same last_modified as the remote
       // one, this represents an object that corresponds to a resolved
       // conflict. Our local version represents the final output, so
-      // we keep that one. (no transaction operation to do).
+      // we keep that one. (No transaction operation to do.)
       // But if our last_modifieds is undefined,
       // that means we've created the same object locally as one on
       // the server, which *must* be a conflict.
-      const updated = markStatus(local, "updated");
-      return {type: "resolved", data: updated, previous: local};
+      return {type: "void"};
     }
     return {
       type: "conflicts",
--- a/src/collection.js
+++ b/src/collection.js
@@ -146,11 +146,13 @@ function importChange(transaction, remote) {
       transaction.update(synced);
       return {type: "updated", data: synced, previous: local};
     }
-    if (local.last_modified === remote.last_modified) {
+    if (local.last_modified !== undefined && local.last_modified === remote.last_modified) {
       // If our local version has the same last_modified as the remote
       // one, this represents an object that corresponds to a resolved
       // conflict. Our local version represents the final output, so
-      // we keep that one.
+      // we keep that one. But if our last_modifieds is undefined,
+      // that means we've created the same object locally as one on
+      // the server, which *must* be a conflict.
       return {type: "updated", data: markSynced(local), previous: local};
     }
     return {
--- a/src/collection.js
+++ b/src/collection.js
@@ -146,6 +146,13 @@ function importChange(transaction, remote) {
       transaction.update(synced);
       return {type: "updated", data: synced, previous: local};
     }
+    if (local.last_modified === remote.last_modified) {
+      // If our local version has the same last_modified as the remote
+      // one, this represents an object that corresponds to a resolved
+      // conflict. Our local version represents the final output, so
+      // we keep that one.
+      return {type: "updated", data: markSynced(local), previous: local};
+    }
     return {
       type: "conflicts",
       data: {type: "incoming", local: local, remote: remote}
--- a/test/collection_test.js
+++ b/test/collection_test.js
@@ -621,8 +621,9 @@ describe("Collection", () => {
         local: local,
         remote: remote,
       };
-      const syncResult = new SyncResultObject();
-      const resolution = Object.assign({}, local, {title: "resolved"});
+      const resolution = Object.assign({}, local, {
+        title: "resolved"
+      });
       sandbox.stub(KintoClientCollection.prototype, "listRecords").returns(
         Promise.resolve({
           data: [
@@ -631,6 +632,7 @@ describe("Collection", () => {
           next: () => {},
           last_modified: "\"42\"",
         }));
+      const syncResult = new SyncResultObject();
       return articles.resolve(conflict, resolution)
         .then(() => articles.pullChanges(syncResult))
         .should.eventually.become({
@@ -639,16 +641,16 @@ describe("Collection", () => {
           errors:    [],
           created:   [],
           published: [],
-          updated:   [{
+          resolved:   [{
             id: local.id,
             last_modified: 42,
             title: "resolved",
-            _status: "synced",
+            _status: "updated",
           }],
           skipped:   [],
           deleted:   [],
           conflicts: [],
-          resolved:  [],
+          updated: []
         });
     });
   });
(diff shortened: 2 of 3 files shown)


test/collection_test.js
```diff
@@ -610,6 +610,49 @@ describe("Collection", () => {
           last_modified: remote.last_modified
         });
     });
+
+    it("should ignore resolved conflicts during sync", () => {
```
leplatrem: nitpick: since this test applies to `pullChanges()` and not `resolve()` it should probably be moved.

test/integration_test.js
```diff
@@ -648,6 +648,80 @@ describe("Integration tests", function() {
...
+        describe("Resolving conflicts doesn't interfere with sync", () => {
+          const conflictingId = uuid4();
+          const testData = {
+            localSynced: [
+              {id: conflictingId, title: "conflicting task", done: false}
+            ],
+            localUnsynced: [],
+            server: [],
+          };
+          beforeEach(() => testSync(testData));
```
leplatrem: nitpick: add blank lines

test/integration_test.js
```diff
@@ -648,6 +648,80 @@ describe("Integration tests", function() {
...
+          it("should sync over resolved records", () => {
+            return tasks.update({id: conflictingId, title: "locally changed title"},
+                         {patch: true})
```
leplatrem: nitpick: identation

test/integration_test.js
```diff
@@ -648,6 +648,80 @@ describe("Integration tests", function() {
...
+              .then(newRecord => {
+                expect(newRecord.data.last_modified).to.exist;
+                return tasks.api.bucket("default").collection("tasks").updateRecord(
```
leplatrem: Maybe add a comment like: `// Change the record remotely to introduce a conflict`

test/integration_test.js
```diff
@@ -648,6 +648,80 @@ describe("Integration tests", function() {
...
+                  {id: conflictingId, title: "remotely changed title"}, {patch: true});
+              })
+              .then(() => tasks.sync())
+              .then(syncResult => {
+                expect(syncResult.ok).eql(false);
+                expect(syncResult.conflicts).to.have.length.of(1);
+                // always pick our version, which has an older last_modified
```
leplatrem: > which has an older last_modified

Instead of this, I would note that `resolve()` is going to take the remote last_modified

test/integration_test.js
```diff
@@ -648,6 +648,80 @@ describe("Integration tests", function() {
...
+                return tasks.resolve(syncResult.conflicts[0], syncResult.conflicts[0].local);
+              })
+              .then(() => tasks.sync())
+              .then(syncResult => {
+                expect(syncResult.ok).eql(true);
+                expect(syncResult.conflicts).to.have.length.of(0);
+                expect(syncResult.updated).to.have.length.of(0);
+              })
+              .then(() => tasks.get(conflictingId))
+              .then(record => {
```
leplatrem: I think you can do `then(({data}) => { `

test/integration_test.js
```diff
@@ -648,6 +648,80 @@ describe("Integration tests", function() {
...
+            return tasks.create({id: conflictingId2, title: "second title"},
+                                {useRecordId: true})
+              .then(() => tasks.sync())
+              .then(() =>
+                rawCollection.updateRecord(
+                  {id: conflictingId, title: "remotely changed title"}, {patch: true})
+                   )
+              .then(() =>
+                rawCollection.updateRecord(
+                  {id: conflictingId2, title: "remotely changed title2"}, {patch: true}))
```
leplatrem: Maybe clearer this way ?
```js
              .then(() => rawCollection.updateRecord({id: conflictingId2, title: "remotely changed title2"},
                                                     {patch: true}))
```

test/integration_test.js
```diff
@@ -648,6 +648,80 @@ describe("Integration tests", function() {
...
+              .then(() => tasks.get(conflictingId))
+              .then(record => {
+                expect(record.data.title).eql("locally changed title");
+                expect(record.data._status).eql("synced");
+              });
+          });
+
+          it("should not skip other conflicts", () => {
+            const conflictingId2 = uuid4();
+            const rawCollection = tasks.api.bucket("default").collection("tasks");
```
leplatrem: nitpick: move that above along `describe()` to share it with both tests ?

test/collection_test.js
```diff
@@ -610,6 +610,49 @@ describe("Collection", () => {
...
+      const syncResult = new SyncResultObject();
+      return articles.resolve(conflict, resolution)
+        .then(() => articles.pullChanges(syncResult))
+        .should.eventually.become({
+          ok: true,
+          lastModified: 42,
+          errors:    [],
+          created:   [],
+          published: [],
+          resolved:   [{
```
leplatrem: Now that we have `void` type, there is nothing about this locally updated record at this staged. It is pending for publication, and will be listed in `published` when `pushChanges()` will be called.

test/integration_test.js
```diff
@@ -648,6 +648,80 @@ describe("Integration tests", function() {
...
+                expect(syncResult.ok).eql(false);
+                expect(syncResult.conflicts).to.have.length.of(1);
+                // always pick our version, which has an older last_modified
+                return tasks.resolve(syncResult.conflicts[0], syncResult.conflicts[0].local);
+              })
+              .then(() => tasks.sync())
+              .then(syncResult => {
+                expect(syncResult.ok).eql(true);
+                expect(syncResult.conflicts).to.have.length.of(0);
+                expect(syncResult.updated).to.have.length.of(0);
```
leplatrem: we should be able to expect `published` to contain the change made locally (i.e. the resolution)

test/integration_test.js
```diff
@@ -648,6 +648,80 @@ describe("Integration tests", function() {
...
+                // resolve just one conflict and ensure that the other
+                // one continues preventing the sync, even though it
+                // happened "after" the first conflict
+                return tasks.resolve(syncResult.conflicts[1], syncResult.conflicts[1].local);
+              })
+              .then(() => tasks.sync())
+              .then(syncResult => {
+                expect(syncResult.ok).eql(false);
+                expect(syncResult.conflicts).to.have.length.of(1);
+                expect(syncResult.updated).to.have.length.of(0);
```
leplatrem: we should be able to expect `published` to contain the change made locally (i.e. the resolution)

test/collection_test.js
glasserc: Good point. Done in 1978de8b4cb4dedb0ff208003343bc670ffc8cc5.

test/integration_test.js
glasserc: Yeah, sorry -- I've been working on a team with someone who was very strict about not exceeding 80 characters per line :) Fixed in 1a20231cd5117fda00a7c9b086499404127e9b9a.

test/collection_test.js
glasserc: Yeah, in fact this was breaking the test. Fixed in 71f367cb031a8f2f9e0f37ef6147a7d46f0750d0.

test/integration_test.js
glasserc: Right now it doesn't, I think because `pullChanges` "fails" with a conflict before we can `pushChanges`.

test/integration_test.js
glasserc: Good call. Fixed in ccaea7d6d247ca895dd23073ed6f6ec44a2970d2.

test/integration_test.js
glasserc: OK. Done in a818f91ab27e36385c5a6ea68799f126a52e14b2.

test/integration_test.js
glasserc: What I actually wanted was to extract `data` but call it `record`, but I didn't know what the [syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assigning_to_new_variable_names) was for it. Fixed in 177b47cd1519ea60568026d0de817bc02cbcb81b.

test/integration_test.js
glasserc: Good idea! I needed to put it in `beforeEach` because it accesses `tasks`, which is defined in a `beforeEach`. See 70d1525526cd34bc93b8d81b44150016083746c0.

test/integration_test.js
glasserc: Done in d25d1047e571a5d034773238bb10fe21f0b56348.

src/collection.js
```diff
@@ -146,6 +146,16 @@ function importChange(transaction, remote) {
       transaction.update(synced);
       return {type: "updated", data: synced, previous: local};
     }
+    if (local.last_modified !== undefined && local.last_modified === remote.last_modified) {
+      // If our local version has the same last_modified as the remote
+      // one, this represents an object that corresponds to a resolved
+      // conflict. Our local version represents the final output, so
+      // we keep that one. (No transaction operation to do.)
+      // But if our last_modifieds is undefined,
```
n1k0: nit: *last_modified*

n1k0:
From @leplatrem on #414:

> * In resolve() mark as `"synced"` if conflict is resolved with remote version. `"updated"` otherwise.

That sounds reasonable, any reason this has not been done in this patch? Would that need a followup issue?

test/integration_test.js
n1k0: We create a new SyncResultObject for each sync() operation, so we're loosing the context from a previous run anyway. I can't think of any way to retrieve this information without heavier refactoring... I suggest we open a followup issue for addressing this if we want to.

n1k0:
I couldn't test kinto-admin with this patch as it involves a bunch of operations like building lib files in a node_modules folder, which I'm not going to do :)

I'm pretty confident this patch does the job just fine regarding the passing tests.

r+ with commits squashed

leplatrem:
> any reason this has not been done in this patch? Would that need a followup issue?

Yes, let's open another issue/pr for this

PR #422, opened 2016-05-09
Do not publish resolved conflicts with remote version (fixes #419)

### LINKED ISSUE #419: resolve() should set status to "synced" or "updated"
According to @leplatrem on #417 and #414, 

>  * In resolve() mark as "synced" if conflict is resolved with remote version. "updated" otherwise.

Do this and add tests to prove it.

leplatrem:
Fixes #419


glasserc:
r+

PR #423, opened 2016-05-09
Add notion of local fields (fixes #173)

### LINKED ISSUE #173: Storing some records fields only locally?
In https://bugzilla.mozilla.org/show_bug.cgi?id=1191773#c45 we identified a need to store some metadata on records in Kinto.js which should not be synced to the remote server, and which should also be left intact if a new version of the record (without our local-only field) comes in from remote.

I'm still thinking if we can maybe find a more elegant solution for this.

leplatrem:
Fixes #173 

feedback? @michielbdejong


n1k0:
Q: What should we expose as a conflict object when localFields are defined?

n1k0:
Also, I'd naturally tend to store additional local-only data in a separate structure, to avoid having to clean the authoritative record data object in many different places; though it might imply a much larger refactoring.

More generally storing record metadata (eg. `last_modified`, `_status`) in the same object as record data has always been *to me* a questionable design decision. If you think about it, `_status` sounds like something we'd want to be local data here... I don't know what to do with that, and I don't have a clear sense of what the path forward is; I'm concerned that maintaining our current *everything-in-data* approach will be hard to scale eventually.

src/collection.js
```diff
@@ -132,7 +133,16 @@ function importChange(transaction, remote) {
     transaction.create(synced);
     return {type: "created", data: synced};
   }
-  const identical = deepEqual(cleanRecord(local), cleanRecord(remote));
+
+  // Compare remote with local record. Omit local fields and synchronization
+  // attributes (like _status and last_modified).
+  const fieldsToClean = RECORD_FIELDS_TO_CLEAN.concat(localFields);
```
leplatrem: todo: should include `schema`, `attachment`

src/collection.js
```diff
@@ -794,15 +816,14 @@ export default class Collection {
...
           toSync.forEach((r) => {
-            const isCreated = r._status === "created";
-            // Do not store status on server.
-            // XXX: cleanRecord() removes last_modified, required by safe.
-            delete r._status;
-            if (isCreated) {
-              batch.createRecord(r);
+            // Clean local fields (like _status) before sending to server.
+            // Keep last_modified required for option safe: true.
+            const published = cleanRecord(r, ["_status"].concat(this.localFields));
```
leplatrem: alternative:

```js
const published = cleanRecord(r, FIELDS_TO_CLEAN.concat(this.localFields));
published.last_modified = r.last_modified;
```

leplatrem:
I realize that we need a public API to clean the local fields of a record. 

Indeed, when we want to verify the signature of a collection, we need to strip the local fields from the records, since the server computed its signature without them.

This patch needs some adjustements.

leplatrem:
I think the current version is ready to be merged.

Especially when I see that we are about to ship code like that one: https://reviewboard.mozilla.org/r/49755/diff/1/#0
capture d ecran de 2016-05-12 13-13-22

docs/api.md
```diff
@@ -500,6 +500,27 @@ function sync() {
 
 Here we're solving encountered conflicts by picking all remote versions. After conflicts being properly addressed, we're syncing the collection again, until no conflicts occur.
 
+## Local fields
+
+By default, kinto.js sends every record attribute stored locally.
+
+In order to store some field only locally, and never publish them to the server,
+you can provide a list of field names in the `localFields` option of `Kinto#collection`:
```
n1k0: μnit: do not manually wrap this line

docs/api.md
```diff
@@ -500,6 +500,27 @@ function sync() {
...
+collection = kinto.collection("articles", {
+  localFields: ["captain", "age"]
+});
+```
+
+A [`#cleanLocalFields()`](https://doc.esdoc.org/github.com/Kinto/kinto.js/class/src/collection.js~Collection.html#instance-method-cleanLocalFields) method is available to clean a record from the local fields:
+
+```js
+stripped = collection.cleanLocalFields(record);
+
```
n1k0: μnit: unnecessary blank line

src/collection.js
```diff
@@ -3,25 +3,23 @@
...
- * @param  {Object} record        The record object.
- * @param  {Array}  excludeFields The list of keys to exclude.
- * @return {Object}               A clean copy of source record object.
+ * Compare two records omitting local fields and synchronization
+ * attributes (like _status and last_modified)
+ * @param {Object} a    A record to compare.
+ * @param {Object} b    A record to compare.
+ * @return {boolean}
  */
-export function cleanRecord(record, excludeFields=RECORD_FIELDS_TO_CLEAN) {
```
n1k0: That's a major change, with no prior notice...FWIW it's used [in kinto-admin](https://github.com/Kinto/kinto-admin/blob/master/scripts/reducers/conflicts.js#L1).

src/collection.js
```diff
@@ -900,6 +908,17 @@ export default class Collection {
   }
 
   /**
+   * Return a copy of the specified record without the local fields.
+   *
+   * @param  {Object} record  A record with potential local fields.
+   * @return {Object}
+   */
+  cleanLocalFields(record) {
```
n1k0: Wondering out loud, wouldn't we want an option to be added to `get()`, `list()` and possibly other methods so we could retrieve cleaned records? eg. `excludeLocalFields` or eq.

n1k0: Note: We may want to consider implementing this the day we actually need it.

src/utils.js
```diff
@@ -138,3 +138,19 @@ export function deepEqual(a, b) {
...
+ * @return {Object}            A copy without the specified keys.
+ */
+export function omitKeys(obj, keys=[]) {
+  return Object.keys(obj).reduce((acc, key) => {
+    if (keys.indexOf(key) === -1) {
+      acc[key] = obj[key];
+    }
+    return acc;
+  }, {});
+}
```
n1k0: μnit: missing carriage return

test/collection_test.js
```diff
@@ -32,17 +32,9 @@ describe("Collection", () => {
...
+    const opts = Object.assign({events, adapter: IDB}, options);
     api = new Api(FAKE_SERVER_URL, events);
-    return new Collection(TEST_BUCKET_NAME, TEST_COLLECTION_NAME, api, {
-      events,
-      idSchema,
-      remoteTransformers,
-      hooks,
-      adapter: IDB
-    });
+    return new Collection(TEST_BUCKET_NAME, TEST_COLLECTION_NAME, api, opts);
```
n1k0: nice :D

test/collection_test.js
```diff
@@ -580,6 +578,23 @@ describe("Collection", () => {
     });
   });
 
+  /** @test {Collection#cleanLocalFields} */
+  describe("#cleanLocalFields", () => {
+    it("should remove the local fields", () => {
+      const collection = testCollection();
+      const record = {id: "1", _status: "synced", last_modified: 42};
+      const cleaned = collection.cleanLocalFields(record);
+      expect(cleaned).eql({id: "1", last_modified: 42});
```
n1k0: μnit: missing blank line before assertion

test/utils_test.js
```diff
@@ -273,4 +274,15 @@ describe("Utils", () => {
       expect(deepEqual({a: [{b: 1}, {c: 1}]}, {a: [{c: 1}, {b: 1}]})).eql(false);
     });
   });
+
+  describe("omitKeys", () => {
+    it("should return same object if no key is passed", () => {
+      const input = {a: 1, b: 2};
+      expect(omitKeys(input)).eql(input);
```
n1k0: μnit: `expect(omitKeys({a: 1, b: 2})).eql(input);` looks okay to me

src/collection.js
leplatrem: `cleanRecord()` was not documented, thus not officially exposed :]

I added some documentation about this change as part of the breaking changes for 3.X.

For [what I could see in the code](https://github.com/Kinto/kinto-admin/blob/42469960835bacabd7d904ee52015fd16c4a8109/scripts/reducers/form.js#L9) of kinto-admin, the function `omitKeys()` could be used instead (I would even suggest to copy its code instead of importing it from kinto.js)

n1k0: Yeah sure, I wasn't too concerned with backporting the feature ;)

n1k0:
r+

michielbdejong:
Sorry for missing this discussion, hadn't gotten round yet to my github notifs after pto last week.

Just a remark in reaction to @n1k0's remark about refactoring to make the distinction between payload and metadata clearer - another option would be to have top level fields `payload`, `local` and `meta` inside the IndexedDB record, so all record fields are nicely organized by function. That way, a developer doesn't have to worry about fieldnames starting with `_` being reserved (the api for listing which fields are local would stay the same as introduced in this PR).

It would still be quite a refactor though, and would require an IDB version update, as well as a migration process for sqllite, so not very attractive to make that change at this point, maybe.

PR #424, opened 2016-05-09
Avoid redownloading our own changes (fixes #144)

### LINKED ISSUE #144: Avoid redownloading changes on Sync

leplatrem:
Fixes #144 

feedback? @Natim

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -727,10 +727,20 @@ export default class Collection {
       lastModified: this.lastModified,
       headers: {},
     }, options);
+
+    // Optionally ignore records when pulling for changes.
+    // (avoid redownloading our own changes on last step of #sync())
+    let filters;
+    if (options.exclude) {
+      // Take the first 50 records to limit size of querystring (~2000 chars).
+      const exclude_id = options.exclude.slice(0, 50).map((r) => r.id).join(",");
+      filters = {exclude_id};
+    }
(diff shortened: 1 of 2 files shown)


src/collection.js
```diff
@@ -727,10 +727,20 @@ export default class Collection {
       lastModified: this.lastModified,
       headers: {},
     }, options);
+
+    // Optionally ignore records when pulling for changes.
+    // (avoid redownloading our own changes on last step of #sync())
+    let filters;
+    if (options.exclude) {
+      // Take the first 50 records to limit size of querystring (~2000 chars).
```
n1k0: Out of curiosity what is the actual length limitation of a query string?

leplatrem: https://github.com/Kinto/kinto.js/issues/144#issuecomment-197787716

Natim:
Seems good to me :)

leplatrem:
Thanks! Merging!

PR #434, opened 2016-05-26
Use IndexedDB by default in collection (fixes #345)

### LINKED ISSUE #345: Creating a collection without adapter throws a `No adapter provided` exception
```es6
import Collection, from "../src/collection";

new Collection(TEST_BUCKET_NAME, TEST_COLLECTION_NAME, api, {hooks});
```
throws a  `No adapter provided` exception

leplatrem:
@almet r?

### DIFF
--- a/fx-src/index.js
+++ b/fx-src/index.js
@@ -56,7 +56,8 @@ export default function loadKinto() {
 
       const defaults = {
         events: emitter,
-        ApiClass: KintoHttpClient
+        ApiClass: KintoHttpClient,
+        adapter: FirefoxAdapter
       };
 
       const expandedOptions = Object.assign(defaults, options);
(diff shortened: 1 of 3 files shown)


leplatrem:
Hum I realize that this should also|only apply to `new Kinto()`

n1k0:
I don't get the use case we're trying to address here; care to elaborate?

leplatrem:
@n1k0 see original issue by @almet : he instantiates a collection object https://github.com/Kinto/kinto.js/issues/345#issue-135728325

(We can also close this PR and the issue if irrelevant)

n1k0:
> @n1k0 see original issue by @almet : he instantiates a collection object #345 (comment)

Ah, I never really expected people to instantiate Collection directly, but as it's part of the documented public API, it probably makes sense.

r+

Issue #446, opened 2016-06-01
Which methods are re-entrant?


glasserc:
It seems like Collection#update and Collection#delete aren't re-entrant, because their access to the database isn't protected within a single transaction. (They do a `get().then(...)`.) Is this correct? Should/can we fix this?

Natim:
To make sure there is a transaction you need to make sure that the ``last_modified`` value from the get is used for the update/delete with in the `If-Match` header.

leplatrem:
> To make sure there is a transaction you need to make sure that the last_modified value from the get is used for the update/delete with in the If-Match header.

This would be for the interaction with the HTTP API. 

In this case, this is about transactions in indexeddb.

> Is this correct? Should/can we fix this?

Yes, you're right, we should fix this. I'll have a look to see if we can.

Natim:
> This would be for the interaction with the HTTP API.
> In this case, this is about transactions in indexeddb.

Oh yes you are right **kinto.js**, sorry for the misunderstanding.

leplatrem:
So... as we discussed in #450, this is not an easy task.

Mainly because of a very strong limitation of IndexedDB which cannot read within a transaction (read is async, and transaction must remain sync).

Since we need to preserve the `last_modified` field - which allows us to publish safely on the server - we cannot virtually delete nor update a record without reading what we have in the store.

One possible solution would be to store the metadata in a separate store for example, so that it remains untouched when updating the records.

glasserc:
That would make it easier to implement atomic `update()` and `delete()` and shift a lot of complexity onto `sync()`. That might be OK; `sync()` is already not re-entrant, and probably can't be because it has to query its database as well as update a bunch of records in its database.

What would we do with the `_status` field? That's the other field that `update()` touches. In my ideal world, we would put a `_status` of `created` if the record didn't already exist, and `updated` if it did. (The current `update()` also tries to preserve a status of `synced` if no real changes are made.) Since its new status depends on its old status, we have to read it in order to update it. I think if we already knew what it was, we could store it to both using one transaction. I think we'll have to change how we store this if we want to behave correctly.

Issue #445, opened 2016-06-01
Collection#update should have a variant that doesn't fail if record doesn't exist


glasserc:
```javascript
/**
 * Sets a record unconditionally.
 * This is like create() if the record doesn't exist or is deleted, or update() if the record already exists.
 */
set(record) {

}
```

This would simplify the use of Kinto.js as a key-value store.

glasserc:
Actually, having looked at this for a while, I'm pretty sure this is the same as the `#update` operation. Closing.

glasserc:
This is now called `put()` and was implemented as part of #455.

Issue #444, opened 2016-05-31
Collection#delete should have a variant that doesn't throw if the object doesn't exist


glasserc:
It seems that at present [Collection#delete](https://github.com/Kinto/kinto.js/blob/master/src/collection.js#L553) throws an error if the record doesn't exist. Some applications just want to delete unconditionally. As with #443, the only way to handle this safely is to check for a certain substring.

glasserc:
It would also be awesome if, in this case, the `#delete` operation returned true/false to indicate whether the object had been deleted.

glasserc:
This was implemented as `deleteAny` in #455.

Issue #443, opened 2016-05-31
Collection#get should have a variant that doesn't throw when records don't exist


glasserc:
It seems that at present, `Collection#get`, when given the ID of a record that doesn't exist, just throws a specific `Error`. This is hard to test for. As a result, if you want to check to see if a record is in the database, you have to call `get()` with a `catch` that examines the string message to see if it contains "not found". It would be nice if there was a clearer way to try to `get()` in a safe (non-throwing) way.

n1k0:
> This is hard to test for.

You could alternatively a little reusable helper which would be much easier to use in your project:

```js
function getRecord(id) {
  return coll.get(id).catch(err => {
    if (/not found/.test(err.message)) {
      return;
    }
    throw err;
  });
}

getRecord("non-existent")
  .then(res => [=> undefined])
```

glasserc:
Yes, this is actually what the existing code for https://bugzilla.mozilla.org/show_bug.cgi?id=1253740 does. That's what I meant when I said that it was "hard to test for". I think running a regular expression to check for a certain kind of exception is incredibly ugly. At the very least, I'd want a separate kind of exception, although with inheritance being the way it is in JS, that might not be advisable.

n1k0:
Yeah we're in js userland so typed exceptions is no easy business. It's possible to *subclass* `Error` but in my own experience itbrings its share of nightmares. Though if the regexp thing is your only gripe, would exposing a `code` property to the error help you feeling better? eg. `if (err.type === NOT_FOUND))` or eq.

Yeah that would basically be monkey patching the error object instance, so you might find this incredibly ugly too. So I don't know, maybe I'm too much used to this crap.

glasserc:
This got implemented as `getRaw()` in #455.

PR #447, opened 2016-06-01
Richer methods

glasserc:
This is an attempt to address #443, #444, and #445 by adding parameters to the `#update`, `#get`, and `#delete` operations that don't fail according to the absence of the records. This makes it easier to write key/value sorts of operations using kinto.js don't have race conditions (although #446 still seems like a problem).

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -518,7 +518,7 @@ export default class Collection {
         const updated = markStatus(source, newStatus);
         return this.db.execute((transaction) => {
           transaction.update(updated);
-          return {data: updated, permissions: {}};
+          return {data: updated, old: existing, permissions: {}};
         });
       });
   }
--- a/src/collection.js
+++ b/src/collection.js
@@ -571,9 +571,9 @@ export default class Collection {
                          includeMissing: options.unconditional})
       .then(res => {
         const existing = res.data;
-        const ret = {data: {id: id}, permissions: {}};
         if (!existing) {
-          return Promise.resolve(ret);
+          return Promise.resolve({data: {id: id}, deleted: false,
+                                  permissions: {}});
         }
         return this.db.execute((transaction) => {
           // Virtual updates status.
@@ -583,7 +583,7 @@ export default class Collection {
             // Delete for real.
             transaction.delete(id);
           }
-          return ret;
+          return {data: {id: id}, deleted: true, permissions: {}};
         });
       });
   }
--- a/src/collection.js
+++ b/src/collection.js
@@ -482,12 +482,14 @@ export default class Collection {
    * - {Boolean} synced: Sets record status to "synced" (default: false)
    * - {Boolean} patch:  Extends the existing record instead of overwriting it
    *   (default: false)
+   * - {Boolean} unconditional: If false, error when no existing
+   *   record (default: false)
    *
    * @param  {Object} record
    * @param  {Object} options
    * @return {Promise}
    */
-  update(record, options={synced: false, patch: false}) {
+  update(record, options={synced: false, patch: false, unconditional: false}) {
     if (typeof(record) !== "object") {
       return Promise.reject(new Error("Record is not an object."));
     }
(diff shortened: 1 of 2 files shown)


Natim:
The word unconditional doesn't make sense to me :(
Can we try to find another word?

Is it the same feature that we have in kinto.py with ``if_not_exists=True``?

leplatrem:
Maybe `silent` (default: `false`) everywhere?

glasserc:
`Collection#update` (as it stands before this PR) is like a `PATCH` method -- it only modifies records if they already exist. So `if_exists` is really what it does already.

I want to implement a `PUT` method, which replaces whatever was there (if anything). `if_exists` / `if_not_exists` are different because when I do a `PUT`, I don't care whether something is there or not. `silent` is wrong because I don't care how errors are reported (they should be reported normally), but instead I want to change what is considered an error.

There's an additional wrinkle here which is that I want each option to be something whose natural default is falsy, because the way kwargs works in JS is that you pass an object with a bunch of properties, and any property you don't pass doesn't get its default value (but instead is `undefined`, as unspecified properties normally are in JS).

glasserc:
Just to clarify, this is to implement chrome.storage.sync `get()`, `set()`, and `remove()` from https://developer.chrome.com/extensions/storage.

n1k0:
Disclaimer: this is a highly theoretical comment from me, I may not have all the storage.sync context in mind, so bear with me :)   

I'm a little reluctant introducing more options to already cluttered methods; I've suggested a naive approach there for having tiny focused helpers https://github.com/Kinto/kinto.js/issues/443#issuecomment-223336016

The problem with having many options for a given method is you then have to test all the their combinations to be 100% safe, and that's usually a pain (to write, to document and to maintain). Delegating the creation of dedicated helpers to developers prevents this to happen, but obviously makes the API a little less straightforward to use (though think for a minute of a developer reading the options list and trying to figure out the possible entanglements...)

n1k0:
Note: We could also decide to make this unconditional by default and bump a major. I need to think about it.

glasserc:
I agree that the current method implementations are already kind of complicated, and I made them more complicated. Adding tests was challenging for exactly the reason you mentioned -- I had to check that deleting works with and without virtual as well as with and without the new option. But writing helpers doesn't reduce complexity, it just moves it somewhere else -- instead of testing combinations of options, you have to test combinations of helpers. In fact, the situation is even worse at the moment because Kinto.js doesn't really offer the tools to write this kind of helper -- there's no clean way to catch only certain kinds of exceptions, or to say "These operations need to be atomic" (instead you run the risk that your `remove()` can cause your `update()` to crash).

Right now, the kinto.js methods offer a "Pythonic" sort of data access model -- if you do a `get()` for something that isn't there, you get an exception. If you do an `update()` for something that isn't there, exception. If you do a `create()` for something that *is* there, exception. I think that's a valid model and useful for a lot of applications (and I'd definitely normally prefer it to a JavaScripty "you just get undefined" in most cases). It's just that in my current application, I want something less application-centric and more raw.

Writing a DSL for data access that covers all the right cases is complicated. I understand if it's outside the scope of Kinto.js. How would you feel about a different set of methods `deleteAny()` and `put()`? I thought about writing these methods first but I saw so much overlap with the existing methods that it just seemed easier to fold them in. If necessary, I could probably write these methods based on accessing the `db` property of collection, but I'm concerned that this means forking `kinto.js` for this one use case.

glasserc:
Another possibility is for me to just implement a complete key-value store in memory using a `Map` or something, and additionally store everything to Kinto for the purpose of synchronization.

n1k0:
> Another possibility is for me to just implement a complete key-value store in memory using a Map or something, and additionally store everything to Kinto for the purpose of synchronization.

I'm confused, in-memory store would mean if you close the browser and reopen it you'd lose everything right? Or would you want to catch ther "browser is closing" event to perform a sync() call to the server? :)

I suspect you'll always need a persistence layer of some sort in the browser. Now for WebExtensions that could be Sqlite.jsm in the browser, but you'd basically have to rewrite a bunch of the synchronization logic. Maybe that's the way to go idk, but that sounds *slightly* overkill to me.

n1k0:
Now back to more constructive stuff ;)

> How would you feel about a different set of methods deleteAny() and put()?

I actually like the idea. It sounds easy to document, to use, to test and to maintain.

glasserc:
This got superseded by #455.

PR #450, opened 2016-06-02
Atomic update and delete (fixes #446)

### LINKED ISSUE #446: Which methods are re-entrant?

leplatrem:
This is how the patch for #446 would look like.

I personally find it a lot cleaner like that, apart from the fact that it prevents race conditions.

Most tests pass, not all. It obviously conflicts with @glasserc in #447 so we should decide in what order we merge ;)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,8 @@ node_js:
 python:
 - "3.5"
 env:
-- SERVER=2.1.0 ACTION=test
+- SERVER=3.1.1 ACTION=test
+- SERVER=2.1.2 ACTION=test
 - SERVER=development ACTION=test
 - ACTION="run lint"
 - ACTION="run dist"       # ensures building dist files doesn't break
--- a/src/collection.js
+++ b/src/collection.js
@@ -497,25 +497,26 @@ export default class Collection {
     if (!this.idSchema.validate(record.id)) {
       return Promise.reject(new Error(`Invalid Id: ${record.id}`));
     }
-    return this.get(record.id)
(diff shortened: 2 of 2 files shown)


leplatrem:
Hum, I was investigating and actually we would have to keep the re-entrant behaviour read+write for `update()` with `patch: true`.

Indeed, indexeddb is unable to read within a transaction (`store.get()` cannot be synchronous, and transactions can only happen in a single tick, ie. synchronously). Hence, we can obtain the existing record to patch it within a single transaction.

n1k0:
> Hence, we cannot obtain the existing record to patch it within a single transaction.

Didn't look at the patch but I thought it was the purpose of preloading?

leplatrem:
> also even simple update tries to "bring over" the last_modified time of any existing record

Yes you're right! And also virtual delete, since it needs the `last_modified` to delete safely.

> last_modified I don't know what to do about...

If you get rid of `last_modified`, the sync() algorithm will process the modified records as creations. If a record already exists on the server, it can only be overwritten with `CLIENT_WINS` (or manual resolve).

An alternative way would be to store the metadata in a separate table (as already suggested by Michiel here https://github.com/Kinto/kinto.js/pull/423#issuecomment-220243479)

Let's continue the discussion in the issue, I'll close this PR which lead nowhere

PR #453, opened 2016-06-03
Factual delete does not raise error if unknown (ref #446)

leplatrem:
That's the only part of #450 I could save. And even that, since it makes the behaviour of factual delete inconsistent with other methods, I'm not sure we should keep it.

Feel free to close :)

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -555,20 +555,19 @@ export default class Collection {
       return Promise.reject(new Error(`Invalid Id: ${id}`));
     }
     // Ensure the record actually exists.
-    return this.get(id, {includeDeleted: true})
-      .then(res => {
-        const existing = res.data;
-        return this.db.execute((transaction) => {
-          // Virtual updates status.
-          if (options.virtual) {
-            transaction.update(markDeleted(existing));
-          } else {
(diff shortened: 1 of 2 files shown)


src/collection.js
```diff
@@ -555,20 +555,19 @@ export default class Collection {
...
-          } else {
-            // Delete for real.
-            transaction.delete(id);
-          }
-          return {data: {id: id}, permissions: {}};
-        });
-      });
+    const getExisting = options.virtual ? this.get(id, {includeDeleted: true})
+                                        : Promise.resolve({data: {id}});
+    return getExisting
```
n1k0: Nit:

```js
const {virtual} = options;
const includeDeleted = virtual;
return Promise.resolve(virtual ? {data: {id}} : this.get(id, {includeDeleted}))
  .then(...)
```

n1k0: PS: brrr.

leplatrem:
This introduces an inconsistency in the API, while fixing the race condition for a very particular situation (other race conditions still exist).

Closing, but feel free to reopen if this becomes a concrete problem in the real world.

PR #455, opened 2016-06-03
Richer methods 2

glasserc:
This is another approach to solve #443, #444, and #445. Unlike #447, which adds options to existing methods, this PR introduces new methods which do not support the old use cases at all but simply support what I need -- raw access to a key-value store without application-level access. These methods are called `getRaw()`, `put()`, and `deleteAny()`. I re-use code from the existing `get()` and `update()` methods by extracting them (`get()` now just calls `getRaw()` and then throws if it's missing; `update()` calls `_updateRaw()`). This seemed better than copying and pasting the entire methods, although I'm open to doing that if the auxiliary methods seem too complicated. I decided not to refactor `delete()`, since very little code is shared.

r? @n1k0

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -61,6 +61,7 @@ Result is:
 >
 > - By default, records identifiers are generated locally using [UUID v4](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_.28random.29);
 but you can also define a [custom ID schema](#id-schemas));
+> - Trying to create a record with the ID of a record that already exists is an error;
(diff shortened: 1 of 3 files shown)


glasserc:
Also @Natim in case he would like to see different names for the methods.

Natim:
I don't have a strong opinion on naming this methods what you did seems fine to me.

n1k0:
Code looks good, names make sense: great work! It's probably worth mentioning these new public methods in the docs.

glasserc:
Good idea! Added in 20639fa972b038e1d0ff388d4287aa5d93008bc0.

n1k0:
r+!

PR #460, opened 2016-06-08
Removing reentrancy (fixes #446)

### LINKED ISSUE #446: Which methods are re-entrant?

leplatrem:
* [x] Preload records within the same transaction in IDB adapter
* [x] Remove re-entrancy in update/delete
* [x] **TODO**: Preload within transaction in Fx storage adapter

### DIFF
--- a/fx-src/FirefoxStorage.js
+++ b/fx-src/FirefoxStorage.js
@@ -77,6 +77,12 @@ const statements = {
       FROM collection_data
         WHERE collection_name = :collection_name;`,
 
+  "listRecordsById": `
+    SELECT record_id, record
+      FROM collection_data
+        WHERE collection_name = :collection_name
+          AND record_id IN (:record_ids);`,
+
   "importData": `
     REPLACE INTO collection_data (collection_name, record_id, record)
       VALUES (:collection_name, :record_id, :record);`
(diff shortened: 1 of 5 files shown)


glasserc:
This is cool! Thanks for digging into this so much!

We're also going to have to update `put` and `deleteAny`, which I added.

test/collection_test.js
```diff
@@ -1760,17 +1760,15 @@ describe("Collection", () => {
     it("should only retrieve the changed record", () => {
       const id1 = uuid4();
       const id2 = uuid4();
-      const list = sandbox.stub(articles, "list").returns(Promise.resolve());
+      const list = sandbox.stub(articles.db, "execute").returns(Promise.resolve([]));
```
glasserc: This should probably be `const execute`

glasserc:
Just to note that this supersedes #450 and attempts to fix #446. This is more complicated than we want because of some interactions between IndexedDB and Firefox Promises. See http://stackoverflow.com/a/28388805/330911 and http://stackoverflow.com/a/10405196. Specifically, it seems like IndexedDB transactions end when they call a callback and that callback doesn't make any more requests. Instead, we were using promises, which (although equivalent to callbacks) don't appear to "fire" before the callback ends. See also https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/ and also its link to https://bugzilla.mozilla.org/show_bug.cgi?id=1193394, which might be implicated.

leplatrem:
Current status: one test in Gecko fails : the one that creates record in parallel
https://dxr.mozilla.org/mozilla-central/rev/ddb6b30149221f00eb5dd180530e9033093d4c2b/services/common/tests/unit/test_kinto.js#74

leplatrem:
@glasserc r?

fx-src/FirefoxStorage.js
```diff
@@ -91,6 +97,7 @@ export default class FirefoxAdapter extends BaseAdapter {
   constructor(collection) {
     super();
     this.collection = collection;
+    this._connection = null;
```
glasserc: Do we need this?

fx-src/FirefoxStorage.js
```diff
@@ -149,24 +156,32 @@ export default class FirefoxAdapter extends BaseAdapter {
...
-      return Promise.reject(e);
-    }
     const conn = this._connection;
+    const collection = this.collection;
+
     return conn.executeTransaction(function* doExecuteTransaction() {
+      // Preload specified records from DB, within transaction.
+      const parameters = {
+        collection_name: collection,
+        record_ids: options.preload.map(r => r.id).join("','")
```
glasserc: This seems like a hack.. can we at least flag this with a FIXME?

src/adapters/IDB.js
```diff
@@ -253,15 +253,31 @@ export default class IDB extends BaseAdapter {
    * @return {Promise}
    */
   execute(callback, options={preload: []}) {
-    const preloaded = options.preload.reduce((acc, record) => {
-      acc[record.id] = record;
-      return acc;
-    }, {});
+    // Transactions in IndexedDB are autocommited when a callback does not
+    // perform any additionnal operation.
```
glasserc: additional

glasserc:
Still need `put()` and `deleteAny()` to be converted.

fx-src/FirefoxStorage.js
leplatrem: I think it doesn't harm, and makes it explicit we expect something to be stored here

leplatrem:
@glasserc final r+ ?

fx-src/FirefoxStorage.js
glasserc: Ah, I understand, sorry.

glasserc:
r+! :100:

PR #461, opened 2016-06-08
Reject when deleting already deleted record (fixes #458)

### LINKED ISSUE #458: Collection#delete behaves differently if record doesn't exist or if it's virtually deleted
If you delete a record that has already been virtually deleted, but not yet synced, you get a successful result. However, once you sync, that record will be actually deleted, and if you call delete again, you'll get an error. I think we should treat these two cases identically, probably by throwing an exception if the record was already deleted virtually.

leplatrem:
It conflicts with #460 but rebase/merge is trivial.

@glasserc r?

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -623,7 +623,7 @@ export default class Collection {
     return this.get(id, {includeDeleted: true})
       .then(res => {
         const existing = res.data;
-        if (existing._status === "deleted") {
+        if (options.virtual && existing._status === "deleted") {
           throw new Error(`Record with id=${id} not found.`);
         }
         return this.db.execute((transaction) => {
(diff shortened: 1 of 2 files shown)


src/collection.js
```diff
@@ -623,6 +623,9 @@ export default class Collection {
     return this.get(id, {includeDeleted: true})
       .then(res => {
         const existing = res.data;
+        if (existing._status === "deleted") {
+          throw new Error(`Record with id=${id} not found.`);
+        }
```
glasserc: I think we could just get rid of the `includeDeleted` option in the call and maybe even remove it from `get`.

glasserc:
Ah yes, we can't completely get rid of `includeDeleted` because if it's a factual delete, we don't want to error out if the record is virtually deleted. Good catch. I'm r+ on this patch.

PR #462, opened 2016-06-08
Specify remote bucket/collection in sync (fixes #451)

### LINKED ISSUE #451: Collection#sync should let you specify a different remote collection
Per https://bugzilla.mozilla.org/show_bug.cgi?id=1276935, we want to avoid leaking information by syncing to a collection whose name is too transparent, but we don't always have access to what the name "should" be. It would be nice if we could synchronize to a remote collection whose name was not the same as the local collection. Perhaps this could be specified as an argument to `Collection#sync`.

leplatrem:
@glasserc r?

### DIFF
--- a/test/collection_test.js
+++ b/test/collection_test.js
@@ -131,21 +131,6 @@ describe("Collection", () => {
       expect(collection.db).to.be.an.instanceOf(MyAdapter);
     });
 
-    it("should select the KintoClient configured bucket and collection", () => {
-      const collection = new Collection(TEST_BUCKET_NAME, TEST_COLLECTION_NAME, api, {
-        adapter: BaseAdapter
-      });
-
-      expect(collection._apiCollection)
-        .to.have.property("options")
-        .to.have.property("bucket")
-        .eql(TEST_BUCKET_NAME);
-
-      expect(collection._apiCollection)
-        .to.have.property("name")
-        .eql(TEST_COLLECTION_NAME);
-    });
-
     describe("transformers registration", () => {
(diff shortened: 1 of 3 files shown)


glasserc:
The tests seem to be failing, although I can't figure out why, but it does seem related to this change.

Can we just create one reference to the local collection and pass that around to `pushChanges` and `pullChanges` instead of giving them `collection` and `bucket` options? I think that would be cleaner.

test/collection_test.js
```diff
@@ -2200,14 +2185,14 @@ describe("Collection", () => {
       it("should transfer the headers option", () => {
         return articles.sync({headers: {Foo: "Bar"}})
           .then(() => {
-            expect(pullChanges.firstCall.args[1]).eql({headers: {Foo: "Bar"}});
+            expect(pullChanges.firstCall.args[1]).to.have.property("headers").eql({Foo: "Bar"});
```
glasserc: Why are these changes necessary? (Although they look OK to me)

leplatrem: If the options have extra entries (defaut values or whatever unreleated to the test description here) they have to be mentionned. With this change we test only a subpart of the specified options, the ones concerned by the test (ie. headers here)

glasserc: Ah, I understand, it's because we're adding options to the `pullChanges` calls to track the collection we're syncing against.

glasserc:
OMG, I just noticed you updated this a while ago and I haven't rereviewed it. Looks good now. Merging.

PR #463, opened 2016-06-08
Keep created status on update (fixes #456)

### LINKED ISSUE #456: what is the true meaning of "created" and "updated"?
I thought created meant "we have it locally, but it doesn't exist on the remote server", and "updated" means "this record exists remotely, but we have changed it since then". However, while writing code for #455, I noticed that if you call `update()` on a `created` record, it becomes `updated`. This seems strange to me, since it doesn't exist on the remote server yet. I would expect it to remain "created" until it synced.

leplatrem:


src/collection.js
```diff
@@ -515,19 +515,17 @@ export default class Collection {
...
     const keepSynced = isIdentical && oldRecord._status == "synced";
-    let newStatus = (keepSynced || synced) ? "synced" : "updated";
-    if (!oldRecord) {
-      newStatus = "created";
-    }
-    updated = markStatus(newRecord, newStatus);
+    const neverSynced = oldRecord && oldRecord._status == "created";
+    const newStatus = (keepSynced || synced) ? "synced"
+                                             : neverSynced ? "created" : "updated";
+    const updated = markStatus(newRecord, newStatus);
```
glasserc: I personally find this logic kind of ugly and I think we could actually simplify `_updateRaw` a lot more than this. Rather than adding all this code to try to preserve the `oldRecord`'s status, let's just identify the cases where we want to *change* the `newRecord`'s status. This is OK so long as the user doesn't do anything crazy with the status, like update a record and set it to `updated` or something. (If they do, the existing code will try to cover it up; my proposal would be to let it go and get mysterious errors later.)

First, `newRecord` should always have a status -- if it doesn't, we should copy the old status; if there was no old status, newRecord should be "created" and we're done.

As far as I can tell, the only cases where we need to change `newRecord`'s status are:

1. when the old record's status was synced but the new record has changed, `newRecord` should be set to "updated"
2. when we get a `synced` option, in which case we should force the status of the new record to be "synced"

I think this will be a lot cleaner than nested ternary operators.

glasserc: Note that if #456 *isn't* a bug, we have to add a third case, which is that if the old record was `created`, the new record should be `updated`, which is why I didn't implement this myself.

glasserc: Also, I guess I just thought of another situation where we want to change it, and that's if we're updating a "deleted" record -- it should become "updated" instead.

leplatrem:
I reworked it, and also improved the conflict automatic resolution which now occurs in a single transaction. Compared to the original implementation (and with re-entrancy), I find this a lot cleaner now!

@glasserc r?

glasserc:
It looks like you meant to rebase this branch and haven't yet :)

leplatrem:
> It looks like you meant to rebase this branch and haven't yet :)

I did :)

src/collection.js
```diff
@@ -525,12 +525,13 @@ export default class Collection {
...
+    // If status is deleted, mark as updated.
     const isIdentical = oldRecord && recordsEqual(oldRecord, updated, this.localFields);
     const keepSynced = isIdentical && oldRecord._status == "synced";
-    let newStatus = (keepSynced || synced) ? "synced" : "updated";
-    if (!oldRecord) {
-      newStatus = "created";
-    }
+    const neverSynced = !oldRecord || (oldRecord && oldRecord._status == "created");
+    const newStatus = (keepSynced || synced) ? "synced"
+                                             : neverSynced ? "created" : "updated";
```
glasserc: I still feel like this is too complicated or ugly to live. (Ternary operators inside ternary operators!!) How about this?

```javascript
let oldStatus = oldRecord && oldRecord._status;
let newStatus;
if (oldStatus == "synced" && !isIdentical) {
    // If the old record was "synced" but we've changed it, the new record should be "updated", 
    // even if the user didn't update the status themselves
    newStatus = "updated";
} else if (oldStatus == "deleted") {
    // If the old record was "deleted" (i.e. deleted locally but not synced) and 
    // we're doing a save on it, we should "undelete" it. From the server's 
    // perspective, it's "updated"
    newStatus = "updated";
} else if (synced) {
    // The `synced` option means we should force it to "synced" status
    newStatus = "synced";
} else {
    // Otherwise, default to whatever the user gave us. 
    // If that's missing, copy from the old record. 
    // If there was no old record, then we're "created".
    newStatus = updated._status || oldStatus || "created";
}
```

It's a lot longer, but I feel like it's a lot simpler by virtue of only specifying the cases when a status should change instead of the cases where it shouldn't.

glasserc:
Yes, this is much cleaner. r+wc

glasserc:
Oh, maybe I shouldn't have merged the other one, since that means this one needs a rebase.

Natim:
The other way around would have done a merge conflict anyway you can merge this one on top of master if you want.

glasserc:
I'll leave it up to @leplatrem since I also commented on thi code.

src/collection.js
leplatrem: I'm embarrassed because I think the ternary operators are actually simpler and easier to read: *«`synced` if keepSynced or synced is forced, else `created` if neverSynced else `updated`»* . 

That's a lot of code to read and process which relies boolean arithmetic (e.g. newStatus is updated when oldStatus is deleted only if never synced!).

After discussing with @n1k0, he seemed to agree also :/

Note that chaining ternary operators is not exactly prohibited, especially if the variables in the conditions are well named.

n1k0: What I really appreciate with the ternary operator form here is that you can almost read it as if it were plain English, thanks to the well named variables :)

PR #468, opened 2016-06-10
Return deleted records

glasserc:
Another fix for my webextensions work. This facilitates generating notifications for deleted records.


n1k0:
LGTM, r+.

PR #470, opened 2016-06-14
Leverage the babel rest-object spread plugin #304

### LINKED ISSUE #304: Leverage the babel rest-object spread plugin
In order to avoid the verbosity of `Object.assign`, we should use [ES6 Object Rest/Spread Properties](https://github.com/sebmarkbage/ecmascript-rest-spread).

So instead of writing things like

```js
const updated = Object.assign({}, initial, {_status: "synced"});
```

We'd get

```js
const updated = {...initial, _status: "synced"};
```

This implies enabling the [object rest spread babel plugin](https://www.npmjs.com/package/babel-plugin-transform-object-rest-spread).

lavish205:
Fixes #304

### DIFF
--- a/fx-src/index.js
+++ b/fx-src/index.js
@@ -60,13 +60,13 @@ export default function loadKinto() {
         adapter: FirefoxAdapter
       };
 
-      const expandedOptions = Object.assign(defaults, options);
+      const expandedOptions = {...defaults, ...options};
       super(expandedOptions);
     }
 
     collection(collName, options={}) {
       const idSchema = makeIDSchema();
-      const expandedOptions = Object.assign({idSchema}, options);
+      const expandedOptions = {...idSchema, ...options};
       return super.collection(collName, expandedOptions);
     }
   }
--- a/src/KintoBase.js
+++ b/src/KintoBase.js
@@ -55,7 +55,7 @@ export default class KintoBase {
       bucket: DEFAULT_BUCKET_NAME,
       remote: DEFAULT_REMOTE,
     };
-    this._options = Object.assign(defaults, options);
+    this._options = {...defaults, ...options};
     if (!this._options.adapter) {
       throw new Error("No adapter provided");
     }
--- a/src/collection.js
+++ b/src/collection.js
@@ -101,7 +101,7 @@ function createUUIDSchema() {
 }
 
 function markStatus(record, status) {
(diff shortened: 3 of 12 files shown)


Natim:
Please rebase on top of master.

fx-src/index.js
```diff
@@ -43,7 +43,7 @@ export default function loadKinto() {
         events: emitter
       };
 
-      const expandedOptions = Object.assign(defaults, options);
+      const expandedOptions = {...defaults, options};
```
leplatrem: I think in this case it should be `{...defaults, ...options}`

src/collection.js
```diff
@@ -59,6 +59,7 @@ export class SyncResultObject {
      */
     this.ok = true;
     Object.assign(this, SyncResultObject.defaults);
+    ///{...this, ...SyncResultObject.defaults};
```
leplatrem: leftover?

src/collection.js
```diff
@@ -814,7 +815,9 @@ export default class Collection {
         // Merge outgoing conflicts into sync result object
         syncResultObject.add("conflicts", conflicts);
         // Reflect publication results locally
-        const missingRemotely = skipped.map(r => Object.assign({}, r, {deleted: true}));
+        const missingRemotely = skipped.map(r => {
+          return {...r, deleted: true};
+        });
```
leplatrem: nitpick: you can write this on one line:
```js
const missingRemotely = skipped.map(r => ({...r, deleted: true}));
```

src/index.js
```diff
@@ -29,7 +29,7 @@ export default class Kinto extends KintoBase {
       events: new EventEmitter()
     };
 
-    super(Object.assign({}, defaults, options));
+    super({...defaults, options});
```
leplatrem: in this case it's `{...defaults, ...options}`

src/KintoBase.js
```diff
@@ -55,7 +55,7 @@ export default class KintoBase {
       bucket: DEFAULT_BUCKET_NAME,
       remote: DEFAULT_REMOTE,
     };
-    this._options = Object.assign(defaults, options);
+    this._options = {...defaults, options};
```
leplatrem: I think in this case it should be `{...defaults, ...options}`

PR #476, opened 2016-06-21
Adapter#execute preload option takes IDs

glasserc:
... instead of records. This matches both the underlying semantics of
the operation (which is fetching these records) as well as the use case
I'm about to implement (which only has access to the IDs).

r? @leplatrem @n1k0


leplatrem:
r+

Issue #464, opened 2016-06-09
Remove coupling between synchronization and storage


leplatrem:
Currently the `collection.js` file handles both storage and synchronization.

We had a discussion around January (I remember diagrams were drawn :)) to split the implementation into two pieces.

The idea would be to have a series of agnostic functions that would receive a list of records to sync and return a sync result.
This would allow, for example, to use the synchronization algorithm with completely different storage implementation (e.g. key-value store instead of our storage adapters etc.)

A first step would be at least to split the source file into two: `collection.js` and `sync.js`. 
The sync function would receive as input the result of `gatherLocalChanges()` but would not be in charge of inspecting the local data.

My only interrogation concerns the split of conflicts detection. But we could think of something using callbacks for example. Importing records locally could be done using the information obtained from the sync result object.

Note that it would help solving #340.

PR #477, opened 2016-06-21
Add transactions to Collection layer (fixes #467)

### LINKED ISSUE #467: implement and document a Collection#batch
See #466. This API would let you record a bunch of writes and execute them atomically, either all succeeding or all failing as a whole.

glasserc:
This is a proof-of-concept of transaction support at the Collection layer (#467). I would like to solicit feedback.

The approach in this branch is to wrap the underlying DB/adapter transaction using a CollectionTransaction and slowly move Collection methods into the CollectionTransaction class. Rather than delete the existing methods, we leave them as "atomic" operations that wrap them in an immediate `execute`.

So far I have moved `get`, `getRaw`, `delete`, and `deleteAny`. There remain `put`, `update`, `create`, and maybe some others.

This approach makes each of the CollectionTransaction functions synchronous, which means that their output can be returned by the user's function however it wants. So example use might be:

```javascript
collection.execute(txn => {
    return [
        txn.deleteAny("id1"),
        txn.deleteAny("id2")
    ];
}).then(res => {
    assertEqual(res[0], {deleted: true, data: {myKey: "a value"}});
    assertEqual(res[1], {deleted: false, data: undefined});
});
```

I wasn't really sure what to do with the tests. I left the tests on Collection for now and merely added some happy-path stuff to the `execute` method. Maybe I should move the tests to a new `CollectionTransaction` block, and add tests to the `Collection` class that verify that it calls `execute`? That seemed like overkill to me.

I wasn't really sure what to call `CollectionTransaction` or where to put it. For now it's in collection.js. I'm open to alternatives.

r? @leplatrem @n1k0

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -373,6 +373,58 @@ Result:
 > - Clearing the local collection will mark the collection as never synchronized;
 > - Detailed API documentation for `Collection#clear()` is available [here](https://doc.esdoc.org/github.com/Kinto/kinto.js/class/src/collection.js~Collection.html#instance-method-clear).
 
+## Transactions
+
+Kinto.js exposes a concept of a "local transaction", which guarantees
+that a set of local operations will happen atomically -- that is, all
+or nothing.
+
+Note that as Kinto is still a distributed database, transactions
+aren't consistent. In other words, your "transaction" can still
+conflict with other operations on some other node.
+
+To initiate a transaction, call `Collection#execute()` like this:
+
+```js
+articles.execute(txn => {
+    let article1 = txn.get(id1);
+    let article2 = txn.get(id2);
+    return [article1, article2];
+  }, {preloadIds: [id1, id2])
+  .then(console.log.bind(console));
+```
+
+The `execute` function takes two arguments. The first is a function
+that will be called with a transaction, and can perform operations on
+it. These operations are synchronous and so don't produce
+promises. The full list of operations is available
+[here](https://doc.esdoc.org/github.com/Kinto/kinto.js/class/src/collection.js~CollectionTransaction.html).
+
+The second argument to `execute()` should include a set of record IDs
+on which your transaction wants to operate. These IDs will be read at
+the beginning of your transaction, and the corresponding records will
+be made available to the transaction. Most operations, including even
+`put` and `delete`, will require that you provide the relevant IDs.
+
+Result:
+
+```js
+[
+    {
+      data: {
+        id: "2dcd0e65-468c-4655-8015-30c8b3a1c8f8",
+        title: "foo",
+      }
+    },
+    {
+      data: {
+        id: "2dcd0e65-468c-4655-8015-30c8b3a1c8f7",
+        title: "bar",
+      }
+    }
+]
+```
+
 ## Authentication
 
 Authenticating against a Kinto server can be achieved by adding an `Authorization` header to the request.
--- a/docs/api.md
+++ b/docs/api.md
@@ -379,9 +379,10 @@ Kinto.js exposes a concept of a "local transaction", which guarantees
 that a set of local operations will happen atomically -- that is, all
 or nothing.
 
-Note that as Kinto is still a distributed database, transactions
-aren't consistent. In other words, your "transaction" can still
-conflict with other operations on some other node.
+Note that these transactions are local to the browser, and they don't
+go to the server. Other Kinto clients can make changes to the server
+during your transaction, and those changes may still introduce
+conflicts with the changes you've made as part of a transaction.
 
 To initiate a transaction, call `Collection#execute()` like this:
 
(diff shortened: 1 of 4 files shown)


src/collection.js
```diff
@@ -369,7 +369,7 @@ export default class Collection {
 
     const validatedHooks = {};
 
-    for (let hook in hooks) {
+    for (const hook in hooks) {
```
n1k0: We can't use this when targeting Firefox just yet.

src/collection.js
```diff
@@ -490,7 +490,7 @@ export default class Collection {
     if (typeof(record) !== "object") {
       return Promise.reject(new Error("Record is not an object."));
     }
-    if (!record.id) {
+    if (record.id === undefined) {
```
n1k0: Maybe `record.hasOwnProperty("id")`

src/collection.js
```diff
@@ -747,6 +707,38 @@ export default class Collection {
...
+   * called with a transaction. Collection methods are available on
+   * this transaction, and they will return promises as normal. The
+   * promises for individual operations will be resolved after the
+   * transaction commits.
+   *
+   * If you also want to perform read operations like {@link
+   * Collection.get}, pass a second argument which specifies the IDs
+   * of the objects you want to preload as part of the operation.
+   *
+   * @return {Promise} Resolves with the result of the given function
```
n1k0: Nit: it may be worth documenting params as well.

src/collection.js
```diff
@@ -747,6 +707,38 @@ export default class Collection {
...
+   *    when the transaction commits.
+   */
+  execute(doOperations, {preloadIds = []} = {}) {
+    for(const id of preloadIds) {
```
n1k0: nit: space after for

src/utils.js
```diff
@@ -131,7 +131,7 @@ export function deepEqual(a, b) {
   if (Object.keys(a).length !== Object.keys(b).length) {
     return false;
   }
-  for(let k in a) {
+  for(const k in a) {
```
Natim: nit: space after for

test/collection_test.js
```diff
@@ -2265,4 +2278,88 @@ describe("Collection", () => {
...
+  /** @test {Collection#execute} */
+  describe("#execute", () => {
+    it("should support get", () => {
+      const articles = testCollection();
+      return articles.create(article)
+        .then(result => {
+          const id = result.data.id;
+          return articles.execute(txn => txn.get(id), {preloadIds: [id]});
+        })
+        .then(result => expect(result.data.title).eql("foo"));
```
n1k0: Nit: please put the `.eql` call on its own line, I've found that most often this eases reading the tests a lot.

n1k0:
Looks good so far! I like we keep backward compatible atomic operations in the collection instance.

`CollectionTransaction` sounds just fine to me and couldn't come with a better name. For tests, ensuring we call `execute` with the expected arguments looks indeed overkill but I'd feel uncomfortable not testing for that.

Good work!

leplatrem:
It looks good indeed!

We can keep high level tests, since this is what we expose in the docs. Having added a test suites dedicated to execute seems enough.

Let's find a way to document that nicely :)

We can leave the transaction class in the collection.js file. The split that will have to occur some day is #464

glasserc:
OK. I have pushed more commits that address the outstanding issues, extract `put`, `create`, and `update`, and also add some more tests to the `execute` function. I'm removing the `WIP` tag from this PR.

docs/api.md
```diff
@@ -373,6 +373,58 @@ Result:
...
 
+## Transactions
+
+Kinto.js exposes a concept of a "local transaction", which guarantees
+that a set of local operations will happen atomically -- that is, all
+or nothing.
+
+Note that as Kinto is still a distributed database, transactions
+aren't consistent. In other words, your "transaction" can still
+conflict with other operations on some other node.
```
leplatrem: nitpick: I believe some abstraction could be removed from this paragraph. What does it mean for me frontend developer? Is Kinto really «distributed»? What is a «node»?

docs/api.md
```diff
@@ -373,6 +373,58 @@ Result:
...
+that will be called with a transaction, and can perform operations on
+it. These operations are synchronous and so don't produce
+promises. The full list of operations is available
+[here](https://doc.esdoc.org/github.com/Kinto/kinto.js/class/src/collection.js~CollectionTransaction.html).
+
+The second argument to `execute()` should include a set of record IDs
+on which your transaction wants to operate. These IDs will be read at
+the beginning of your transaction, and the corresponding records will
+be made available to the transaction. Most operations, including even
+`put` and `delete`, will require that you provide the relevant IDs.
```
leplatrem: nitpick++: `put()` and `delete()`

src/collection.js
```diff
@@ -1187,3 +1145,207 @@ export default class Collection {
...
+   * @param  {String} id
+   * @param  {Object} options
+   * @return {Object}
+   */
+  get(id, options={includeDeleted: false}) {
+    const res = this.getRaw(id);
+    if (!res.data ||
+        (!options.includeDeleted && res.data._status === "deleted")) {
+      throw new Error(`Record with id=${id} not found.`);
+    } else {
```
leplatrem: nitpick: else is superfluous because of throw

leplatrem:
My feedbacks are only nits :) r+ !

PR #478, opened 2016-06-23
build listRecordsById query dynamically

glasserc:
The old code ended up generating queries like:

SELECT record_id, record
  FROM collection_data
    WHERE collection_name = 'foo'
      AND record_id IN ('1\',\'2\',\'3');

... which is completely wrong. Instead of generating a list of values,
it just assumed the quotes and commas were part of one large value.

According to http://stackoverflow.com/a/4789034, "There is no way to
directly bind a text parameter to multiple integer (or, for that matter,
multiple text) parameters." Instead we commit this atrocity.

r? @leplatrem

### DIFF
--- a/fx-src/FirefoxStorage.js
+++ b/fx-src/FirefoxStorage.js
@@ -75,11 +75,13 @@ const statements = {
       FROM collection_data
         WHERE collection_name = :collection_name;`,
 
+  // N.B. we have to have a dynamic number of placeholders, which you
+  // can't do without building your own statement. See `execute` for details
   "listRecordsById": `
     SELECT record_id, record
       FROM collection_data
-        WHERE collection_name = :collection_name
-          AND record_id IN (:record_ids);`,
+        WHERE collection_name = ?
+          AND record_id IN `,
 
   "importData": `
     REPLACE INTO collection_data (collection_name, record_id, record)
@@ -163,12 +165,17 @@ export default class FirefoxAdapter extends BaseAdapter {
 
     return conn.executeTransaction(function* doExecuteTransaction() {
       // Preload specified records from DB, within transaction.
-      const parameters = {
-        collection_name: collection,
-        // XXX: would be easier if a list of strings could be bound instead.
-        record_ids: options.preload.join("','")
-      };
-      const rows = yield conn.execute(statements.listRecordsById, parameters);
+      const parameters = [
+        collection,
+        ...options.preload
+      ];
+      const placeholders = [];
+      for (let i in options.preload) {
+        placeholders.push('?');
+      }
+      const stmt = statements.listRecordsById + "(" + placeholders.join(",") + ");";
+      const rows = yield conn.execute(stmt,
+                                      parameters);
 
       const preloaded = rows.reduce((acc, row) => {
         const record = JSON.parse(row.getResultByName("record"));


n1k0:
Do'h.

cc @mozmark

fx-src/FirefoxStorage.js
```diff
@@ -163,12 +165,17 @@ export default class FirefoxAdapter extends BaseAdapter {
...
-      };
-      const rows = yield conn.execute(statements.listRecordsById, parameters);
+      const parameters = [
+        collection,
+        ...options.preload
+      ];
+      const placeholders = [];
+      for (let i in options.preload) {
+        placeholders.push('?');
+      }
```
n1k0:
```js
const placeholders = options.preload.map(_ => "?")
```

glasserc: Nice! We don't run eslint on this directory, so no problems...

fx-src/FirefoxStorage.js
```diff
@@ -163,12 +165,14 @@ export default class FirefoxAdapter extends BaseAdapter {
...
-        // XXX: would be easier if a list of strings could be bound instead.
-        record_ids: options.preload.join("','")
-      };
-      const rows = yield conn.execute(statements.listRecordsById, parameters);
+      const parameters = [
+        collection,
+        ...options.preload
+      ];
+      const placeholders = options.preload.map(_ => "?");
+      const stmt = statements.listRecordsById + "(" + placeholders.join(",") + ");";
```
Natim: Could you explain me why this is better than the previous version? It genuinely though it was a bad idea to concatenate parameters like that without having an SQL injection escaping (usually done by the client)? I think it was the previous behavior intention.

n1k0: `record_id IN ('1\',\'2\',\'3')` wasn't working very well I suppose.

glasserc: Right. The previous version wasn't working, and this one does, so that makes it better :) The old version tried to generate a string that would generate a list by "unescaping" the commas, but of course it actually got escaped because it was fed into the query using a placeholder, so only turned into one giant string rather than a list of strings.

Yes, this is horrible, but I'm not super worried about the potential for a SQL injection attack -- we aren't doing interpolation ourselves, we're just generating a string with a variable number of placeholders. Substituting the placeholders for actual values is still done by the SQL library.

Note that I think we use a similar technique in the Kinto server at https://github.com/Kinto/kinto/blob/master/kinto/core/permission/postgresql/__init__.py#L285-L324.

Natim:
Thank you for the clarifications, do you think it would be possible to add a test that was failing before and now passes with that fix? I know that this kind of tests should be in Gecko, but can you link the patch to Gecko here?

Thanks.

n1k0:
> I know that this kind of tests should be in Gecko

FWIW note: I still think this whole adapter code source + tests should live in Gecko, or in a dedicated gh repo with proper isolation (setting up a gecko test platform on travis will be hard though); we've been discussed just this many times already, I just wish it could happen *soon*, because I agree that having implementation here and tests elsewhere is not *healthy* :)

n1k0:
Also note that with heavy stubbing we could probably start writing a minimal set of unit tests covering the non-gecko related logic here. This might actually be the easiest way to go as a very first step.

n1k0:
We've paired with @Natim and started a simple unit test suite for the FirefoxAdapter here https://github.com/glasserc/kinto.js/pull/2

glasserc:
Awesome, thanks!

n1k0:
r+!

PR #480, opened 2016-06-28
Rename raw methods (fixes #474)

### LINKED ISSUE #474: Naming of «raw» methods
`getRaw()`, `put()` and `deleteAny()` were introduced. I don't find the naming very affordant.

What do you think if we rename them to `rawGet()`, `rawUpdate()`, `rawDelete()` instead? I believe it would be easier to grasp that they are *raw* alternatives to the common get, update and delete methods, and could be documented in a single doc section.

leplatrem:
fixes #474

@glasserc r?


glasserc:
r+

PR #488, opened 2016-07-01
restore idSchema in fx-src

glasserc:
The intended shape of the expanded options is:

```
{
  idSchema: {
    generate: ....
    validate: ....
  },
  ... other options
}
```

I don't see an easy way to test this because it calls its superclass,
which isn't easy to stub or mock.

This is one of two changes that need to be made to get FF tests to pass (the other is in the FF test itself).

r? @leplatrem @n1k0 @Natim @lavish205


leplatrem:
r+

PR #491, opened 2016-07-05
bump version to 3.1.1

glasserc:
This release fixes a bug in the FirefoxAdapter part (#488) of the Kinto client and updates a few dependencies.

r? @n1k0 @leplatrem

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "kinto",
-  "version": "3.1.0",
+  "version": "3.1.1",
   "description": "An Offline-First JavaScript client for Kinto.",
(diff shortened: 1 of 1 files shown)


leplatrem:
nothing against this change :) r+

as said before, we could post the resulting changelog in the PR so that release preparation can be fully reviewed

glasserc:
This was the changelog! The only other changes that happened to Kinto.js since the last release were Greenkeeper dependency bumps, which I didn't think were worth mentioning in the release notes. But I can add them if you like...

leplatrem:
Ah ok :) no problem then! :shipit: :)

Issue #490, opened 2016-07-05
Add CRUD events 


leplatrem:
```js
collection.on('delete', (event) => {...})
```

Same for create and update.

glasserc:
We also have an implementation of something like this in the storage.sync extension which could be simplified if kinto.js offered this.

leplatrem:
Excellent to hear! 
Please don't hesitate to open issues if you think there are more minor improvements like this that could make your life easier :)

PR #493, opened 2016-07-06
Trigger events on collection operations (ref #490)

leplatrem:
* [x] Send events
* [x] Adjust payload
* [x] Should we have a single event `"changed" {action: "create", ...} ` ?
* [x] feedback? @glasserc 
* [x] Documentation

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -782,9 +782,9 @@ const kinto = new Kinto();
 const articles = kinto.collection("articles");
 
 articles.events.on("change", function(event) {
-  const first = events.targets[0];
+  const first = event.targets[0];
   console.log(first.action);
-  console.log(first.data);
+  console.log(first.data.id);
 });
 
 articles.delete(recordId);
@@ -798,7 +798,7 @@ Every event contains the following attributes:
 
 - ``data``: the record that was created, updated or deleted
 
-The ``update`` event contains an additionnal attribute:
+The ``update`` event contains an additional attribute:
 
 - ``oldRecord``: the previous version of the updated record.
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -732,9 +732,11 @@ articles.sync({ignoreBackoff: true})
 
 ## Events
 
-The `Kinto` instance and its other dependencies expose an `events` property you can subscribe public events from. That `events` property implements nodejs' [EventEmitter interface](https://nodejs.org/api/events.html#events_class_events_eventemitter).
+### `Kinto` instance
 
-### The `backoff` event
+Using the `events` on a `Kinto` instance property you can subscribe public events from. That `events` property implements nodejs' [EventEmitter interface](https://nodejs.org/api/events.html#events_class_events_eventemitter).
+
+#### The `backoff` event
 
 Triggered when a `Backoff` HTTP header has been received from the last received response from the server, meaning clients should hold on performing further requests during a given amount of time.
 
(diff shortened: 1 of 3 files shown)


glasserc:
How should this interact with transactions? My current implementation of this runs a transaction with a bunch of sets (or deletes, or whatever) and then emits one event only if the transaction succeeds. Of course, this means that the format of the event is like `{id: {oldValue: ..., newValue: ...}}`.

leplatrem:
> then emits one event only if the transaction succeeds

Good catch! I'll rework this!

docs/api.md
```diff
@@ -765,6 +767,60 @@ kinto.events.on("deprecated", function(event) {
...
+The `event` argument contains the following information:
+
+- `targets`: a list of mappings with an `action` attribute and the other attributes of the respective event as described below.
+
+```js
+const kinto = new Kinto();
+const articles = kinto.collection("articles");
+
+articles.events.on("change", function(event) {
+  const first = events.targets[0];
```
glasserc: I think this should be `event`.

docs/api.md
```diff
@@ -765,6 +767,60 @@ kinto.events.on("deprecated", function(event) {
...
+
+#### Per action event
+
+Atomic events are sent for ``create``, ``update`` and ``delete``.
+
+Every event contains the following attributes:
+
+- ``data``: the record that was created, updated or deleted
+
+The ``update`` event contains an additionnal attribute:
```
glasserc: additional

glasserc: Also, is there an `id` provided?

src/collection.js
```diff
@@ -1159,6 +1161,27 @@ export class CollectionTransaction {
   constructor(collection, adapterTransaction) {
     this.collection = collection;
     this.adapterTransaction = adapterTransaction;
+
+    this._events = [];
+  }
+
+  _windEvent(action, payload) {
```
glasserc: I've never seen the word "wind"/"wound" used in this kind of context. It's only a private method so it's probably not a big deal. What do you think about "queued" or "buffered"?

test/collection_test.js
```diff
@@ -2447,4 +2447,160 @@ describe("Collection", () => {
...
+    beforeEach(() => {
+      articles = testCollection();
+      return articles.create({title: "foo"})
+        .then(({data}) => article = data);
+    });
+
+    it("should emit an event on create", (done) => {
+      articles.events.on("create", () => done());
+      articles.create({"title": "win"});
+    });
```
glasserc: How do these tests work? If you don't call `done()`, does the test fail?

test/collection_test.js
```diff
@@ -2447,4 +2447,160 @@ describe("Collection", () => {
...
+
+    it("should emit an event on deleteAny", (done) => {
+      articles.events.on("delete", () => done());
+      articles.deleteAny(article.id);
+    });
+
+    it("should not emit if deleteAny fails", (done) => {
+      articles.events.on("delete", () => done(new Error("fail")));
+      articles.deleteAny(uuid4());
+      setTimeout(() => done(), 5);
```
glasserc: Instead of `setTimeout`, can you just use `.then()` ?

docs/api.md
glasserc: OK, I see that there's no special field for `id`, but you can retrieve them from the records provided as `data` and `oldRecord`.

test/collection_test.js
glasserc: I see. If you don't call `done()`, the test fails after a timeout of 2 sec.

glasserc:
This doesn't even require changes to `fx-src` because an event emitter has been present for almost a year now. r+ with nits. Great work!

PR #496, opened 2016-07-13
Return promises

### LINKED ISSUE #495: hooks cannot return promises
If they do, we get a `{} has no 'changes' property`, per https://github.com/Kinto/kinto.js/blob/master/src/collection.js#L834-L842. This sanity check was introduced in https://github.com/Kinto/kinto.js/pull/436.

glasserc:
Fixes #495. Also includes some fixes for other things that distracted me when I was trying to get the tests to work correctly.

r? @leplatrem @Natim

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -834,10 +834,10 @@ export default class Collection {
     return waterfall(this.hooks[hookName].map(hook => {
       return record => {
         const result = hook(payload, this);
-        const resultThenable = result && result.then;
+        const resultThenable = result && typeof result.then === "function";
         const resultChanges = result && result.hasOwnProperty("changes");
         if (!(resultThenable || resultChanges)) {
-          throw new Error(`Invalid return value for hook: ${JSON.stringify(result)} has no 'then' or 'changes' properties`);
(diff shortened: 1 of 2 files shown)


src/collection.js
```diff
@@ -834,8 +834,10 @@ export default class Collection {
     return waterfall(this.hooks[hookName].map(hook => {
       return record => {
         const result = hook(payload, this);
-        if (!result || !result.hasOwnProperty("changes")) {
-          throw new Error(`Invalid return value for hook: ${JSON.stringify(result)} has no 'changes' property`);
+        const resultThenable = result && result.then;
```
leplatrem: nit: I would test something like `typeof result.then == 'function'` instead

src/collection.js
```diff
@@ -834,8 +834,10 @@ export default class Collection {
...
+        const resultChanges = result && result.hasOwnProperty("changes");
+        if (!(resultThenable || resultChanges)) {
+          throw new Error(`Invalid return value for hook: ${JSON.stringify(result)} has no 'then' or 'changes' properties`);
```
leplatrem: nit: `has no 'then()' or 'changes' property`

leplatrem:
r+ with nitpicks or not

PR #497, opened 2016-07-14
Prepare 3.1.2

glasserc:
This release fixes a bug with the `incoming-changes` hook which prevents hooks from returning Promises (#495, #496).

r? @leplatrem


glasserc:
Self merging because I'm blocking @mozmark .

PR #502, opened 2016-07-20
Upgrade Kinto Server version.

Natim:
r? @n1k0


n1k0:
r+ with a green pill.

PR #510, opened 2016-08-10
Transform even deleted records

glasserc:
The docs say that remote transformers must preserve `id`. I can't figure out why that should be the case, except that it enables a tiny optimization where we don't have to run the transformer on deleted records. This might be intentional, in case we wanted to spare client code from handling deleted records specially.

In my use case, I'd like to obfuscate record IDs on the server so that they don't leak information about what application is using Kinto locally.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -860,7 +860,8 @@ coll = kinto.collection("articles", {
 >
 > - The `decode` method should be the strict reverse version of `encode`;
 > - Your transformer will be called even on deleted records, so be sure to handle those correctly in both encoding and decoding;
-> - Be careful when altering `record.id`, since this can interfere with syncing;
+> - Most transformers should pass `id` and `last_modified` through unaltered, since they are used in syncing;
+> - If you do alter `id` or `last_modified`, be careful, since this can interfere with syncing;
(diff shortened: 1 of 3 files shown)


Natim:
Feedback+ I don't know what would be the impact on existing Syncto code.

glasserc:
Specifically @Natim points to https://github.com/mozilla-b2g/gaia/blob/26fb1315c57b029eb66cc1421d5df7bbd0d9acb3/shared/js/sync/engine.js#L112, which would certainly be affected. It's probably possible to make this change in a backwards-compatible way using an API like `remoteIDTransformers`, but I'm hesitant to do that unless we can't find another simpler strategy to avoid breakage. Changing the transformer in Gaia to avoid transforming deleted records (which is what all transformers do now) would be easy if we know who can review the code.

glasserc:
I asked seanlee on IRC, and he said:

```
<seanlee> Hey Ethan, gaia is not in development now due to Mozilla's
    strategy. If your changes won't affect to server-side codes of
    Kinto, it's fine to gaia.
```

leplatrem:
r+
(*...and it's slightly cleaner*)

PR #515, opened 2016-08-17
documentation: correct typo in list sorting

elelay:
add a note about sorting according to a single field


leplatrem:
Thanks for spotting and fixing!

PR #514, opened 2016-08-16
fix #513 - HTTPie invocation

elelay:
Following #513, adding bucket and sending the updated title inside a `{data: }` wrapper.

PR #516, opened 2016-08-17
Preparing 4.0.0

glasserc:
This release causes remote transformers to be invoked even on deleted records, and thereby allows you to write a remote transformer that mutates record IDs (#510). It is a backwards-incompatible change because now a remote transformer must handle deleted records, which are missing all the normal fields you would expect from your records.

There are also some fixes to documentation errors (#515, #514) and updates of dependencies (#508, #512, #509, #502).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "kinto",
-  "version": "3.1.2",
+  "version": "4.0.0",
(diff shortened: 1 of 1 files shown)


Natim:
Please merge #517 before but otherwise it looks good to me.

PR #518, opened 2016-08-18
Use correct ES6 name for string function

glasserc:
The demo wasn't working for me, failing that "err.message.contains is not a function". I believe that is because the function is called `includes` (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes).


leplatrem:
r+!

PR #520, opened 2016-08-18
Prepare 4.0.1

glasserc:
I made a mistake in releasing 4.0.0 and published something wrong to NPM. This release is just a rebuild of 4.0.0, which shouldn't be used.


leplatrem:
r+

PR #521, opened 2016-08-18
Update SRIs for 4.0.1

glasserc:
Second time's the charm...

r? @leplatrem


glasserc:
Just for fun, I also tried computing the SRIs against npmcdn and the one for kinto.min.js was slightly different. Following @n1k0's advice, I've chosen to use that one in this PR.

n1k0:
r+ after having discussed this over IRC. Basically we decided to compute the SRIs against the version actually published to npm.

PR #522, opened 2016-08-22
Fix conflict handling with published deletions.

n1k0:
This fixes an issue brought to our attention by @ochameau (thanks!), where a conflicting remote deletion couldn't be properly handled during the publication part of the synchronization flow.

Basically kinto-http.js, when aggregating results from a batch operation, exposes the request data body as the local (initial) version of the conflict; but in the case of deletion, there is no request body sent.

This patch simply ensures there's data available in the received `local` part of the conflict object, and fallbacks to `{}` if there's none.

r=? @leplatrem @glasserc

### DIFF
--- a/test/integration_test.js
+++ b/test/integration_test.js
@@ -720,7 +720,7 @@ describe("Integration tests", function() {
       });
 
       describe("Outgoing conflicting deletion", () => {
-        let id, syncResult;
+        let id, conflicts;
 
         beforeEach(() => {
           return tasks.create({title: "initial"})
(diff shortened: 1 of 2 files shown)


Natim:
looks good to me :+1: Great job !

glasserc:
:+1:

test/integration_test.js
```diff
@@ -719,6 +719,57 @@ describe("Integration tests", function() {
...
+            .then(() => {
+              return fetch(`${TEST_KINTO_SERVER}/buckets/default/collections/tasks/records/${id}`, {
+                method: "PUT",
+                headers: {
+                  "Accept":        "application/json",
+                  "Content-Type":  "application/json",
+                  "Authorization": "Basic " + btoa("user:pass"),
+                },
+                body: JSON.stringify({data: {title: "server-updated", done: true}})
+              });
```
leplatrem: Nit: why not use kinto-client here ? 

Something like (*untested*) :
```js
return tasks.api.bucket("default").collection("tasks").deleteRecord(id)
```

n1k0: Haha, good point! Addressed in https://github.com/Kinto/kinto.js/pull/522/commits/7053627f17a5cb2d74f325e353462bcda0b84fd2

leplatrem:
r+

PR #523, opened 2016-08-22
WIP: Decode conflicts

glasserc:
If you use a remote transformer + SERVER_WINS syncing strategy + have a conflict, you can get undecoded records in your database. This seems to fix it for me but doesn't include tests. f? @leplatrem @n1k0 ?

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -889,11 +889,15 @@ export default class Collection {
 
         // The result of a batch returns data and permissions.
         // XXX: permissions are ignored currently.
-        const conflicts = synced.conflicts.map(({type, local, remote}) => {
+        const conflicts = Promise.all(synced.conflicts.map(({type, local, remote}) => {
           // Note: we ensure that local data are actually available, as they may
           // be missing in the case of a published deletion.
-          return {type, local: local && local.data || {}, remote};
-        });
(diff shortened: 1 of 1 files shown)


src/collection.js
```diff
@@ -889,11 +889,15 @@ export default class Collection {
 
         // The result of a batch returns data and permissions.
         // XXX: permissions are ignored currently.
-        const conflicts = synced.conflicts.map(({type, local, remote}) => {
+        const conflicts = Promise.all(synced.conflicts.map(({type, local, remote}) => {
           // Note: we ensure that local data are actually available, as they may
           // be missing in the case of a published deletion.
-          return {type, local: local && local.data || {}, remote};
-        });
+          return this._decodeRecord("remote", local || {}).then(realLocal => {
```
leplatrem: Do we really need to decode the local one?

n1k0: I share the same question :) Maybe writing a few tests would help figuring this out!

glasserc:
I've written a couple tests in #525. I pushed them as a separate branch/separate WIP PR because this fix seems to break a bunch of other stuff.

glasserc:
Closing this in favor of #525, which has the correct fix, correctly copied out of the other source tree that I got working.

PR #525, opened 2016-08-24
Decode conflicts

glasserc:
The test in the first commit fails. A similar bug exists for SERVER_WINS strategies, obviously, but I haven't written it yet. The tests in the second commit don't fail, but I had already written them and I want to make sure they don't start failing when I apply the fix.

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -856,6 +856,7 @@ export default class Collection {
       return Promise.resolve(syncResultObject);
     }
     const safe = !options.strategy || options.strategy !== Collection.CLIENT_WINS;
+    let synced;
 
     // Fetch local changes
     return this.gatherLocalChanges()
@@ -880,7 +881,8 @@ export default class Collection {
         }, {headers: options.headers, safe, aggregate: true});
       })
       // Update published local records
-      .then((synced) => {
+      .then(batchResult => {
+        synced = batchResult;
         // Merge outgoing errors into sync result object
         syncResultObject.add("errors", synced.errors.map(error => {
           error.type = "outgoing";
@@ -889,11 +891,17 @@ export default class Collection {
 
         // The result of a batch returns data and permissions.
         // XXX: permissions are ignored currently.
-        const conflicts = synced.conflicts.map(({type, local, remote}) => {
+        return Promise.all(synced.conflicts.map(({type, local, remote}) => {
           // Note: we ensure that local data are actually available, as they may
           // be missing in the case of a published deletion.
-          return {type, local: local && local.data || {}, remote};
-        });
+          const safeLocal = local && local.data || {};
+          return this._decodeRecord("remote", safeLocal).then(realLocal => {
+            return this._decodeRecord("remote", remote).then(realRemote => {
+              return {type, local: realLocal, remote: realRemote};
+            });
+          });
+        }));
+      }).then(conflicts => {
         // Merge outgoing conflicts into sync result object
         syncResultObject.add("conflicts", conflicts);
 
--- a/test/integration_test.js
+++ b/test/integration_test.js
@@ -852,15 +852,8 @@ describe("Integration tests", function() {
           // Ensure that the remote record looks like something that's
           // been transformed
           return collection._encodeRecord("remote", record).then(record => {
-            return fetch(`${TEST_KINTO_SERVER}/buckets/default/collections/${collection._name}/records`, {
-              method: "POST",
-              headers: {
-                "Accept":        "application/json",
-                "Content-Type":  "application/json",
-                "Authorization": "Basic " + btoa("user:pass"),
-              },
-              body: JSON.stringify({data: record})
-            });
+            return collection.api.bucket("default").collection(collection._name)
+              .createRecord(record);
           })
             .then(_ => collection.sync())
             .then(res => {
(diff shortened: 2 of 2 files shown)


glasserc:
OK, I figured out what was going wrong with the other tests. This seems to be correct now. r? @leplatrem @n1k0

test/integration_test.js
```diff
@@ -542,6 +546,62 @@ describe("Integration tests", function() {
...
+          });
+
+          it("should publish resolved conflict using local version", () => {
+            expect(syncResult.published).to.have.length.of(1);
+            expect(recordsEqual(syncResult.published[0], {
+              id: conflictingId,
+              title: "task4-local",
+              done: false,
+            })).eql(true);
+          });
```
glasserc: The code path for changes that come from `pullChanges` is pretty different, and records are decoded before they are checked for conflicts, so these tests doesn't contribute anything in the current codebase. However, the functionality is kind of important to me. I'd be OK with deleting them, or keeping them.

src/collection.js
```diff
@@ -889,11 +891,17 @@ export default class Collection {
...
           // Note: we ensure that local data are actually available, as they may
           // be missing in the case of a published deletion.
-          return {type, local: local && local.data || {}, remote};
-        });
+          const safeLocal = local && local.data || {};
+          return this._decodeRecord("remote", safeLocal).then(realLocal => {
+            return this._decodeRecord("remote", remote).then(realRemote => {
+              return {type, local: realLocal, remote: realRemote};
+            });
+          });
```
n1k0: This may improve legibility a little:

```js
const safeLocal = local && local.data || {};
return this._decodeRecord("remote", safeLocal)
  .then(realLocal => {
    return this._decodeRecord("remote", remote);
  })
  .then(realRemote => {
    return {type, local: realLocal, remote: realRemote};
  });
```

glasserc: If you do this, `realLocal` goes out of scope before you can use it.

test/integration_test.js
n1k0: Let's keep them.

test/integration_test.js
```diff
@@ -766,26 +846,35 @@ describe("Integration tests", function() {
...
+              return collection.delete(recordId, { virtual: false });
+            }).then(_ => collection.create({
               id: recordId,
               title: "task1-local",
               done: false
             }, { useRecordId: true }));
+        }
+
+        beforeEach(() => {
+          return setupConflict(tasks).then(() => setupConflict(tasksTransformed));
```
n1k0: In general I'm finding promise-based code more readable when `.then` calls are on their own line/block:

```js
return setupConflict(tasks)
  .then(() => setupConflict(tasksTransformed));
```

What do you think?

Note: we'll really have to consider movig to generators at some point:

```js
yield setupConflict(tasks)
yield setupConflict(tasksTransformed);
```

Later.

test/integration_test.js
```diff
@@ -1005,6 +1094,27 @@ describe("Integration tests", function() {
           });
         });
 
+        describe("CLIENT_WINS strategy with transformers", () => {
+          beforeEach(() => {
+            return tasksTransformed.sync({strategy: Kinto.syncStrategy.CLIENT_WINS})
+            .then(res => {
```
n1k0: nit: 2 spaces indent

test/integration_test.js
```diff
@@ -1110,6 +1220,40 @@ describe("Integration tests", function() {
             });
           });
         });
+
+        describe("SERVER_WINS strategy with transformers", () => {
+          beforeEach(() => {
+            return tasksTransformed.sync({strategy: Kinto.syncStrategy.SERVER_WINS})
+            .then(res => {
```
n1k0: Ditto.

test/integration_test.js
```diff
@@ -766,26 +846,35 @@ describe("Integration tests", function() {
...
-              "Accept":        "application/json",
-              "Content-Type":  "application/json",
-              "Authorization": "Basic " + btoa("user:pass"),
-            },
-            body: JSON.stringify({data: {title: "task1-remote", done: true}})
+          const record = {title: "task1-remote", done: true};
+          // Ensure that the remote record looks like something that's
+          // been transformed
+          return collection._encodeRecord("remote", record).then(record => {
+            return fetch(`${TEST_KINTO_SERVER}/buckets/default/collections/${collection._name}/records`, {
```
n1k0: I'm pretty sure you can use the kinto-http API here.

n1k0:
LGTM, nice work.

test/integration_test.js
glasserc: OK, no problem. Sometimes I do this to try to minimize the diff in surrounding lines, but here there was no reason to do so.

src/collection.js
n1k0: Bleh. Oh well.

test/integration_test.js
glasserc: Ah, that's a good point. I wasn't looking at it too closely, I thought it *was* the kinto-http API.

test/integration_test.js
n1k0: Yeah I've been doing that a lot as well for the same reason, but now in my experience spreading these blocks vertically really helps grasping the flow.

leplatrem:
:shipit:

PR #527, opened 2016-08-25
Fix IDB on Safari

### LINKED ISSUE #524: Safari only DataError: DOM IDBDatabase Exception 0: The data provided does not meet requirements
screen shot 2016-08-23 at 12 32 19 pm

Natim:
Attempt to fix #524


src/adapters/IDB.js
```diff
@@ -148,7 +148,8 @@ export default class IDB extends BaseAdapter {
         const db = event.target.result;
         // Main collection store
         const collStore = db.createObjectStore(this.dbname, {
-          keyPath: "id"
+          keyPath: "id",
+          autoIncrement: true
```
n1k0: Hmm, how will this behave with string ids as we use by default?

Natim: I removed it.

src/adapters/IDB.js
```diff
@@ -427,7 +427,8 @@ function transactionProxy(store, preloaded = []) {
     },
 
     update(record) {
-      store.put(record);
+      // We need to put undefined there so that Safari and IE are happy.
+      store.put(record, undefined);
```
n1k0: Well, I'd be surprised this changes anything as `store.put(record)` and `store.put(record, undefined)` are equivalent in js, where args have `undefined` as a value by default.

Natim: This is the case for Chrome and Firefox but apparently Safari and IE checks for the parameters length rather than if the value is set to `undefined`: http://stackoverflow.com/a/19318034/186202

n1k0: Okay, I don't have words.

leplatrem:
I'm a bit doubtful about the fix: they don't do that in PouchDB :https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-adapter-indexeddb/src/bulkDocs.js#L264

aegaas:
Just tried this using the demo in Safari 9 in Mac OS 10.11: Unfortunately it didn't help: here's the error from the demo's textarea:

```
{
  "ok": false,
  "lastModified": 1472113378860,
  "errors": [
    {
      "type": "incoming",
      "message": "DataError: DOM IDBDatabase Exception 0",
      "stack": "put@[native code]\nupdate@http://127.0.0.1:8080/kinto.js:938:1415\nimportChange@http://127.0.0.1:8080/kinto.js:944:1270\nhttp://127.0.0.1:8080/kinto.js:944:9212\nmap@[native code]\nhttp://127.0.0.1:8080/kinto.js:944:9180\nhttp://127.0.0.1:8080/kinto.js:938:4802\nhttp://127.0.0.1:8080/kinto.js:938:2599"
    }
  ],
  "created": [],
  "updated": [],
  "deleted": [],
  "published": [],
  "conflicts": [],
  "skipped": [],
  "resolved": []
}
```

aegaas:
@leplatrem I tried your jsfiddle as well in the same browser but ran into a new Safari error on JSFiddle. The same code runs great in Chrome.

https://github.com/jsfiddle/jsfiddle-issues/issues/628

aegaas:
This fix does not seem to help. We did learn that the kinto.js demo works in v3.0.0 but is broken in the next release v3.1.0. I recommend we close this PR and not merge to master.

PR #529, opened 2016-08-29
deepEqual: Remove instanceof Object check

glasserc:
This is useful when the Object constructor might differ from object to
object. This has been true for me when hacking on Gecko -- the objects
that make it to the deepEqual function are instances of Object (from the
kinto-http-client scope), but the reference Object points to
Object (from the kinto-offline-client scope). We might also expect this
sort of thing to come up when someone uses Kinto in a WebExtension; the
scope of the WebExtension will be different from the scope of the code
running on the page. Instead of using the "instanceof Object" check, use
the "typeof" check. Unfortunately, because JavaScript, null and
undefined also have typeof "object", so try to rule out falsy things
explicitly.

r? @n1k0

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -125,7 +125,7 @@ export function deepEqual(a, b) {
   if (typeof(a) !== typeof(b)) {
     return false;
   }
-  if (!(a instanceof Object) || !(b instanceof Object)) {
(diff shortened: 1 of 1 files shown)


n1k0:
This makes perfect sense, and the patch is fine. r+

Natim:
Javascript is so peaky...

PR #531, opened 2016-08-30
Compute SRIs from npmcdn and add version to table

glasserc:
Since the version of the webpacked assets is not reproducible, use curl to fetch the version of the asset from npmcdn and use that to compute the SRI. Also, add a note in the header of the table to show what version the SRIs correspond to.

r? @Natim @n1k0 @elelay

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -37,8 +37,7 @@ integrity with the hash provided below:
     </script>
 ```
 
-| Filename                | Hash                                                                    |
-|-------------------------|-------------------------------------------------------------------------|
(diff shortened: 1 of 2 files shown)


Natim:
gg :) r+

n1k0:
Woah, TIL about `npm view`. Smart script =) At some point we may want to move these long commands to separated bash scripts maybe.

r+

elelay:
sweet!

leplatrem:
So, we trust npmcdn then! ;)

PR #533, opened 2016-09-02
doc: NOP idSchema example

elelay:


docs/api.md
```diff
@@ -977,6 +977,24 @@ coll = kinto.collection("articles", {
...
 > - In a real application, you want to make sure you do not generate twice the same record ID on a collection. This dummy example doesn't take care of ID unicity. In case of ID conflict you may loose data.
 
+For ids chosen by your application (like "config", "last-save", etc.), you'll want a NOP id generator:
+```js
+const nopSchema = {
+    generate: function() {
+        throw new Error("can't generate keys");
+    },
+    validate: function(id) {
+        return true
```
glasserc: Missing semicolon?

elelay:
Added semicolon.
Good catch @glasserc !

Issue #535, opened 2016-09-02
lastModified isn't saved even though handleConflicts resolved a conflict


glasserc:
This is because `saveLastModified` is called from `importChanges`, which is called before `handleConflicts`. Additionally, `handleConflicts` doesn't clear the `ok` field for a `SyncResultObject` so even if the order were reversed, it would still refuse to save the `lastModified`.

glasserc:
Released in v4.0.4.

PR #536, opened 2016-09-04
Introduce failing test for updating lastModified

glasserc:
I wasn't sure what the correct behavior for manual conflict resolution
was, but I'm pretty sure that we should update lastModified for all of
these cases. So far, we only do so in three of them.

This is a test case illustrating the failure in #535.

### DIFF
--- a/test/integration_test.js
+++ b/test/integration_test.js
@@ -445,6 +445,7 @@ describe("Integration tests", function() {
 
           it("should have updated lastModified", () => {
             expect(syncResult.lastModified).to.equal(tasks.lastModified);
+            expect(tasks.db.getLastModified()).eventually.equal(syncResult.lastModified);
           });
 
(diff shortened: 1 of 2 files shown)


test/integration_test.js
```diff
@@ -443,6 +443,10 @@ describe("Integration tests", function() {
             expect(syncResult.lastModified).to.be.a("number");
           });
 
+          it("should have updated lastModified", () => {
+            expect(syncResult.lastModified).to.equal(tasks.lastModified);
```
n1k0: We should also check for the collection's `lastModified` value stored in IndexedDB (it's the actual one we use to fetch the next chunk of changes, if any).

glasserc:
@n1k0 I've added tests and implemented the fix you suggested. r?

n1k0:
LGTM, r+

Issue #534, opened 2016-09-02
optimisation: apply collection.list() filters to each document in the IDB adapter, not to the full array


elelay:
Searching through a big collection (8k records) is slow and allocates a lot of memory.
I attribute it to the fact that a big part of the collection is loaded in memory, when not searching by id, _status, last_modified.
What would be nice would be to apply filters to each document when it is fetched (in [IDB.js](https://github.com/Kinto/kinto.js/blob/master/src/adapters/IDB.js#L19).

leplatrem:
If I'm not mistaken, this is very related to #66 

We could be more explicit about this limitations in the docs (give order of magnitude): http://kintojs.readthedocs.io/en/latest/limitations/#in-memory-filteringordering

elelay:
I think filtering as you go is a lower hanging fruit than indexes (doesn't require incrementing the IDB version number, etc.).
I'll propose code in a PR for filtering as you go this week.

leplatrem:
Great! 👍

PR #543, opened 2016-09-10
534: filter documents as they are retrieved from IDB

elelay:
Code for #534 : don't put documents in the results to be discarded later but filter them earliest.

I thought it would give a moderate performance boost and indeed get 8% time improvement and no memory change (1% not significant)  for a query rejecting every document.

My test code (oldTickets contains 7626 documents, none match)
```
function perfTest(cnt, acc){
	if(cnt <= 0){
		avg = acc.length && (acc.reduce(function(s, x) { return s + x; }, 0) / acc.length);
		console.log("perfTest avg=", avg);
	}else{
		if(cnt % 10 === 0){
			console.log("perfTest", cnt, "remaining");
		}
		before = new Date();
		oldTickets.list({filters: {status: "open"}}).then(function(){
		  after = new Date();
		  //console.log("list() took", after - before, "ms");
		  acc.push(after - before);
		  perfTest(cnt - 1, acc);
		});
	}
}

perfTest(50, []);
```

1. load oldTickets at https://trackers.elelay.fr/trackers-base.html
2. quit firefox
3. open firefox with a single tab to https://trackers.elelay.fr/trackers-base.html
4. let the app load (labels on buttons)
5. run the test code above
6. quit firefox
7. repeat 3. - 6. with https://trackers.elelay.fr/trackers-filter-as-you-go.html

Results:

|   | average time | virtual mem  | resident mem |
| ---- | ----------------- | --------------- | ------------ |
| master | 2.156s | 1257MB | 360MB |
| filter-as-you-go | 1.969s | 1274MB | 357MB |
| **improvement** | 8.7% faster | 1.4% more | 0.8% less |

### DIFF
--- a/src/adapters/IDB.js
+++ b/src/adapters/IDB.js
@@ -1,7 +1,7 @@
 "use strict";
 
 import BaseAdapter from "./base.js";
-import { filterObject, sortObjects } from "../utils";
+import { filterObject, omitKeys, sortObjects } from "../utils";
 
 const INDEXED_FIELDS = ["id", "_status", "last_modified"];
 
@@ -332,10 +332,8 @@ export default class IDB extends BaseAdapter {
     return this.open().then(() => {
       return new Promise((resolve, reject) => {
         let results = [];
-        // Prepare filters
-        const remainingFilters = {...filters};
         // If `indexField` was used already, don't filter again.
-        delete remainingFilters[indexField];
+        const remainingFilters = omitKeys(filters, indexField);
 
         const {transaction, store} = this.prepare();
         createListRequest(store, indexField, value, remainingFilters, (_results) => {
--- a/src/adapters/IDB.js
+++ b/src/adapters/IDB.js
@@ -1,7 +1,7 @@
 "use strict";
 
 import BaseAdapter from "./base.js";
-import { reduceRecords } from "../utils";
+import { filterObject, sortObjects } from "../utils";
 
 const INDEXED_FIELDS = ["id", "_status", "last_modified"];
 
@@ -11,12 +11,14 @@ const INDEXED_FIELDS = ["id", "_status", "last_modified"];
  * @type {Object}
  */
 const cursorHandlers = {
-  all(done) {
+  all(filters, done) {
     const results = [];
     return function(event) {
       const cursor = event.target.result;
       if (cursor) {
-        results.push(cursor.value);
+        if(filterObject(filters, cursor.value)){
+          results.push(cursor.value);
+        }
         cursor.continue();
       } else {
         done(results);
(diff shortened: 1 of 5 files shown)


leplatrem:
I like it! 

I'm ok to merge! @n1k0 @glasserc any objection :) ?

src/adapters/IDB.js
```diff
@@ -11,12 +11,14 @@ const INDEXED_FIELDS = ["id", "_status", "last_modified"];
...
-  all(done) {
+  all(filters, done) {
     const results = [];
     return function(event) {
       const cursor = event.target.result;
       if (cursor) {
-        results.push(cursor.value);
+        if(filterObject(filters, cursor.value)){
+          results.push(cursor.value);
+        }
```
leplatrem: 👍 

Thanks for taking the time to contribute this!

src/adapters/IDB.js
```diff
@@ -340,12 +348,9 @@ export default class IDB extends BaseAdapter {
       });
     })
     .then((results) => {
-      // The resulting list of records is filtered and sorted.
-      const remainingFilters = {...filters};
-      // If `indexField` was used already, don't filter again.
-      delete remainingFilters[indexField];
+      // The resulting list of records is sorted.
       // XXX: with some efforts, this could be fully implemented using IDB API.
-      return reduceRecords(remainingFilters, params.order, results);
```
n1k0: As `reduceRecords` isn't used anymore, it should be removed (check for tests too)

src/adapters/IDB.js
```diff
@@ -329,8 +332,13 @@ export default class IDB extends BaseAdapter {
     return this.open().then(() => {
       return new Promise((resolve, reject) => {
         let results = [];
+        // Prepare filters
+        const remainingFilters = {...filters};
+        // If `indexField` was used already, don't filter again.
+        delete remainingFilters[indexField];
```
n1k0: I know you didn't write this code, but let's use the `omitKeys` function from the `utils` module here.

```js
const remainingFilters = omit(filters,  [indexKey]);
```

n1k0:
Looks awesome indeed, thanks!

src/adapters/IDB.js
elelay: done (b40140b)

src/adapters/IDB.js
elelay: done (a14d317)

n1k0:
This now just need rebasing against latest master and we're good to go methinks :)

leplatrem:
Oh careful with having removed the `reduceRecords()`, it is a breaking change! And it will break the Firefox storage layer (no big deal we can put the function there)

n1k0:
Wow, I never realized it was part of the public API. So that means bumping a major if we land this patch, which still sounds okay to me.

n1k0:
As @leplatrem mentioned, we still need to re-add the `reduceRecords` function locally in `./fx-src/FirefoxStorage.js`.

n1k0:
LGTM, thank you!

leplatrem:
Excellent contribution! Thank you!

PR #547, opened 2016-09-15
Fix skipped remote deletion (fixes #540)

### LINKED ISSUE #540: Failed to execute 'delete' on 'IDBObjectStore' exception while syncing
I'm used to have this exception on Chrome for Android, it took me a while to debug that via remote devtools.
I'm not sure it has anything to do with Chrome or android, I imagine it has to do with the state that exists on the device.

I traced this exception back to this code:
https://github.com/Kinto/kinto.js/blob/3630426125a04650069f87d2461a8dc62fc2a82a/src/collection.js#L928
```
 transaction.delete(record.id);
```
Here, record.id is undefined.  record is just `{ deleted: true }`.

leplatrem:
This pull-request relies on kinto-http.js 2.5.1 

The `skipped` attribute in sync result now contains the record id that was skipped instead of a plain 404 error response body.

An integration test now executes a skipped remote deletion.

@n1k0 r?

### DIFF
--- a/test/collection_test.js
+++ b/test/collection_test.js
@@ -2059,7 +2059,10 @@ describe("Collection", () => {
         published: [],
         errors:    [],
         conflicts: [],
-        skipped:   [{data: {id: records[0].id}}]
+        skipped:   [{
+          id: records[0].id,
+          error: {errno: 110, code: 404, error: "Not found"}
+        }]
       }));
       return articles.pushChanges(client, result)
         .then(_ => articles.get(records[0].id, {includeDeleted: true}))
--- a/test/collection_test.js
+++ b/test/collection_test.js
@@ -2061,7 +2061,7 @@ describe("Collection", () => {
         conflicts: [],
         skipped:   [{
           id: records[0].id,
-          error: {errno: 110, code: 404, error: "Not found"}
+          error: {errno: 110, code: 404, error: "Not found"},
         }]
       }));
       return articles.pushChanges(client, result)
--- a/test/integration_test.js
+++ b/test/integration_test.js
@@ -809,7 +809,6 @@ describe("Integration tests", function() {
 
       });
 
-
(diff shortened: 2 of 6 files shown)


test/integration_test.js
```diff
@@ -809,50 +809,87 @@ describe("Integration tests", function() {
 
       });
 
+
```
n1k0: Nit: A single line is enough

test/collection_test.js
```diff
@@ -2059,7 +2059,10 @@ describe("Collection", () => {
         published: [],
         errors:    [],
         conflicts: [],
-        skipped:   [{data: {id: records[0].id}}]
+        skipped:   [{
+          id: records[0].id,
+          error: {errno: 110, code: 404, error: "Not found"}
```
n1k0: Nit: trailing comma

test/integration_test.js
```diff
@@ -809,50 +809,87 @@ describe("Integration tests", function() {
...
+          it("should properly list the encountered conflict", () => {
+            expect(conflicts).to.have.length.of(1);
+          });
+
+          it("should list the proper type of conflict", () => {
+            expect(conflicts[0].type).eql("outgoing");
+          });
+
+          it("should have the expected conflicting local version", () => {
+            expect(conflicts[0].local).eql({});
```
n1k0: Nit: could we have a more explicit local version definition here? Empty object is vague and may be false-positive prone methinks...

test/integration_test.js
```diff
@@ -809,50 +809,87 @@ describe("Integration tests", function() {
...
+          });
+
+          it("should provide the record", () => {
+            expect(result.skipped[0])
+              .to.have.property("id").eql(id);
+          });
         });
       });
 
+
```
n1k0: Nit: single empty line

Issue #526, opened 2016-08-24
investigate using permanent storage for IndexedDB adapter


elelay:
This way there is no limit on the DB size.
I keep hitting the Quota Exceeded error when I'm low on free space (< 700M) and sync a biggish db.

https://developer.mozilla.org/en-US/docs/Web/API/IDBFactory/open#Example
in [src/adapters/IDB.js:145](https://github.com/Kinto/kinto.js/blob/master/src/adapters/IDB.js#L145)

```
const request = indexedDB.open(this.dbname, {version: 1, storage: "persistent"});
```

This generates a prompt to the Firefox user, so it could be an option in the Kinto.js api.
I don't know if it's supported on other browsers.

leplatrem:
This is specific to Firefox [according the docs](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Browser_storage_limits_and_eviction_criteria#Different_types_of_data_storage), but it is perfectly ok to add it if it doesn't harm other platforms!

elelay:
It would be easy to put it into adapterOptions of the Kinto constructor, but what do we do browsers not supporting it?
 - I don't know how to do feature testing for it;
 - browser sniffing would check for "gecko > 26";
 - or is it OK to let the application determine when to pass the option use it when it's there (it feels wrong because we hide other IndexedDB details but it's the easiest) ?
 - the [spec](https://www.w3.org/TR/IndexedDB/#widl-IDBFactory-open-IDBOpenDBRequest-DOMString-name-unsigned-long-long-version) doesn't specify what to do if the 2nd argument is not a long. Otherwise we could try with storage name + version,  catch any exception and fallback to just the version.

leplatrem:
Apparently this has changed in the meantime, and can be done outside kinto.js using the Storage API [0]

```js
if (navigator.storage && navigator.storage.persist)
  navigator.storage.persist().then(function(persistent) {
    if (persistent)
      console.log("Storage will not be cleared except by explicit user action");
    else
      console.log("Storage may be cleared by the UA under storage pressure.");
  });
}
```

I propose to close this. Please reopen if I'm wrong!

[0] https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Browser_storage_limits_and_eviction_criteria#Different_types_of_data_storage

PR #551, opened 2016-09-16
implement #526 permanent storage option for IDB adapter

elelay:
Tries to specify storage (Firefox experimental option) and falls back to
unspecified storage. Shouldn't hurt other platforms: they throw on open
with non integer version parameter (tested in chromium only).

1 integration test fails but not from my changes:

>  1) Integration tests Default server configuration Synchronization Outgoing conflicting deletion With remote deletion "before each" hook for "should properly list the encountered conflict":
>     DataError: Data provided to an operation does not meet requirements.

### DIFF
--- a/src/adapters/IDB.js
+++ b/src/adapters/IDB.js
@@ -116,11 +116,22 @@ export default class IDB extends BaseAdapter {
   /**
    * Constructor.
    *
-   * @param  {String} dbname The database nale.
+   * Options:
+   * - `{String}` `storage` The storage to use: "persistent" or "temporary".
+   *
+   * If storage is specified, kinto will try to open the IDB with the Firefox
+   * experimental storage option. If it fails, it will retry without the option
+   * so your storage choice will basically be ignored in other browsers.
+   * For more information on the storage option, see
+   * https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Browser_storage_limits_and_eviction_criteria#Firefox_specifics
+   *
+   * @param  {String} dbname The database name.
+   * @param  {Object} options The database options.
    */
-  constructor(dbname) {
+  constructor(dbname, options = {}) {
     super();
     this._db = null;
+    this._options = { ... options };
     // public properties
     /**
      * The database name.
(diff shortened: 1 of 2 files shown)


src/adapters/IDB.js
```diff
@@ -146,7 +157,20 @@ export default class IDB extends BaseAdapter {
...
+          request = indexedDB.open(this.dbname, {
+            storage: this._options.storage,
+            version: version
+          });
+        } catch(e) {
+          request = indexedDB.open(this.dbname, version);
+        }
+      } else {
+        request = indexedDB.open(this.dbname, version);
+      }
```
leplatrem: I think we could simplify this code a bit, especially this `if (this._options.storage)`, with something like `const options = this._options ? {version, ...this._options} : version;`

src/adapters/IDB.js
```diff
@@ -146,7 +157,20 @@ export default class IDB extends BaseAdapter {
...
-      const request = indexedDB.open(this.dbname, 1);
+      const version = 1;
+      let request;
+      if(this._options.storage) {
+        try {
+          request = indexedDB.open(this.dbname, {
+            storage: this._options.storage,
+            version: version
+          });
+        } catch(e) {
```
leplatrem: Could you please add a comment why we have to catch here? Maybe also add `console.warn()` no?

src/adapters/IDB.js
```diff
@@ -116,11 +116,22 @@ export default class IDB extends BaseAdapter {
...
+   * experimental storage option. If it fails, it will retry without the option
+   * so your storage choice will basically be ignored in other browsers.
+   * For more information on the storage option, see
+   * https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Browser_storage_limits_and_eviction_criteria#Firefox_specifics
+   *
+   * @param  {String} dbname The database name.
+   * @param  {Object} options The database options.
    */
-  constructor(dbname) {
+  constructor(dbname, options = {}) {
```
leplatrem: How can the options be passed from the outside? Please add some docs to mention the `permanent` option.

src/adapters/IDB.js
elelay: I want to try indexedDB.open() twice with fallback when there is this._options.storage.
How would you write it?

dstaley:
The `storage` property in Firefox is deprecated, with the replacement being `navigator.storage.persist`, which would be up to the user to call rather than something we should worry about in this library.

Issue #340, opened 2016-02-19
Import of data should be done after conflict handling.


almet:
This is a spinoff of https://github.com/Kinto/kinto.js/pull/302. Currently, the import of the data is done and then the conflicts are handled (see https://github.com/Kinto/kinto.js/blob/master/src/collection.js#L572-L583)

leplatrem:
@almet could you detail what was your use-case here?

You'd like to detect conflicts before altering the local db ?

almet:
It is just weird to me to first import data and then fix the conflicts,  it seems more logical to first fix the conflicts and then import so that the database isn't in a conflicting state. 

From what I remember this issue wasn't opened with a specific use case in mind, it was something I wasn't expecting when reading the source code. 

Feel free to close the issue if you think it doesn't make sense, or ask more questions! Cheers

leplatrem:
I think that makes sense. We'll probably have to rethink a bit the sync flow and the conflict API resolution for this though.

leplatrem:
Done in #553

PR #553, opened 2016-09-18
Cleanup/refactor sync code

### LINKED ISSUE #418: pushChanges is missing a case
The `then` block in https://github.com/Kinto/kinto.js/blob/d696e76dfba7f36a107f0d3a910cee142bbd39d7/src/collection.js#L854-L873 has a set of if-else-if blocks that aren't exhaustive. If we get something `resolved` which is not synced (`resolvedUnsynced.length != 0`), and our strategy isn't either `CLIENT_WINS` or `SERVER_WINS`, we fail to return anything at all. This caused me problems when I was trying to put together a fix for #414 and tried to mark a conflict as "resolved".

I'm not sure if we consider this case something that "can't happen". It seems like the only strategy this applies to is `MANUAL`, and we could assume that unsynced resolutions should never get to `pushChanges` in `MANUAL` because the conflicts stopped the sync. If this is how we want to hande this, I think we should put a default "else" case in `pushChanges` that throws a "can't happen" error, or at the very least add a comment explaining how someone got here and suggesting what they do to fix it.

leplatrem:
Now that we have switched to async/await it's a lot easier to read the code of the sync process and spot inconsistencies :)

I did that refactoring by keeping an eye on the integration tests. I'll have to change the specs/code of the unit tests now.

* [x] Refactor / added some comments
* [x] Update docstrings
* [ ] Document changes
* [x] Make unit tests pass

Fixes #418 #340

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -70,7 +70,13 @@ export class SyncResultObject {
     if (!Array.isArray(this[type])) {
       return;
     }
-    this[type] = this[type].concat(entries);
+    // Deduplicate entries by id. If the values don't have `id` attribute, just
+    // keep all.
+    const deduplicated = this[type].concat(entries).reduce((acc, cur) => {
+      const existing = acc.filter((r) => cur.id && r.id ? cur.id != r.id : true);
+      return existing.concat(cur);
+    }, []);
+    this[type] = deduplicated;
     this.ok = this.errors.length + this.conflicts.length === 0;
     return this;
   }
@@ -689,14 +695,10 @@ export default class Collection {
     syncResultObject.add("published", published);
 
     if (resolved.length > 0) {
-      // Resolved conflicts are merged with previous ones.
-      const resolvedIds = resolved.map((r) => r.id);
-      const mergedResolved = syncResultObject.resolved.filter((r) => resolvedIds.indexOf(r.id) < 0)
-                                                      .concat(resolved);
       syncResultObject
         .reset("conflicts")
         .reset("resolved")
-        .add("resolved", mergedResolved);
+        .add("resolved", resolved);
     }
     return syncResultObject;
   }
--- a/src/collection.js
+++ b/src/collection.js
@@ -627,18 +627,7 @@ export default class Collection {
    * occurred on the server since the last successful synchronization.
    * @return {Promise}
    */
-  async importChanges(syncResultObject, changeObject) {
-    const decodedChanges = await Promise.all(changeObject.changes.map(change => {
-      return this._decodeRecord("remote", change);
-    }));
-
-    syncResultObject.lastModified = changeObject.lastModified;
-
-    // No change, nothing to import.
-    if (decodedChanges.length === 0) {
-      return syncResultObject;
-    }
-
+  async importChanges(syncResultObject, decodedChanges) {
     // Retrieve records matching change ids.
     try {
       const imports = await this.db.execute(transaction => {
@@ -664,6 +653,29 @@ export default class Collection {
     }
 
     return syncResultObject;
+  }
+
+  async importPublications(toApplyLocally) {
+    if (toApplyLocally.length === 0) {
+      return [];
+    }
+    const toDeleteLocally = toApplyLocally.filter((r) => r.deleted);
+    const toUpdateLocally = toApplyLocally.filter((r) => !r.deleted);
+
+    const saved = await this.db.execute((transaction) => {
+      const updated = toUpdateLocally.map((record) => {
+        const synced = markSynced(record);
+        transaction.update(synced);
+        return {data: synced};
+      });
+      const deleted = toDeleteLocally.map((record) => {
+        transaction.delete(record.id);
+        // Amend result data with the deleted attribute set
+        return {data: {id: record.id, deleted: true}};
+      });
+      return updated.concat(deleted);
+    });
+    return saved;
   }
 
   /**
@@ -729,6 +741,7 @@ export default class Collection {
         }
       });
     });
+    this._lastModified = null;
     await this.db.saveLastModified(null);
     return unsynced.data.length;
   }
@@ -768,8 +781,12 @@ export default class Collection {
     if (!syncResultObject.ok) {
       return syncResultObject;
     }
+
+    const since = this.lastModified ? this.lastModified
+                                    : await this.db.getLastModified();
+
     options = {strategy: Collection.strategy.MANUAL,
-      lastModified: this.lastModified,
+      lastModified: since,
       headers: {},
       ...options};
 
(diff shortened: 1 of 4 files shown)


src/collection.js
```diff
@@ -884,59 +950,24 @@ export default class Collection {
...
-        return {data: synced};
-      });
-      const deleted = toDeleteLocally.map((record) => {
-        transaction.delete(record.id);
-        // Amend result data with the deleted attribute set
-        return {data: {id: record.id, deleted: true}};
-      });
-      return updated.concat(deleted);
-    });
+    await this.importPublications(syncResultObject, decoded, options.strategy);
```
leplatrem: note: add `if decoded.length > 0`

leplatrem:
@n1k0 @glasserc r? :)

leplatrem:
As a summary, here is what I did:
* `pushChanges()` is not in charge of `gatheringChanges()` anymore
* `importChanges()` is not in charge of decoding the records anymore
* created an `importPublications()` for `pushChanges()` results to mimic the code of `pullChanges()` with `importChanges()`
* I simplified the way `_handleConflicts()` is called, and changed it so that shares the same transaction as `importChanges()` and `importPublications()`
* the lastmodified option is now used explicitly in the second `pullChanges()` during sync instead of storing intermediate values as member variable during the sync process.

src/collection.js
```diff
@@ -1047,33 +1045,40 @@ export default class Collection {
...
     try {
-      const result = new SyncResultObject();
-      const pullResult = await this.pullChanges(client, result, options);
-      const pushResult = await this.pushChanges(client, pullResult, options);
-      // Avoid performing a last pull if nothing has been published.
-      if (pushResult.published.length === 0) {
-        syncResult = pushResult;
-      } else {
+      // Fetch last changes from the server.
+      await this.pullChanges(client, result, options);
```
glasserc: @leplatrem explained to me that although we used to have this sort of return-value chaining, which suggests immutability, things were also being modified in-place, so he took out the "fake" return values.

src/collection.js
```diff
@@ -652,33 +660,65 @@ export default class Collection {
...
+      return {published, resolved};
+    });
+
+    syncResultObject.add("published", published);
+
+    if (resolved.length > 0) {
+      // Resolved conflicts are merged with previous ones.
+      const resolvedIds = resolved.map((r) => r.id);
+      const mergedResolved = syncResultObject.resolved.filter((r) => resolvedIds.indexOf(r.id) < 0)
+                                                      .concat(resolved);
```
glasserc: Can we do better here? Maybe we should have a `Set` of resolved records, or at least a `Set` of resolved IDs. The indexOf call seems a little ugly to me. Maybe this complexity ought to be in the `SyncResultObject#add` method?

src/collection.js
```diff
@@ -667,6 +656,73 @@ export default class Collection {
...
+        return synced;
+      });
+      const deleted = toDeleteLocally.map((record) => {
+        transaction.delete(record.id);
+        // Amend result data with the deleted attribute set
+        return {id: record.id, deleted: true};
+      });
+      const published = updated.concat(deleted);
+      // Handle conflicts, if any
+      const resolved = this._handleConflicts(transaction, conflicts, strategy);
```
glasserc: It seems a little strange to have `importPublications`, which I assumed was just going to handle updating state about records that were successfully published, to also handle conflicts (records that weren't actually published). Can we have the caller manage the transaction and call both `importPublications` and `_handleConflicts`? Or maybe we can change the function name to something like `importPushResult`, and make `conflicts` be a parameter to the function?

src/collection.js
```diff
@@ -667,6 +656,73 @@ export default class Collection {
...
   /**
+   * Imports the responses of pushed changes into the local database.
+   * Basically it stores the timestamp assigned by the server into the local
+   * database.
+   * @param  {SyncResultObject} syncResultObject The sync result object.
+   * @param  {Array}            toApplyLocally   The list of changes to import in the local database.
+   * @param  {String}           strategy         The {@link Collection.strategy}.
+   * @return {Promise}
+   */
+  async importPublications(syncResultObject, toApplyLocally, strategy) {
```
n1k0: Same remark wrt default strategy value

src/collection.js
```diff
@@ -667,6 +656,73 @@ export default class Collection {
...
+    const toDeleteLocally = toApplyLocally.filter((r) => r.deleted);
+    const toUpdateLocally = toApplyLocally.filter((r) => !r.deleted);
+
+    const conflicts = syncResultObject.conflicts;
```
n1k0:
```js
const {conflicts} = syncResultObject; 
```

src/collection.js
```diff
@@ -768,8 +825,12 @@ export default class Collection {
     if (!syncResultObject.ok) {
       return syncResultObject;
     }
+
+    const since = this.lastModified ? this.lastModified
+                                    : await this.db.getLastModified();
+
     options = {strategy: Collection.strategy.MANUAL,
```
n1k0: Nit: while you're here, please format this object like everywhere else

src/collection.js
```diff
@@ -837,18 +907,18 @@ export default class Collection {
...
    * @param  {KintoClient.Collection} client           Kinto client Collection instance.
    * @param  {SyncResultObject}       syncResultObject The sync result object.
+   * @param  {Object}                 changes          The change object.
+   * @param  {Array}                  changes.toDelete The list of records to delete.
+   * @param  {Array}                  changes.toSync   The list of records to create/update.
    * @param  {Object}                 options          The options object.
    * @return {Promise}
    */
-  async pushChanges(client, syncResultObject, options={}) {
+  async pushChanges(client, {toDelete=[], toSync}, syncResultObject, options={}) {
```
n1k0: Now I'm wondering if this new second argument shouldn't be split as two distinct ones. Thoughts?

src/collection.js
```diff
@@ -1047,33 +1057,43 @@ export default class Collection {
...
-      const result = new SyncResultObject();
-      const pullResult = await this.pullChanges(client, result, options);
-      const pushResult = await this.pushChanges(client, pullResult, options);
-      // Avoid performing a last pull if nothing has been published.
-      if (pushResult.published.length === 0) {
-        syncResult = pushResult;
-      } else {
+      // Fetch last changes from the server.
+      await this.pullChanges(client, result, options);
+      const {lastModified} = result;
```
n1k0: Nit
```js
const {ok, lastModified, resolved, published} = result;
```

So you can directly reference these later.

src/collection.js
```diff
@@ -1047,33 +1064,43 @@ export default class Collection {
         new Error(`Server is asking clients to back off; retry in ${seconds}s or use the ignoreBackoff option.`));
     }
 
-    this._lastModified = await this.db.getLastModified();
```
glasserc: Why are we getting rid of this?

PR #562, opened 2016-09-22
Remove Firefox storage adapter from this repo

leplatrem:
Now that we don't transpile ES6 classes, @mozmark figured out that we can reach the base adapter from `Kinto.adapters.BaseAdapter`. Since the tests are in Gecko codebase, it makes a lot more sense to move the Firefox storage adapter code there.

We leave a wrapper here since Gecko relies on `Component.utils.import(...)` and not ES `import`s. Plus, the Gecko code base provides dependencies likes event emitters, uuids, ....

**Breaking change**: the exported symbol for Gecko is now the class `Kinto` and not `loadKinto` anymore.

Usage of kinto.js in Gecko becomes something like:

``` js
const { Kinto } = Cu.import("resource://services-common/kinto-offline-client.js");
const { FirefoxAdapter } = Cu.import("resource://services-common/kinto-storage-adapter.js");

let config = {
    remote: "http://server.com/v1",
    bucket: "app-bucket",
    adapter: FirefoxAdapter,
};
const kinto = new Kinto(config);
```

Here is what `kinto-storage-adapter.js` will be https://gist.github.com/leplatrem/7d60ff6c3b2b8e9c9785bfd4f2f70a8a

@n1k0 @glasserc r?


package.json
```diff
@@ -8,10 +8,10 @@
...
-    "dist-prod": "browserify --ignore process -s Kinto -g uglifyify -e src/index.js -o dist/kinto.min.js -t [ babelify --sourceMapRelative . ]",
-    "dist-fx": "BABEL_ENV=firefox browserify -s loadKinto --bare --ignore uuid --ignore kinto-http --ignore isomorphic-fetch -e fx-src/index.js -o temp.jsm -t [ babelify --sourceMapRelative . ] && mkdir -p dist && cp fx-src/jsm_prefix.js dist/moz-kinto-offline-client.js && echo \"\n/*\n * Version $npm_package_version - $(git rev-parse --short HEAD)\n */\n\" >> dist/moz-kinto-offline-client.js && cat temp.jsm >> dist/moz-kinto-offline-client.js && rm temp.jsm",
+    "dist-dev": "BABEL_ENV=development browserify --ignore process -s Kinto -d -e src/index.js -o dist/kinto.js -t [ babelify --sourceMapRelative . ]",
+    "dist-noshim": "BABEL_ENV=development browserify --ignore process -s Kinto -g uglifyify --ignore isomorphic-fetch --ignore babel-polyfill -e src/index.js -o dist/kinto.noshim.js -t [ babelify --sourceMapRelative . ]",
+    "dist-prod": "BABEL_ENV=development browserify --ignore process -s Kinto -g uglifyify -e src/index.js -o dist/kinto.min.js -t [ babelify --sourceMapRelative . ]",
```
n1k0: Hmm it's a bit awkward to use a `development` env to build production files. As far as I can see, the `development` env is actually the standard, non-firefox one, so we may just rename it as `production`, or `default`; what do you think?

Also, I'm surprised the `development` env is being used for the `noshim` target; one may expect this build to not embed any shim at all, but I suspect this doesn't even compile if we don't embed these, right?

n1k0: Also, I wonder about a dedicated `noshim` .babelrc environment, where we wouldn't enable any transform; would that work?

leplatrem:
@n1k0 please make the changes you want in .babelrc, that's the best I could do :)

glasserc:
Is there already a bug somewhere in Bugzilla for importing this code? Are you going to do that or is it going to fall on the hapless individual who is next to need to use an updated Kinto in Gecko?

Otherwise no strong opinions. The API you propose in the PR description seems fine to me.

leplatrem:
I was thinking of doing it having merged/tagged and didn't went ahead with bugzilla. I left as much details as possible here just in case ;)

leplatrem:
@n1k0 final feedback / r+ ?

glasserc:
I'm hoping not to merge this until https://bugzilla.mozilla.org/show_bug.cgi?id=1308515 goes through. There seems to be some kind of bug, which might be in Kinto.js or it might be in the code that uses it.

glasserc:
I've rebased this and would like to merge it. Changes to Gecko weren't too difficult to make. I don't understand `browserify` well enough to try to address @n1k0 's concerns, so I'm just going to merge this as soon as the tests pass and we can iterate later.

Issue #569, opened 2016-10-03
outgoing records aren't encrypted on CLIENT_WINS during outgoing conflict


glasserc:
In the [first call to pushChanges](https://github.com/Kinto/kinto.js/blob/1df13819077ae7b3662d5ee3c0730c364e86e5bc/src/collection.js#L1075-L1079), the `toSync` records are encoded because `gatherLocalChanges` encodes them. In the [second call to pushChanges](https://github.com/Kinto/kinto.js/blob/1df13819077ae7b3662d5ee3c0730c364e86e5bc/src/collection.js#L1081-L1085), the `toSync` records are the decoded conflicts. Since the server responds with the same data that was POSTed, this means that we double-decode records.

glasserc:
Something similar happens on incoming conflicts, but it isn't stored (see #568).

PR #570, opened 2016-10-03
Encode conflict resolutions during CLIENT_WINS (fixes: #569)

glasserc:
I don't remember if I didn't add this test the first time around because it wasn't working, or because I didn't think it could not be working, but add the test and add the fix.

@n1k0 @leplatrem r?


leplatrem:
r+

n1k0:
r+

Issue #567, opened 2016-09-29
I would like a Kinto#resetSyncStatus


glasserc:
This would operate on all collections in the local database.

n1k0:
In the case of IDB, that would mean keeping track of all collections db names handled by the library (solely relying on the name pattern seems fragile).

glasserc:
For my purposes, it would be good enough for it to just work in the Firefox storage.

glasserc:
With #562, it seems like this would be better suited to being as part of the FirefoxStorage class which will presumably be in Gecko, so I'll close this until someone else wants it.

leplatrem:
If it's feasible with indexeddb then no problem to expose that at the collection level!

leplatrem:
Note: https://bugzilla.mozilla.org/show_bug.cgi?id=934640

PR #571, opened 2016-10-03
FirefoxStorage: add resetSyncStatus

glasserc:
Unlike Collection#resetSyncStatus, which resets sync status for all records in a collection, this method resets sync status across all collections. Refs: #567.

This is incompatible with #562 so I will probably have to land it again in Gecko, but I thought I'd share it since it seems to work.

### DIFF
--- a/fx-src/FirefoxStorage.js
+++ b/fx-src/FirefoxStorage.js
@@ -84,8 +84,11 @@ const statements = {
 
   "importData": `
     REPLACE INTO collection_data (collection_name, record_id, record)
-      VALUES (:collection_name, :record_id, :record);`
+      VALUES (:collection_name, :record_id, :record);`,
 
+  "scanAllRecords": `SELECT * FROM collection_data;`,
+
+  "clearCollectionMetadata": `DELETE FROM collection_metadata;`,
 };
 
 const createStatements = ["createCollectionData",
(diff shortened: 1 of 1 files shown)


fx-src/FirefoxStorage.js
```diff
@@ -299,6 +302,40 @@ export default class FirefoxAdapter extends BaseAdapter {
...
+        const collection_name = row.getResultByName("collection_name");
+        if (record._status === "deleted") {
+          // Garbage collect deleted records.
+          promises.push(conn.execute(statements.deleteData, {collection_name, record_id}));
+        }
+        else {
+          const newRecord = Object.assign({}, record, {
+            _status: "created",
+            last_modified: undefined,
+          });
```
leplatrem: nitpick: we use object rest spread (maybe you left it like that 'cuz you know it'll be moved to gecko)

fx-src/FirefoxStorage.js
```diff
@@ -299,6 +302,40 @@ export default class FirefoxAdapter extends BaseAdapter {
         return result[0].getResultByName("last_modified");
       });
   }
+
+  /**
+   * Reset the sync status of every record and collection we have
+   * access to.
+   */
+  resetSyncStatus() {
```
leplatrem: How is it supposed to be used if it only exists on a specific adapter? Don't we want to expose a method at the root object?

fx-src/FirefoxStorage.js
glasserc: Oh, yeah, I got into the habit of doing this when hacking Gecko, sorry.

fx-src/FirefoxStorage.js
glasserc: I've been doing `collection.db.resetSyncStatus()`, because based on #567 it seemed like it wouldn't be as easy/beautiful to add something similar for the IndexedDB adapter.

leplatrem: Yes, you're right!
So far we can't enumerate the local databases with IndexedDB : https://bugzilla.mozilla.org/show_bug.cgi?id=934640

PR #549, opened 2016-09-16
Fix safari issue on IDB cursor with empty values

magopian:
Fixes #524

Issue #538, opened 2016-09-08
Switch to async/await


leplatrem:
Revamp PR #45 now that babelisation is more stable :)

PR #574, opened 2016-10-06
Prepare version 5.0.0

glasserc:
This release:

- Massive refactor of most methods from manually-handled promises to async/await (#538).
- Updates the documentation with a "NOP" schema example for use when you don't really need an `idSchema` (#533).
- Optimize `collection#list` by doing filtering before putting documents in a list rather than after (#543).
- Upgrades tests to Kinto 4.3 (#573).
- Fix for a puzzling bug involving IndexedDB use on Safari (#549).
- Fixes yet another bug in conflict resolution during `CLIENT_WINS` when using transformers (#570).
- Add a `FirefoxAdapter#resetSyncStatus` that resets sync status across all collections, rather than the one you happen to have gotten your hands on (#571). This is not possible as yet in IndexedDB, so it's only in `FirefoxAdapter` for now.

This might be the last release before we take `FirefoxStorage` out of this repo and move it to Gecko.

r? @leplatrem @Natim

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "kinto",
-  "version": "4.0.4",
+  "version": "5.0.0",
   "description": "An Offline-First JavaScript client for Kinto.",
   "main": "lib/index.js",
   "scripts": {


Natim:
> This might be the last release before we take FirefoxStorage out of this repo and move it to Gecko.

Final last words :D

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kinto",
-  "version": "4.0.4",
```
Natim: Did we really released a 4.0.4 version of Kinto.js? I didn't find it...

n1k0: yup https://github.com/Kinto/kinto.js/releases/tag/v4.0.4

n1k0:
fwiw, lgtm

package.json
Natim: /me was joking of course 👍

n1k0: /me grabs moar coffee

PR #576, opened 2016-10-12
[WIP] feature: deleteAll records (fix #556)

### LINKED ISSUE #556: Add deleteAll() method on collection
The available `clear()` method would wipeout the local data but won't delete records remotely.

We would need a method to delete all records locally and delete records remotely on sync...

This method on the collection class would set the «sync status» of its records to `deleted`, as if they would have been individually deleted with `deleteRecord()`
https://github.com/Kinto/kinto.js/blob/833a879b3aecd6c0abc407499eff29358b35a078/src/collection.js#L604

A test should be added here:
https://github.com/Kinto/kinto.js/blob/833a879b3aecd6c0abc407499eff29358b35a078/test/collection_test.js#L1169
That:
- creates some records in setup
- calls the new `deleteAll()`
- verifies that every records are still here but marked as `deleted` (ie. virtually deleted) using something like:
```js
articles.list(
            { filters: { _status: "deleted" } },
            { includeDeleted: true }
);
```

I think we may also want to check that the `deleteAll()` method emits an event.

enguerran:
- [X] add test
- [ ] pass the test
- [ ] refactor

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -581,6 +581,17 @@ export default class Collection {
   }
 
   /**
+   * Deletes all record from the local database.
+   * 
+   * @return {Object}
+   */
+  deleteAll(options={virtual: true}) {
+    return this.execute(transaction => {
+      return transaction.deleteAll(options);
+    });
+  }
+
+  /**
    * The same as {@link CollectionTransaction#deleteAny}, but wrapped
    * in its own transaction.
    *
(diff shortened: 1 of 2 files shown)


n1k0:
What's the difference with `clear()`?

leplatrem:
`clear()` would wipe everything out locally. `deleteAll()` would replicate the deletions remotely.

enguerran:
It is linked to #556

`clear()` deletes the whole local database and wait for a `sync()` to repopulate it.

`deleteAll()` marks every records as `mark_as_deleted` and wait for a `sync()` to deleteAll on the server.

n1k0:
Ok, makes sense.

PR #584, opened 2016-10-25
Use constraints for Kinto installation

leplatrem:
We can do this, or wait for 4.3.2. 

@Natim @n1k0 what do you think ?


n1k0:
I'd rather wait to a released version, or add a new server branch target to the matrix

Natim:
I don't understand why you would wait for this change. It seems quite sane to me.

n1k0:
ok then

Issue #588, opened 2016-11-02
Can we share an Adapter across Collections?


glasserc:
In Gecko, managing simultaneous Sqlite database connections is unwise, and trying to close and open them when they might be in use can lead to contention and NS_ERROR_STORAGE_BUSY. But the code that I inherited does exactly this, with client code implicitly creating new `Collection`s, each of which has to be `open()`ed and `close()`d at appropriate times. This would be a lot better if I could create and initialize just one, and tear it down on application shutdown, but it's impossible to pass an existing `adapter` when creating a `Collection`. The best I can do is pass in the adapter "class" (although it's still called `adapter`).

I can't figure out if the current situation, where each `Collection` has its own `Adapter`, is desirable or just an accident of history. If it's not desirable, we could change it for everyone. If it is desirable except in some situations, we could allow client code to explicitly share Adapters when it's correct to do so.

Otherwise, I guess as a workaround, I can create a SharedSqliteAdapter that wraps FirefoxAdapter but with a global `connection` handle instead of having a separate one for each `Collection`.

glasserc:
I think I can manage this on my own by changing the FirefoxStorage class to take an existing connection as one of its options (which are passed as `adapterOptions`). This doesn't require any changes to the architecture of Kinto, which currently creates a new adapter for each collection.

Why do we have an adapter for each collection? Per conversation with @n1k0, the answer is probably "historical reasons". In general, we'd expect a Kinto application to have one "database" with multiple "tables", one for each collection; the current architecture sort of favors having one database per collection instead. Maybe we should change this structure, but it would be a major version and could break client code. We should be sure to be careful to see how FirefoxStorage in Gecko uses this.

One reason we have this structure is because IDB only gives you a very small quota per database, and we want to allow users to spread their collections out to try to work around that quota. Maybe if we fix #526, then we can stick everything in one giant IDB database and lay out our table similarly to how we do it in FirefoxStorage.

glasserc:
[...]

PR #589, opened 2016-11-03
FirefoxStorage: Allow connection lifetimes to be managed externally

glasserc:
This is hopefully the last change I'll have to make before merging #562.

This addresses the parts of #588 that I need right now.

r? @n1k0

### DIFF
--- a/fx-src/FirefoxStorage.js
+++ b/fx-src/FirefoxStorage.js
@@ -108,8 +108,9 @@ const currentSchemaVersion = 1;
 export default class FirefoxAdapter extends BaseAdapter {
   constructor(collection, options={}) {
     super();
+    const {sqliteHandle=null} = options;
     this.collection = collection;
-    this._connection = null;
+    this._connection = sqliteHandle;
     this._options = options;
   }
 
@@ -144,9 +145,9 @@ export default class FirefoxAdapter extends BaseAdapter {
   open() {
     const self = this;
     return Task.spawn(function* (){
-      const path = self._options.path || SQLITE_PATH;
-      const opts = { path, sharedMemoryCache: false };
       if (!self._connection) {
+        const path = self._options.path || SQLITE_PATH;
+        const opts = { path, sharedMemoryCache: false };
         self._connection = yield Sqlite.openConnection(opts).then(self._init);
       }
     });
--- a/fx-src/FirefoxStorage.js
+++ b/fx-src/FirefoxStorage.js
@@ -114,6 +114,9 @@ export default class FirefoxAdapter extends BaseAdapter {
     this._options = options;
   }
 
+  // We need to be capable of calling this from "outside" the adapter
+  // so that someone can initialize a connection and pass it to us in
+  // adapterOptions.
   static _init(connection) {
     return Task.spawn(function* () {
       yield connection.executeTransaction(function* doSetup() {
--- a/fx-src/FirefoxStorage.js
+++ b/fx-src/FirefoxStorage.js
@@ -114,7 +114,7 @@ export default class FirefoxAdapter extends BaseAdapter {
     this._options = options;
   }
 
-  _init(connection) {
+  static _init(connection) {
     return Task.spawn(function* () {
       yield connection.executeTransaction(function* doSetup() {
         const schema = yield connection.getSchemaVersion();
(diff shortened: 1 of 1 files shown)


fx-src/FirefoxStorage.js
```diff
@@ -108,12 +108,13 @@ const currentSchemaVersion = 1;
...
     super();
+    const {sqliteHandle=null} = options;
     this.collection = collection;
-    this._connection = null;
+    this._connection = sqliteHandle;
     this._options = options;
   }
 
-  _init(connection) {
+  static _init(connection) {
```
leplatrem: Maybe add a comment to explicit what does the `static` solves?

glasserc:
No, nothing is incompatible with #562. I just pass an `adapterOptions` argument, right below `adapter`.

fx-src/FirefoxStorage.js
```diff
@@ -108,12 +108,16 @@ const currentSchemaVersion = 1;
...
-    this._connection = null;
+    this._connection = sqliteHandle;
     this._options = options;
   }
 
-  _init(connection) {
+  // We need to be capable of calling this from "outside" the adapter
+  // so that someone can initialize a connection and pass it to us in
+  // adapterOptions.
+  static _init(connection) {
```
n1k0: If this is intended to be called from outside, it's part of the public API and shouldn't be denoted as private with an underscore in the name. Also, have you considered moving this as a separate function? Maybe the FFx module export system isn't flexible enough to make usage painless?

fx-src/FirefoxStorage.js
```diff
@@ -144,10 +148,10 @@ export default class FirefoxAdapter extends BaseAdapter {
   open() {
     const self = this;
     return Task.spawn(function* (){
-      const path = self._options.path || SQLITE_PATH;
-      const opts = { path, sharedMemoryCache: false };
       if (!self._connection) {
-        self._connection = yield Sqlite.openConnection(opts).then(self._init);
+        const path = self._options.path || SQLITE_PATH;
+        const opts = { path, sharedMemoryCache: false };
+        self._connection = yield Sqlite.openConnection(opts).then(FirefoxAdapter._init);
```
n1k0: I really feel like we shouldn't create default connection from within the adapter; feels magic and fragile. Would it be hard to update other consummers so they always initiate their own connection?

n1k0:
Woops, missed it's been merged already. Questions are still valid to me though.

fx-src/FirefoxStorage.js
glasserc: Yeah, I was hesitant to even try fighting the build system. When @leplatrem was in New York, I saw him struggling with #562 and how there was only allowed to be one exported symbol. I was in a hurry at the time I was writing this, and I knew that #562 would probably follow this -- when it does, all of this code becomes part of the Gecko codebase. Because of that, I was comfortable considering `_init` as not part of the public API from the perspective of Gecko.

glasserc: I'd be happier if `_init` had a clearer name that made it obvious that callers needed to use it!

fx-src/FirefoxStorage.js
glasserc: Personally, I like "simple things should be simple; hard things should be possible" as a principle of API design, so in my perfect world, consumers wouldn't have to manage connections unless they had a reason to (as I do). That said, the existing API is not simple even in simple cases, since the consumer has to explicitly `open()` and `close()` the connections (because Promises). So maybe I will update the consumers.

glasserc: I'm reminded of the old Rusty Russell blog posts about API design ([intro](http://ozlabs.org/~rusty/index.cgi/tech/2008-03-18.html), [part 1](http://ozlabs.org/~rusty/index.cgi/tech/2008-03-30.html), [part 2](http://ozlabs.org/~rusty/index.cgi/tech/2008-04-01.html)). I'm nervous about this API being easy to misuse. The original API *requires* you to call `open` or else it always breaks. The new one allows you to pass in an uninitialized Sqlite handle, and it *might* break depending on the previous state of the DB. I'd like the "default" interface to be safe, even if some people with special needs can avoid it if they need to. Maybe setting an attribute on a SqliteConnection that says it's been initialized would be a good idea?

glasserc:
To close the loop, I addressed these concerns in https://bugzilla.mozilla.org/show_bug.cgi?id=1319884.

PR #596, opened 2016-11-15
Prepare version 6.0.0

glasserc:
This merges #562, which is a major version change because it removes the
Firefox storage adapter. This code is specific to Gecko and will move to
that repository.


glasserc:
Not waiting for review because velocity.

PR #620, opened 2017-01-12
Add sync events

leplatrem:
That is a trivial addition.

I think it is useful for pluggabilty and flexibility, but I feel I could be convinced otherwise :)

For example:

```js
const kinto = new Kinto();
kinto.events.on("sync:success", function(result) {
  telemetry.count("success");
});
kinto.events.on("sync:error", function(result) {
  telemetry.count("error");
});
collection.sync();
```

Versus:
```js
const kinto = new Kinto();

collection.sync()
  .then(() => telemetry.count("success"))
  .catch(() => telemetry.count("error"));
```

Thoughts ?

### DIFF
--- a/test/collection_test.js
+++ b/test/collection_test.js
@@ -2382,6 +2382,40 @@ describe("Collection", () => {
             expect(onerror.called).eql(true);
           });
       });
+
+      it("should send an error event", () => {
+        articles.pushChanges.throws(new Error("boom"));
+        return articles.sync()
+          .catch(() => {
+            expect(onsuccess.called).eql(false);
+            expect(onerror.called).eql(true);
+          });
+      });
+
+      it("should provide success details about sync", () => {
+        return articles.sync()
+          .then(() => {
+            const data = onsuccess.firstCall.args[0];
+            expect(data).to.have.property("result");
+            expect(data).to.have.property("remote");
+            expect(data).to.have.property("bucket");
+            expect(data).to.have.property("collection");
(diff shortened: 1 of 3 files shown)


n1k0:
:+1:

test/collection_test.js
```diff
@@ -2350,6 +2350,29 @@ describe("Collection", () => {
...
+        sandbox.stub(articles.db, "getLastModified").returns(Promise.resolve({}));
+        sandbox.stub(articles, "pullChanges");
+        sandbox.stub(articles, "pushChanges");
+      });
+
+      it("should send a success event", () => {
+        const callback = sinon.spy();
+        articles.events.on("sync:success", callback);
+        return articles.sync()
+          .then(() => expect(callback.called).eql(true));
```
glasserc: We should probably check that `sync:error` *wasn't* called here.

test/collection_test.js
```diff
@@ -2350,6 +2350,29 @@ describe("Collection", () => {
...
+      });
+
+      it("should send an error event", () => {
+        const callback = sinon.spy();
+        articles.events.on("sync:error", callback);
+        articles.pushChanges.throws(new Error("boom"));
+        return articles.sync()
+          .catch(() => expect(callback.called).eql(true));
```
glasserc: Same here.

leplatrem:
> spooky-action-at-a-distance

Completely agree. We'll be careful when plugging with Telemetry

PR #621, opened 2017-01-13
Retry once by default

leplatrem:


test/collection_test.js
```diff
@@ -2351,6 +2361,38 @@ describe("Collection", () => {
...
     });
 
+    describe("Retry", () => {
+      let fetch;
+
+      beforeEach(() => {
+        // Disable stubbing of kinto-http, stub low-level fetch instead.
+        sandbox.restore();
+        fetch = sandbox.stub(global, "fetch");
+        sandbox.stub(global, "setTimeout", (fn) => setImmediate(fn));
```
n1k0: Nit: it may be worth inserting a comment explaining why we're doing this here (same as in kinti-http tests)

PR #640, opened 2017-01-26
Encode local deletes as remote keeps

### LINKED ISSUE #638: Support remotetransforming deletes into record updates
When a record is deleted locally, kinto.js sends that local deletion as a DELETE to the remote server. However, if you're using a RemoteTransformer to encrypt the records, you probably want to encrypt deletes as well so that you don't leak information about the fact that a delete occurred (or to store information that a normal tombstone wouldn't). Right now we don't seem to support this; records that end up in `toDelete` get the transformer run over them, but whatever the transformer returns is still passed to `batch.deleteRecord`. It would be better if we could turn records from `toDelete` to `toSync`.

glasserc:
Try to fix #638.

A remote transformer normally maps local deletes to remote deletes and local updates/creates ("keeps") to remote keeps. However, we can also imagine remote transformers that map "across" these categories -- that is, mapping local deletes to remote keeps, or local keeps as remote deletes. This PR implements the first of these categories as part of fixing https://bugzilla.mozilla.org/show_bug.cgi?id=1333810, https://bugzilla.mozilla.org/show_bug.cgi?id=1333814, and https://bugzilla.mozilla.org/show_bug.cgi?id=1333818.

Note that currently mapping local keeps to remote deletes isn't going to work very well. Remote deletes on the Kinto server become tombstones, which don't have a lot of information, and probably won't have enough information to recreate the local keep. This means that other clients are S.O.L., which means that this sync is decidedly asymmetric, which is somewhat unfortunate. Additionally, the current implementation of the syncing logic relies on the response from the server, which it takes as the "new" version of the record. Because this new record is a tombstone, it will of necessity destroy the local record as well.

- [x] Update documentation with an example of this kind of transformer as well as its requirements
- [x] Test that this actually solves the problem I'm trying to solve in Gecko
- [x] Add a test case that proves that this works even if fetching a record "cold" (i.e. without a matching local record)

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -894,6 +894,48 @@ coll.create({title: "foo"}).then(_ => coll.sync())
 >
 > This mechanism is especially useful for implementing a cryptographic layer, to ensure remote data are stored in a secure fashion. Kinto.js will provide one in a near future.
 
+Normally, a record that is deleted locally will be deleted remotely, and a record that is not deleted locally will not be deleted remotely. However, with a remote transformer, it's possible to change this. The records given to `encode()` have a `_status` field which represents their local status (`synced`, `created`, `updated`, or `deleted`); by turning a `deleted` into a `created` or `updated`, or by turning a `created` or `updated` into a `deleted`, you can control what happens to the remote record. Similarly, the records given to `decode()` have a `deleted` field, which is true if the record was deleted on the remote end; by turning `true` to `false` or `false` to `true`, you can control what happens to the local version of this record.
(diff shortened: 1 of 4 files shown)


glasserc:
This isn't done yet, and I haven't actually tested that it does what I need it to, but I think it's approaching a state where it could be reviewed. r? @Natim

glasserc:
OK, this wasn't so bad. I think this PR is ready to be reviewed and maybe merged.

PR #641, opened 2017-01-27
Prepare release 8.0.0

glasserc:
This release:
- updates a bunch of dependencies (#625, #626, #627, #628, #629, #630, #633, #634, #635, #636)
- refactors some internals of the `sync()` method to allow users to transform local deletes into remote keeps (#640)

This is a major version because it changes one aspect of how remote transformers work. Previously, a remote transformer could return anything for a deleted record, and that record would still be deleted. Now, if a transformer changes a record's `_status`, it will be respected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "kinto",
-  "version": "7.1.0",
+  "version": "8.0.0",
   "description": "An Offline-First JavaScript client for Kinto.",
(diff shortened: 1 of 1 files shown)


glasserc:
After thinking about this more, I think this should be a major version, since it changes some aspects of the RemoteTransformer contract.

Issue #632, opened 2017-01-20
Pass record to custom ID schema


blarghmatey:
I am beginning work on an application where I would like to be able to generate the ID for a record that is computed from other attributes of the record. Looking at the custom ID schema capabilities there doesn't appear to be any way of getting a handle on the record that the ID is generated for inside the `generate` function.

For example, every record will have a `url` attribute and I would like to compute the ID based on the value of that attribute to ensure that there is never a duplicate value entered for that attribute.

Is something like this possible and if not what would I need to do to enable it?

leplatrem:
That sounds like a reasonable use case to me! And would invite you to draft a PR ;)

@glasserc @Natim what do you think?

Natim:
Yes I agree, in the past we did it by forcing the computation of the id before save the record rather than using `generate` for that but it might make sense to do it there. However, what happens when you update the record with a new `url`?

blarghmatey:
That's a valid point that I did not yet consider. I am still very early in the design/execution of this project, so I'm sure there are some cases that I'm not yet considering.

In the event that the URL changes then I think that the behavior should be to delete the old record and create a new one. Alternatively, if there is a different/better way to ensure uniqueness of a field in a record then it might be worth pursuing that instead of/in addition to the id generation.

glasserc:
I agree with @Natim that a record's ID "changing" seems like a bad thing, and I think detecting something like that would require a certain amount of work. (As of this writing, we only ever call `generate()` when a record doesn't have an ID. When would we check if the record's ID needed to be updated?) I guess I'm not opposed to supporting this if we can manage it, but Kinto doesn't really provide any mechanisms for guaranteeing uniqueness apart from ID, and I am hesitant to commit to anything beyond that. Because we don't offer any kind of "move" operation, conflicts become a bit uglier. In particular, if one client updates a record, and the other "moves" it, I'm worried about the possibility of keeping both IDs.

blarghmatey:
So, I've started digging into this and I'm not sure how viable this option will be.

I can easily pass the record to the `generate` function, the problem comes up because of also having to pass the record to the `validate` function which gets called on every invocation of `execute`. For some of the invocations the record is available to pass through, but for functions such as `get`, `getAny`, or `delete` the only value available is the id.

glasserc:
Yes, the `validate` method is meant to just validate the syntax of the `id` field itself. You won't be able to validate that the ID matches the record. I guess your `validate` method would just have to return `true`..

[...]

PR #642, opened 2017-01-31
Added the ability to optionally pass the record to the generate function

blarghmatey:
In order to support generating IDs that are based on the contents of a record, added the ability to optionally pass the record to the `generate` function of a custom ID schema.

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,7 +132,7 @@
     "coveralls": "^2.11.15",
     "esdoc": "^0.5.2",
     "esdoc-importpath-plugin": "0.1.4",
-    "eslint": "3.19.0",
+    "eslint": "4.1.0",
     "estraverse-fb": "^1.3.1",
     "fake-indexeddb": "2.0.3",
     "gh-pages": "^1.0.0",


test/collection_test.js
```diff
@@ -62,6 +62,21 @@ describe("Collection", () => {
...
   }
 
+  function createKeyListIdSchema() {
+    return {
+      generate(record) {
+        let id = [];
+        for (let key in record) {
+          id.push(key);
+        }
+        return id;
```
leplatrem: I personnally find this is a bit weird... the usage example you had in the docs would make a lot more sense IMO :)

src/collection.js
```diff
@@ -464,9 +464,9 @@ export default class Collection {
       return reject("Extraneous Id; can't create a record having one set.");
     }
     const newRecord = {...record,
-      id:      options.synced ||
-                   options.useRecordId ? record.id :
-                                     this.idSchema.generate(),
+      id: options.synced ||
+                       options.useRecordId ? record.id :
+                       this.idSchema.generate(record),
```
leplatrem: I'd rather use an intermediary variable here, the resulting indentation is not very pleasant (not your fault I know ;))

blarghmatey:
Thanks for the feedback @leplatrem, I'll get those comments addressed and push an update soon.

test/collection_test.js
blarghmatey: Looking at this again I remembered why I didn't use the example from the docs. When I tried to run the tests from the CLI they failed because Node doesn't have the `btoa` and `atob` function available. When I tried using the recommendation from [here](http://stackoverflow.com/a/23097961) it threw a different error, so I went with the test that you see here. I'll take another crack at building a more useful example that isn't quite so strange.

src/collection.js
blarghmatey: When you mention using an intermediary variable, are you referring to the `id` or the `record` being passed to the generate command?

glasserc: I think @leplatrem means a variable representing the boolean condition, plus perhaps one for the entire `id` field if you can't get it to fit on one line.

test/collection_test.js
```diff
@@ -392,6 +407,16 @@ describe("Collection", () => {
...
     });
 
+    it("should accept a record for the 'generate' function", () => {
+      articles = testCollection({
+        idSchema: createKeyListIdSchema()
+      });
+
+      return articles.create(article)
+        .then(result => result.data.id)
+        .should.eventually.be.a("Array");
```
leplatrem: From what I read in the generate function, it comes out as a string

gabisurita:
What's the status of this PR? @leplatrem @glasserc

glasserc:
Oops, sorry, I didn't notice that another commit was pushed that addresses Mat's feedback. It needs a rebase to land now. If someone wants to do that, go ahead, otherwise I'll try to sort it out.

glasserc:
Refs: #632.

test/collection_test.js
```diff
@@ -64,6 +64,21 @@ describe("Collection", () => {
...
   }
 
+  function createKeyListIdSchema() {
+    return {
+      generate(record) {
+        let id = [];
+        for (let key in record) {
+          id.push(key);
+        }
+        return id.join(",");
```
n1k0: Two things. First, this could be rewritten as:

```js
return Object.keys(record).join(",")
```

Also, we should ensure the keys are always ordered the same way to ensure consistency and predictability, so I'd suggest:

```js
return Object.keys(record).sort().join(",")
```

test/collection_test.js
```diff
@@ -64,6 +64,21 @@ describe("Collection", () => {
...
+      },
+      validate(id) {
+        return id !== null && Array.isArray(id.split(","));
```
n1k0: Several things here:

`id != null` would check for both `null` and `undefined`, which may be a worthy addition.

Also a split string will always be an array (`"".split(",") // [""]`) so you may want to update this check as well. 

How about just checking for a non-empty string?

test/collection_test.js
```diff
@@ -456,6 +471,16 @@ describe("Collection", () => {
...
     });
 
+    it("should accept a record for the 'generate' function", () => {
+      articles = testCollection({
+        idSchema: createKeyListIdSchema()
+      });
+
+      return articles.create(article)
+        .then(result => result.data.id.split(","))
+        .should.eventually.be.a("Array");
```
n1k0: I'd rather directly check for the resulting string. Though I suspect you did this to avoid having intermittent failures because object keys order is never guaranteed; if so, we should probably ensure always sorting the keys before joining.

PR #643, opened 2017-02-03
Demo : Use unpkg latest release link

ioO:
The kinto.js script is not provided with demo source. As documentation tutorial use unpkg service I use this ressource for src attribute.


demo/index.html
```diff
@@ -40,7 +40,7 @@
       </li>
     </template>
   </div>
-  <script src="kinto.js"></script>
+  <script src="https://unpkg.com/kinto@8.0.0/dist/kinto.min.js"></script>
```
leplatrem: We could even remove the `8.0.0` part ?

ioO: Also we could, because it's redirected to latest kinto version.

n1k0:
Yeah the demo was intended to be runnable using `npm start` so you could hack on the code and test the resulting build against the demo. But I'm okay siplifying the life of newcomers. Your call.

ioO:
There is limitation with npm command in the case of demo. I did not use git command to clone the whole kinot.js repo and get the demo. I red the tutorial first and wrote the code to understand. At the end I did a wget on index.html because I played a lot to learn and something was broken. I had no results and I quickly found out the network error for kinto.js file.

leplatrem:
I'm still not entirely convinced, maybe we should improve the tutorial instead.
In any case, the `builddemo`, `demo`, `publishdemo` commands in `package.json` must be updated/removed if we follow this strategy (and some docs probably have to be updated too).

ioO:
I understand and I also got the idea. But instead of not doing a quick fix that is a solution for the actual tutorial and wait that another work resolves this issue - which is a great solution but ask more time - I submit this pull request.
I choose this way because when I do a tutorial, I never clone a repo. I write code and do not want to manage git command, I focus on learning. As I am not really good at npm and webpack, I use my own way `python -m http.server 8000`. I think it is a good practice to keep something easy to learn, even if it is not the art of coding for a real project. 
I thought about doing a pull request with tutorial change only, but it was more work to do. So if like me you have not enough time, I suggest you accept this pull request. Since I play with kinto I have more work to do and also I do not feel comfortable at this time to contribute to code and tutorial.
You can reject this pull request and follow your strategy, I understand and I won't be disappointed ;)

PR #652, opened 2017-02-10
[WIP] Reducing library size

### LINKED ISSUE #650: 175 kb size?
I added `kinto.js` to my project (see also #649) as a `npm` dependency and saw that it increased my final bundle file size about 175 kb in size 😳 

Is there anything I can do to reduce it's size? It seems a little bit too large in my eyes compared to other libraries / frameworks.

screendriver:
Trying to resolve #650


screendriver:
Short intermediate state: with my simple changes here I reduced `kinto.js` (in a production / minified build) from `199 kb` to `111 kb` 😎

.babelrc
```diff
@@ -1,34 +1,6 @@
...
-        "transform-object-rest-spread",
-        "transform-async-to-generator"
-      ]
-    },
-    "firefox": {
-      "plugins": [
-        "transform-es2015-modules-commonjs",
-        "transform-object-rest-spread",
-        "transform-async-to-generator"
-      ]
```
leplatrem: we'll have to keep the `firefox` target, because we ship this library in the browser :) https://dxr.mozilla.org/mozilla-central/source/services/common/kinto-offline-client.js

src/KintoBase.js
```diff
@@ -1,7 +1,5 @@
-"use strict";
-
```
leplatrem: out of curiosity why did you remove `use strict`? is it useless because of babel? #noob ;)

.babelrc
screendriver: Ah I see.

1. `transform-es2015-modules-commonjs` is already included in the preset `es2015` (which is included in `latest`.  I added `latest` 😉 )
2. `transform-object-rest-spread` is included in `stage-3` (I included that as well)
3. same for `transform-async-to-generator` 😉 

So this is all redundant and can be removed ☺️ 

One thing: you ship this library in the browser. Ok. This is for people that reference `kinto.js` directly in their `index.html` and work with the global Kinto object. This approach is a little bit outdated, but ok, I understand that this is necessary for some people out there. For those people that **only** uses Kinto it is necessary to include the `babel-polyfill` and **all** stuff that it's necessary that it can run in the browser. For all other developers that use modern build tools like webpack and so on this workflow is not recommended.

I don't have a solution for this at the moment... Maybe we need a second build? One build for the **big** browser solution that runs out of the box and one small build for modern build tools? 🤔

src/KintoBase.js
screendriver: > is it useless because of babel? 

yep 😉

screendriver:
At my research I figured out that [kinto-http.js](https://github.com/Kinto/kinto-http.js) is also very huge in size (although it has almost no dependencies). Maybe I should start from the bottom: first "fix" `kinto-http.js` and after that `kinto.js`

screendriver:
Another thing: I see that you are using `browserify` to build the stuff. Could I replace it with webpack? Webpack is more mature in the meantime and has, at least in my experience, the biggest community at the moment out there regarding build tools.

Natim:
Yes we are using webpack for other things so I guess you could use webpack. However keep in mind that we need to keep the firefox target at least until kinto-http.rs takes over.

screendriver:
> However keep in mind that we need to keep the firefox target

Of course. I try to make it fully transparent and just reduce the size. Functionality will be 100% as before 😎

screendriver:
FYI: if you run `npm run analyze-bundle` you can see the bundle content as convenient interactive zoomable treemap :sunglasses:

screendriver:
I think I need a little bit help here from you guys:

1. [kinto-http.js](https://github.com/Kinto/kinto-http.js) has the same size problem as `kinto.js` (you can see it when you run `npm run analyze-bundle`). As long as `kinto-http.js` stays big in it's size, `kinto.js` can't be reduced to a maximum level. So we have to optimise `kinto-http.js` as well (I would do that if you want)
2. I don't really understand the npm script `dist-fx`. Can anyone help me with this or explain what it really does?
3. how do you test that all bundles work as expected in all browsers? With your unit tests? That brings me to my last point
4. 330 unit tests are running and green. 5 of them fails and I don't get it why (`before all` hook failed in integration tests with `Uncaught Error: spawn pserve ENOENT`)

n1k0:
>[kinto-http.js](https://github.com/Kinto/kinto-http.js) has the same size problem as `kinto.js` (you can see it when you run `npm run analyze-bundle`). As long as `kinto-http.js` stays big in it's size, `kinto.js` can't be reduced to a maximum level. So we have to optimise `kinto-http.js` as well (I would do that if you want)

We'd love having you working on this!

>I don't really understand the npm script `dist-fx`. Can anyone help me with this or explain what it really does?

This command builds the JSM module for Gecko use, and saves it into `dist/moz-kinto-http-client.js` so we can move it to the gecko tree and consume the API from there.

>how do you test that all bundles work as expected in all browsers? With your unit tests? That brings me to my last point

Yes. We know it's not enough, but we've been burnt in the past with functional testing and don't have much time to dedicate to this task.

>330 unit tests are running and green. 5 of them fails and I don't get it why (`before all` hook failed in integration tests with `Uncaught Error: spawn pserve ENOENT`)

That's because you need to setup a local Kinto server to run the integration tests against. Setup is documented [here](https://kintojs.readthedocs.io/en/latest/contributing/#tests).

leplatrem:
> That's because you need to setup a local Kinto server to run the integration tests against. Setup is documented here.

Or just have a look at what is done in `.travis.yml` it is rather easy

screendriver:
Thank you for your help. Quick update: I setup a test server as described [here](https://kintojs.readthedocs.io/en/latest/contributing/#tests) and run the tests again: `505 passing` 😎 Cool! So I didn't crash anything while the switch to webpack.

So the last things are:

- `dist-fx`
- verify that all builds run in the browser (for that I believe I need your help)
- optimize `kinto-http.js`

screendriver:
Now I tried to fix `dist-fx`.

Could anyone checkout this branch and verify that everything is still running as expected at the moment?

leplatrem:
Thanks for your motivation on this!

I looked at the generated `dist-fx`. Unfortunately, it contains a lot of transpiling :/ 
Ideally we'd like to keep the Firefox export as close as possible to vanilla ES6 (classes etc.). The only things that have to be transpiled are:
* imports 
* object rest spread (`{...options, a:  1}`, until https://bugzilla.mozilla.org/show_bug.cgi?id=1339395 is fixed)

We used to transpile async/await because browserify didn't support them, but they are supported by Firefox so if we can avoid their transpilation, it'd be better.

screendriver:
So that was the reason why I [asked](#issuecomment-279303036) what the `dist-fx` task is doing ;-)

I changed `.babelrc` for the firefox build.

And what about the no-shim build? I don't really understand 100% for what it is for. I believe I've did it wrong as well 😕

Natim:
`no-shim` is a build with no babel polyfill at all.

screendriver:
I'm asking because at the moment in `no-shim` I transpile everything as usual and **only** exclude any polyfill related (in fact I deleted the whole polyfill and included the dynamically evaluated `babel-runtime`)

I'm just a little but unsure that everything is working as expected at the moment. Unit tests are fine but I don't know if the bundled output is ok at the moment and that everything works as expected.

glasserc:
I ran the dist-fx build and copied it into Firefox and ran all the tests and everything seems to work fine, although of course the output looks very different!

.gitignore
```diff
@@ -6,3 +6,6 @@
 /esdoc
 /node_modules
 /demo/kinto*
+/.vscode
+.DS_Store
```
n1k0: I think these should be added to users global `.gitconfig` instead (eg. `~/.gitconfig`). We're not gonna ignore all possible kinds of stuff like this :)

src/KintoBase.js
```diff
@@ -1,7 +1,5 @@
...
-import Collection from "./collection";
 import BaseAdapter from "./adapters/base";
+import Collection from "./collection";
```
n1k0: Out of curiosity, is this matter of ordering imports alphabetically?

webpack.config.js
```diff
@@ -0,0 +1,63 @@
...
+      test: /\.js$/,
+      exclude: /node_modules/,
+      loader: "babel-loader"
+    }]
+  },
+  devtool: "source-map"
+};
+
+switch (process.env.npm_lifecycle_event) {
+case "dist-prod":
```
n1k0: Let's indent cases with 2 spaces

glasserc:
Isn't the second `collection.js` from `kinto-http.js`?

n1k0:
> Isn't the second collection.js from kinto-http.js?

Ah, hence the `lib` above it, right.

screendriver:
> Now we're not shipping the babel polyfill anymore, we should update the docs to warn users about that, and explain how to use it if they want to (or alternative solutions)

That's not necessary. The babel polyfill in fact will not be included, but I added the babel plugin `transform-runtime`. This one analyzes every import and dynamically add only the polyfill that are **really** needed. So the bundle gets significant but includes still everything needed for older browsers.

> Also I'm seeing polyfills for promises and symbols being part of that chart; hope we're not bundling them as target browsers implement them already :)

Unfortunately that's not true 😕 There are still "modern" Browser out there that does not support ES2015 features. For instance Microsoft IE11 [does not support](http://caniuse.com/#feat=promises) Promises.

> Isn't the second collection.js from kinto-http.js?

Does `collection.js` the same in both packages? If yes it looks like that it should encapsulated an independent package for me to reduce duplicates.

webpack.config.js
screendriver: I only did that because of your lint rules. If I make 4 spaces my editor complains about the linting rules.

src/KintoBase.js
screendriver: Oh, sorry. This happens automatically in my editor. In Atom and Visual Studio Code I have an auto import-sort plugin enabled.

.gitignore
screendriver: Hmm 🤔  I believe that most people (like me for example) have a global `.gitconfig` and only a per-project config. I can remove that if you want?

n1k0: Yes please :)

webpack.config.js
n1k0: Hmm that's strange as linting is part of our CI flow. Does your plugin parses the .eslintrc file? Also, it should be indented with two spaces.

n1k0:
> Does collection.js the same in both packages?

Nope they're not.

screendriver:
So I'm almost done. The biggest part remaining is to optimize `kinto-http.js`. I will start with that soon (also with a `WIP` pull request). When that is finished we can release `kinto-http.js` first and after that `kinto.js` 😎 🎉

n1k0:
Thank you for your work on this, cheers :+1:

screendriver:
Your welcome :relaxed:

dstaley:
Closing since v13 dropped the bundle size down to ~15 KB.

PR #660, opened 2017-03-10
esdoc: remove esdoc-es7-plugin

glasserc:
Per https://github.com/esdoc/esdoc/issues/390, esdoc-es7-plugin is
deprecated and esdoc has itself grown functionality equivalent to
it. Now using esdoc-es7-plugin causes esdoc generation to fail.

Following the fix in the linked issue, just turn on all the options,
since I'm not sure which ones we use specifically.

Thanks tiberiuichim for reporting this issue on Slack.

### DIFF
--- a/test/integration_test.js
+++ b/test/integration_test.js
@@ -41,8 +41,12 @@ describe("Integration tests", function() {
   this.timeout(0);
 
   before(() => {
+    let kintoConfigPath = __dirname + "/kinto.ini";
+    if (process.env.SERVER && process.env.SERVER !== "master") {
+      kintoConfigPath = `${__dirname}/kinto-${process.env.SERVER}.ini`;
+    }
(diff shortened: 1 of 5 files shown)


glasserc:
Same as: https://github.com/Kinto/kinto-http.js/pull/166

glasserc:
This still doesn't build because we need different configuration files for Kinto 5.3.6 and master :(

test/integration_test.js
```diff
@@ -41,8 +41,12 @@ describe("Integration tests", function() {
   this.timeout(0);
 
   before(() => {
+    let kintoConfigPath = __dirname + "/kinto.ini";
+    if (process.env.SERVER && process.env.SERVER !== "master") {
```
n1k0: We could also rename `kinto.ini` to `kinto-master.ini`

n1k0:
r+

PR #662, opened 2017-03-13
Suppress lint warnings.

n1k0:
I remember we kept `let` assignments in for loops because of an incompatibility in Firefox back in the days, but I believe this has been addressed since then. Can we reintroduce `const` assignments in for loops?


glasserc:
It seems like that got fixed as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1263355 so yes.

PR #682, opened 2017-04-20
Update Prettier to 1.2.2

glasserc:
Obsoletes #677, #678, #679, #680, and #681.

The great advantage of using an automatic formatting tool is that it encodes a specific style that we can normalize all code against, removing needless reformats of our code. Now formatting tool was updated, so we get to needlessly reformat our code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,7 +141,7 @@
     "kinto-node-test-server": "1.0.0",
     "lint-staged": "^3.3.1",
     "mocha": "^3.2.0",
-    "prettier": "^1.0.0",
+    "prettier": "^1.2.2",
     "prettier-check": "^1.0.0",
     "sinon": "^2.0.0",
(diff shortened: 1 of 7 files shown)


n1k0:
These are overall improvements, no? We can always revert to manually formatted code if you prefer.

glasserc:
I'm lukewarm about the code changes themselves (changing `foo => myFoo = foo` to `foo => (myFoo = foo)` and reformatting some long conditionals in `if` statements). I think the idea of automatic enforcement of a coding style is a good one, but the problem is when that automatic coding style changes, as it does here. You end up with CI failures because the version of the tool changed, and giant blame-polluting commits reformatting things. But all things considered I guess this is still the best option we have, so feel free to ignore my complaints, I'm just cranky.

PR #692, opened 2017-05-08
Fix remote delete

### LINKED ISSUE #691: Fix handling of conflicts when server version is deleted
The first, original symptom of https://bugzilla.mozilla.org/show_bug.cgi?id=1359879 was a `record is null` error. However, the client was just doing a sync() with `SERVER_WINS`.

glasserc:
Fixes #691.

My use case also requires generally using `SERVER_WINS`, but detecting when the remote delete was not desirable and restoring the deleted record. See https://bugzilla.mozilla.org/show_bug.cgi?id=1359879 for why this might be important. In order to support this use case, I changed the values in the `resolved` list to no longer be the resolutions, but `{accepted, rejected}` objects. This will require a major version.

While I was documenting this, I noticed that I hadn't updated the `upgrading.md` file for a while, so fix that too.


docs/upgrading.md
```diff
@@ -2,6 +2,22 @@
 
 This page lists the breaking API changes between major versions of Kinto.js, as well as upgrade tips.
 
+## 8.x to 9.x
+
+* When fixing #691, the types of values in SyncResultObject.resolved were changed. Previously, they were just the resolution for a given record; now, they are {accepted, rejected}, with accepted being the new value. Any code using elements of this list (e.g. SyncResultObject.resolved[i]) should now use the accepted property of that element (e.g. SyncResultObject.resolved[i].accepted).
```
leplatrem: nit: add backticks for ``SyncResultObject.resolved`` and ``SyncResultObject.resolved[i].accepted``

src/collection.js
```diff
@@ -760,9 +761,20 @@ export default class Collection {
...
         ? conflict.local
         : conflict.remote;
-      const updated = this._resolveRaw(conflict, resolution);
-      transaction.update(updated);
-      return updated;
+      const rejected = strategy === Collection.strategy.CLIENT_WINS
+        ? conflict.remote
+        : conflict.local;
+      let accepted;
+      if (resolution === null) {
```
leplatrem: Maybe here I would add a comment to specify that it mainly applies to server wins (since ``conflict.local`` might never be null) and add a link to the piece of code in kinto-http that uses `error.details` of *412 responses*

src/collection.js
```diff
@@ -1185,9 +1200,9 @@ export default class Collection {
       await this.pushChanges(client, toSync, result, options);
 
       // Publish local resolution of push conflicts to server (on CLIENT_WINS)
-      const resolvedUnsynced = result.resolved.filter(
-        r => r._status !== "synced"
-      );
+      const resolvedUnsynced = result.resolved
+        .map(r => r.accepted)
+        .filter(r => r._status !== "synced");
```
leplatrem: 👍

glasserc:
Hmm, I didn't realize the tests had failed. It seems like they rely on the behavior in https://github.com/Kinto/kinto/pull/1102, which is only true in Kinto 6.0.0 and greater. (I didn't try with Kinto 5.3.6.) I bet that before this version, failures were turning into "Server was flushed" (like in https://bugzilla.mozilla.org/show_bug.cgi?id=1350088). I don't think it's possible to manufacture a conflict on a remote delete using Kinto 5.3.6.

I've done some investigation and it seems like Mozilla in production is running Kinto 6.0.4 and newer so I'm going to update the travis config in this PR to test against that instead of 5.3.6.

test/integration_test.js
```diff
@@ -1422,10 +1569,119 @@ describe("Integration tests", function() {
...
+            expect(syncResult.updated).to.have.length.of(0);
+          });
+
+          it("should list resolved records", () => {
+            expect(syncResult.resolved).to.have.length.of(1);
+            // FIXME: This should maybe resolve as `null` rather than
+            // `{id, status}`??
+            const accepted = syncResult.resolved[0].accepted;
+            expect(accepted).property("_status", "deleted");
+            expect(accepted).property("id");
```
glasserc: I wasn't sure whether to maybe extract `id` and `_status` and put them in the resolution rather than in `accepted`/`rejected`. I think I probably should add them to the resolution, since it's useful to not have to guess which one is null. I might also remove them from the accepted/rejected because I already have to make a major version, may as well go all the way.

n1k0: I agree with all your suggestions.

glasserc:
Coveralls build took loo long, I got impatient.

PR #710, opened 2017-05-26
turn on eslint "sort-keys" option

glasserc:
This is an experiment born out of
https://reviewboard.mozilla.org/r/138446/#review146618: turn on the
"sort-keys" option and just see what turns up.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,7 +8,8 @@
         "no-unused-vars": [2, {"vars": "all", "args": "none"}],
         "no-var": 2,
         "prefer-const": 1,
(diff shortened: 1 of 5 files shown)


glasserc:
I think we can safely call this experiment a failure. Lots of keys are ordered according to some semantic meaning of what's most important coming first. Almost all of the changes proposed by eslint removed emphasis and made code harder to read. Maybe this change is helpful sometimes, but I don't think it should be an eslint rule. Therefore, closing.

glasserc:
cc: @kmaglione

PR #711, opened 2017-05-26
Alphabetize a couple of things

glasserc:
Unlike #710, these seem semantically null, so we can reorder them
without hurting code readability.

This is a more targeted fix that addresses just the examples cited in https://reviewboard.mozilla.org/r/138446/#review146618.

This supersedes #710.

cc: @kmaglione


glasserc:
N.B. that although `sort-vars` exists, it doesn't cover the object-destructuring case yet (https://github.com/eslint/eslint/issues/7776).

PR #733, opened 2017-07-05
Add indexeddb support in Gecko export (fixes #731)

### LINKED ISSUE #731: Add support for IndexedDB adapter in Firefox export
I think we should add the `Kinto.adapters.IDB` and `Kinto.adapters.Sqlite` static properties, and add `Components.utils.importGlobalProperties(["indexedDB"]);` in the Firefox header.

* https://bugzilla.mozilla.org/show_bug.cgi?id=1377539
* https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Using_IndexedDB_in_chrome

leplatrem:
Fixes #731

* [x] POC
* [x] Validate approach by switching blocklist to indexeddb
* [ ] Check shutdown behaviour etc.
* [ ] Update https://wiki.mozilla.org/Firefox/Kinto

### DIFF
--- a/fx-src/index.js
+++ b/fx-src/index.js
@@ -16,7 +16,7 @@
 "use strict";
 
 ChromeUtils.import("resource://gre/modules/Timer.jsm");
-Cu.importGlobalProperties(["fetch"]);
+Cu.importGlobalProperties(["fetch", "indexedDB"]);
 const { EventEmitter } = ChromeUtils.import(
   "resource://gre/modules/EventEmitter.jsm",
   {}
@@ -31,14 +31,24 @@ const { KintoHttpClient } = ChromeUtils.import(
 );
 
 import KintoBase from "../src/KintoBase";
+import BaseAdapter from "../src/adapters/base";
+import IDB from "../src/adapters/IDB";
 import { RE_UUID } from "../src/utils";
 
 export default class Kinto extends KintoBase {
+  static get adapters() {
+    return {
+      BaseAdapter,
+      IDB,
+    };
+  }
+
   constructor(options = {}) {
     const events = {};
     EventEmitter.decorate(events);
 
     const defaults = {
+      adapter: IDB,
       events,
       ApiClass: KintoHttpClient,
     };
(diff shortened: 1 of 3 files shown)


fx-src/index.js
```diff
@@ -18,23 +18,44 @@
...
+import BaseAdapter from "./adapters/base";
+import IDB from "./adapters/IDB";
 import { RE_UUID } from "../src/utils";
 
 
 export default class Kinto extends KintoBase {
+
+  static get adapters() {
+    return {
+      BaseAdapter: BaseAdapter,
```
n1k0: Nit: maybe just `BaseAdapter` though I don't know if it works in Firefox yet.

leplatrem: yes it does.
The only think that doesn't is object spread rest :(

glasserc:
Looks fine. I'm not looking forward to writing the migration here :(

leplatrem:
> I'm not looking forward to writing the migration here :(

Well I don't think we'll migrate for WE. For remote settings, it's rather straightforward: local read only copy with server wins strategy.

We can keep both adapters actually.

My idea was to have clients only import `Kinto` and use `Kinto.adapters.Sqlite`, but it won't work. I import `FirefoxAdapter` to add it to the list of adapters. But `FirefoxAdapter` imports `Kinto` to extend `Kinto.adapters.BaseAdapter`. I wonder how I can fix this...

glasserc:
All the Gecko code that uses FirefoxAdapter imports it explicitly. What's wrong with that? I don't see a reason we have to change that as part of this PR.

leplatrem:
@glasserc I tested this and it works properly (most blocklist tests pass in Gecko)

fx-src/jsm_prefix.js
```diff
@@ -30,4 +30,4 @@
 // more details.
 const global = this;
 
-this.EXPORTED_SYMBOLS = ["Kinto"];
+var EXPORTED_SYMBOLS = ["Kinto"];
```
glasserc: What's this change about?

leplatrem: https://bugzilla.mozilla.org/show_bug.cgi?id=1440284 :)

PR #772, opened 2017-10-18
Fixed esdoc generation

zakaluka:
See https://doc.esdoc.org/github.com/zakaluka/kinto.js/ for what is now being generated.

Had to remove the 'testing' configuration as I don't understand why it is failing.  

I've left esdoc-integrate-test-plugin as a dev dependency on the off-chance that someone knows how to integrate it into the "new" esdoc configuration file (and hosting service).

### DIFF
--- a/esdoc.json
+++ b/esdoc.json
@@ -3,6 +3,14 @@
   "destination": "./esdoc",
   "plugins": [
     {
+      "name": "esdoc-importpath-plugin",
+      "option": {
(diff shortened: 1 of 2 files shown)


zakaluka:
I couldn't tell what the old docs looked like.  However, looking at the source, a lot of the docs have NOT been generated correctly.  Will close this pull request for now and re-submit when I can get that documentation to show.

PR #774, opened 2017-10-20
Remove IDB#resetSyncStatus

glasserc:
Revert 3017452aca9672681c9328ec93853ce472d31faf.

It turns out that there's already a Collection#resetSyncStatus, so
IDB#resetSyncStatus just serves to confuse the user.

The place where I intended to use this, I already accidentally used
Collection#resetSyncStatus, so there truly isn't any need for this
method.

The commit that this reverts got squashed as part of #767, so the
commit ID in this commit may not exist in this repository. Instead,
see 7a23ff9b16f3f85dcc92c1d998b2faa86d367006.

### DIFF
--- a/src/adapters/IDB.js
+++ b/src/adapters/IDB.js
@@ -409,47 +409,6 @@ export default class IDB extends BaseAdapter {
   }
 
   /**
-   * Reset the sync status of this collection.
-   * This is useful when a server has been wiped, for instance.
-   *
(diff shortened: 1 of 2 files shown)


glasserc:
I guess this is what happens when you forget to read the documentation before writing code.

There is a FirefoxAdapter#resetSyncStatus, but this is more powerful -- it resets the sync status for *all* the collections, which isn't possible in the IDB adapter.

PR #775, opened 2017-10-22
Minor documentation updates

zakaluka:


zakaluka:
The error appears to be with the `just-extend` package.

zakaluka:
The Travis CI server is **VERY** temperamental...

src/adapters/IDB.js
```diff
@@ -189,7 +189,7 @@ export default class IDB extends BaseAdapter {
   }
 
   /**
-   * Returns a transaction and a store objects for this collection.
+   * Returns a transaction and a object store for this collection.
```
glasserc: This should be `an object store`.

zakaluka: Updated.

PR #778, opened 2017-11-06
Fix the result of upsert.

francois2metz:


leplatrem:
Excellent catch! Thank you!

PR #787, opened 2017-12-05
Update prettier and reformat everything

glasserc:


Natim:
Back to double quotes man 👍

PR #796, opened 2018-02-12
Firefox updates

glasserc:
This is similar to https://github.com/Kinto/kinto-http.js/pull/273, but for this library.

### DIFF
--- a/fx-src/jsm_prefix.js
+++ b/fx-src/jsm_prefix.js
@@ -18,6 +18,16 @@
  * This file is generated from kinto.js - do not modify directly.
  */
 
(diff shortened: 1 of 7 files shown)


fx-src/jsm_prefix.js
```diff
@@ -12,9 +12,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+"use strict";
 
 /*
  * This file is generated from kinto.js - do not modify directly.
  */
 
+const global = this;
```
leplatrem: Should we add a comment about why this is necessary?

PR #799, opened 2018-03-04
Add support for deleteAll method

agawish:
Fixes: #556

Changes the following Files:
1. src/collection.js: Added deleteAll function in two places:
   a. As an async function inside Collections class - Takes no parameters -
   b. As a function inside CollectionTransaction class - Takes records parameter - to execute the delete transaction and emit "deleteAll" event

2. test/collection_test.js: Added two tests:
   a. Inside Events test suite to test for "deleteAll" event
   b. Inside #execute test suite to test for the soft delete
   All tests are passing.

Also ran cs-format to follow coding style and added the git pre-commit hook

Signed-off-by: Amr Gawish <amr.gawish@gmail.com>

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -611,6 +611,23 @@ export default class Collection {
   }
 
   /**
+   * Soft Delete all records from the local database.
+   *
+   * @return {Promise}
+   */
+  async deleteAll() {
+    const records = await this.db.list();
+    // await results.forEach( (record) => {
+    //     delete(record.id);
+    // });
+    // this._queueEvent("deleteAll", { data: [] });
+    // return { data: [], permissions: {} };
+    return this.execute(transaction => {
+      return transaction.deleteAll(records);
+    }, {});
+  }
+
+  /**
    * The same as {@link CollectionTransaction#deleteAny}, but wrapped
    * in its own transaction.
    *
@@ -1420,6 +1437,19 @@ export class CollectionTransaction {
     }
     this._queueEvent("delete", { data: existing });
     return { data: existing, permissions: {} };
+  }
+
+  /**
+   * Soft Delete all records from the local database.
+   * @param  {Object} allRecords       All the records in the collection.
+   * @return {Object}
+   */
+  deleteAll(allRecords) {
+    allRecords.forEach(record => {
+      delete record.id;
+    });
+    this._queueEvent("deleteAll", { data: [] });
+    return { data: [], permissions: {} };
   }
 
   /**
--- a/src/collection.js
+++ b/src/collection.js
@@ -617,11 +617,12 @@ export default class Collection {
    */
   async deleteAll() {
     const { data } = await this.list({}, { includeDeleted: false });
+    const recordIds = data.map(record => record.id);
     return this.execute(
       transaction => {
-        return transaction.deleteAll(data.map(record => record.id));
+        return transaction.deleteAll(recordIds);
       },
-      { preloadIds: data.map(record => record.id) }
+      { preloadIds: recordIds }
     );
   }
 
--- a/src/collection.js
+++ b/src/collection.js
@@ -611,20 +611,18 @@ export default class Collection {
   }
 
   /**
-   * Soft Delete all records from the local database.
+   * Same as {@link Collection#deleteAll}, but wrapped in its own transaction, execulding the parameter.
    *
    * @return {Promise}
    */
   async deleteAll() {
-    const records = await this.db.list();
-    // await results.forEach( (record) => {
-    //     delete(record.id);
-    // });
-    // this._queueEvent("deleteAll", { data: [] });
-    // return { data: [], permissions: {} };
-    return this.execute(transaction => {
-      return transaction.deleteAll(records);
-    }, {});
+    const { data } = await this.list({}, { includeDeleted: false });
+    return this.execute(
+      transaction => {
+        return transaction.deleteAll(data.map(record => record.id));
+      },
+      { preloadIds: data.map(record => record.id) }
+    );
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/collection.js
```diff
@@ -610,6 +610,23 @@ export default class Collection {
...
+  async deleteAll() {
+    const records = await this.db.list();
+    // await results.forEach( (record) => {
+    //     delete(record.id);
+    // });
+    // this._queueEvent("deleteAll", { data: [] });
+    // return { data: [], permissions: {} };
+    return this.execute(transaction => {
+      return transaction.deleteAll(records);
+    }, {});
```
leplatrem: I believe we can list and delete in the same transaction. In other words, I think the transaction.deleteAll() could be without parameters maybe

src/collection.js
```diff
@@ -610,6 +610,23 @@ export default class Collection {
...
+   *
+   * @return {Promise}
+   */
+  async deleteAll() {
+    const records = await this.db.list();
+    // await results.forEach( (record) => {
+    //     delete(record.id);
+    // });
+    // this._queueEvent("deleteAll", { data: [] });
+    // return { data: [], permissions: {} };
```
leplatrem: nitpick: leftover

src/collection.js
```diff
@@ -1422,6 +1439,19 @@ export class CollectionTransaction {
...
   }
 
+  /**
+   * Soft Delete all records from the local database.
+   * @param  {Object} allRecords       All the records in the collection.
+   * @return {Object}
+   */
+  deleteAll(allRecords) {
+    allRecords.forEach(record => {
+      delete record.id;
```
leplatrem: this won't work, it only deletes the `id` property from the `record` object

src/collection.js
```diff
@@ -1422,6 +1439,19 @@ export class CollectionTransaction {
...
+    });
+    this._queueEvent("deleteAll", { data: [] });
```
leplatrem: I believe it would be more useful for developers to have information about the deleted records instead of an empty list

Maybe, instead of adding a new `deleteAll` event, we could emit an individual `delete` event for each delete record

src/collection.js
```diff
@@ -1422,6 +1439,19 @@ export class CollectionTransaction {
...
+    return { data: [], permissions: {} };
```
leplatrem: shouldn't we return the deleted records?

test/collection_test.js
```diff
@@ -2859,6 +2859,13 @@ describe("Collection", () => {
         .then(result => expect(result.data._status).eql("deleted"));
     });
 
+    it("should support deleteAll", () => {
+      articles.deleteAll();
+      articles
+        .list({ filters: { _status: "deleted" } }, { includeDeleted: true })
+        .then(result => expect(result.data._status).eql("deleted"));
```
leplatrem: I don't believe this work :thinking: 
`list()` returns a list of records, I can't see how `result.data._status` could be equal to `deleted`

src/collection.js
agawish: Sure we can, it will just be returning the same as `.list()` anyway :)

src/collection.js
agawish: `delete` events will be emitted since I'm using the delete method, which emits delete. I thought you might want to know that `deleteAll` is triggered for future reference!

test/collection_test.js
agawish: why not, I'm calling `delete` within `deleteAll` which should set the `_status` to `"deleted"` so it should work :)

src/collection.js
agawish: if I recall the delete method takes the record Id, looks it up and set the `_status` to delete (default option is soft delete) so theoretically, it should work and the unit test agrees!

src/collection.js
agawish: I didn't find examples of that, and I thought it might be useful to parameterize those records, maybe it can be utilized by a future functionality.

test/collection_test.js
leplatrem: But `result` here, in the `then()` takes the result of the `list()` method call, which is an array. This code reads the `data` attribute of the array, that's why I'm puzzled!

src/collection.js
leplatrem: yes OK

leplatrem: But we could add a test then (that asserts we receive events for individual records)

src/collection.js
leplatrem: But here you don't call the `delete()` method, you use the `delete` keyword which does something else :)

agawish: I'm calling the delete method, it was formatted this way by `cs-format` for some reason 😕

src/collection.js
agawish: Good one, I'll do that then

test/collection_test.js
agawish: I can trace that down, to see why this is working!

src/collection.js
```diff
@@ -610,6 +610,21 @@ export default class Collection {
...
+   *
+   * @return {Promise}
+   */
+  async deleteAll() {
+    const { data } = await this.list({}, { includeDeleted: false });
+    return this.execute(
+      transaction => {
+        return transaction.deleteAll(data.map(record => record.id));
+      },
+      { preloadIds: data.map(record => record.id) }
```
leplatrem: nitpick: avoid iterating twice (store in variable like `recordIds`)

test/collection_test.js
```diff
@@ -1167,6 +1167,33 @@ describe("Collection", () => {
...
+
+    it("should be able to soft delete all articles", () => {
+      return articles.deleteAll().then(res => {
+        res.data.forEach(record => {
+          articles
+            .get(record.id)
+            .then(res => res.data._status)
+            .should.eventually.eql("deleted");
+        });
+      });
```
leplatrem: I believe this test could be simplified by just testing `deleteAll()` → `list()` returns 0 elements → `list({includeDeleted: true})` returns 5 elements

src/collection.js
agawish: Good spot, map is too useful and easy to use I didn't even notice it 😄

agawish:
Thanks @leplatrem for your guidance, I loved every bit of this :+1:

leplatrem:
Excellent! 

See you around I hope :)

PR #801, opened 2018-03-07
Pull every pages during sync (fixes #355)

### LINKED ISSUE #355: Add support for crawling paginated list of changes
Right now we only fetch the first page sent by the server.

Note: this is a limitation when the server defines a `paginate_by` setting.

leplatrem:
fixes #355

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -956,7 +956,7 @@ export default class Collection {
       since: options.lastModified ? `${options.lastModified}` : undefined,
       headers: options.headers,
       retry: options.retry,
-      // Fetch every page by default (see #277 about limiting)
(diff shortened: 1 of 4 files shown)


src/collection.js
```diff
@@ -956,6 +956,8 @@ export default class Collection {
       since: options.lastModified ? `${options.lastModified}` : undefined,
       headers: options.headers,
       retry: options.retry,
+      // Fetch every page by default (see #277 about limiting)
```
glasserc: Is the reference to #277 meant to be an acknowledgement of future work? If so, maybe put `FIXME` or something?

Issue #820, opened 2018-05-28
Store every collections in one indexeddb  


leplatrem:
Currently it looks like it creates a DB per bucket/collection (+optional prefix)

We should be able to have:

- one database per prefix
- one timestamp table (bid/cid, timestamp)
- one records table (bid/cid, rid, data)
- one status table (remote, bid/cid/rid, status) for #637 

Later 
- one collection table (bid, cid, data) for #666

PR #823, opened 2018-06-25
Add DBName option to IndexedDB adapter (ref #820)

leplatrem:
Just one little step of #820 

We need this for Firefox, where we create a lot of IndexedDBs that only have one store.
With this PR, will be able to gather them all under one *RemoteSettings* one.

screenshot from 2018-06-25 16-16-53

For clients upgrading to a version containing this change, the first synchronization will fetch all records from the remote collection instead of using `?_since`. But since identical ones will be ignored during import, there is no major impact.


leplatrem:
Unfortunately I'm having problems with this approach because of the way the IndexedDB API works (that's why it's so popular right?!)

Basically: 
- store first collection data: database is created with the `__meta__` store and the `{bid}/{cid}` 
- store second collection data: the database is already at expected version. On upgrade needed not executed and the second store `{bid}/{cid2}` is not created.

Why does it work in the unit tests here ? Well, good question.

Can you hack and play with the version numbers while inspecting `db.objectStoreNames.contains()` ? Oh, I guess you don't... among the nicest errors:

- `The operation failed because the stored database is a higher version than the version requested.`. Which version bordel ? ... yes once the DB is at version X, you cannot open it at a previous version.

- `NotFoundError: The operation failed because the requested database object could not be found. For example, an object store did not exist but was being opened.`. Which store bordel ? 

- `InvalidStateError: A mutation operation was attempted on a database that did not allow mutations.`. No you can't create stores outside `onupgradeneeded`


So I guess we'll have to declare this *dbname option* feature fully broken, and go for #820 for real

PR #826, opened 2018-07-10
Relax record id validation according to server (fixes #824)

### LINKED ISSUE #824: Synchronization fails with custom record ids
Why do we have this?

https://github.com/Kinto/kinto.js/blob/150b4d708931c84a8bda491c5af25243a90df762/src/collection.js#L118-L120

We shouldn't.

leplatrem:
Fixes #824

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -4,7 +4,7 @@ import BaseAdapter from "./adapters/base";
 import IDB from "./adapters/IDB";
 import { waterfall, deepEqual } from "./utils";
 import { v4 as uuid4 } from "uuid";
-import { isUUID, omitKeys } from "./utils";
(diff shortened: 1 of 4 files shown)


leplatrem:
Thanks!

glasserc:
Wait, doesn't this essentially disable validation that IDs conform to anything, even if they're created on the client? This doesn't seem like the right fix to me.

glasserc:
Thinking this through a little more, I guess the question is whether the `createUUIDSchema` should also enforce that all record IDs are UUIDs. I guess the answer to that question is no, because the name `createUUIDSchema` suggests that it only uses UUIDs when necessary to create an ID.

PR #831, opened 2018-08-09
One database for several collections

### LINKED ISSUE #820: Store every collections in one indexeddb

leplatrem:
Fixes #820

- [x] Create new schema
- [x] Rewrite queries
- [x] Investigate if creating indexes for the field that serves as key is necessary
- [x] Migrate data from version 1
- [x] Documentation API breaking changes
- [x] Update docs about dbPrefix and dbName


leplatrem:
Note: On my machine, querying for the existence of 10000 different IDB databases took around 600ms. So I think this is acceptable to check for older versions when the connection is opened. We can have a flag `migrateOldVersion` defaulted to `false` but that developers upgrading to the version 10.X would set to `true` for example.

leplatrem:
Creating an index for the key appears to be useless, eg. with 100K records:
```
open with-idx version 1
script.js:73 insert-with-index: 7720.32177734375ms
script.js:3 open without-idx version 1
script.js:86 insert-without-index: 8473.73779296875ms
script.js:103 query-with-index: 1.250244140625ms
script.js:120 query-without-index: 0.77294921875ms
```
which confirms what is said here https://stackoverflow.com/questions/31908605/indexeddb-is-keypath-already-an-index

leplatrem:
Migration and tests work \o/

I tested the migration of the data manually with this procedure:

### Create some data with the old version
```
$ git co master
$ npm run demo
```
Then interact with the API (synchronize, create local data, etc.) 

In the inspector, you can see the local database.

### Switch to the new version
```
$ git co 820-one-indexeddb
$ npm run demo
```
Refresh the page. The data is still there.

In the inspector, you can see the new database, the old one was deleted.

docs/upgrading.md
```diff
@@ -2,6 +2,26 @@
 
 This page lists the breaking API changes between major versions of Kinto.js, as well as upgrade tips.
 
+## 11.x to 12.x
+
+* The `dbPrefix` option in the `Kinto` and `Collection` constructors was dropped in favor of the `dbName` field in the `adapterOptions`.
+
+**Database Schema Change**
+
+This version introduces a major change: instead of having one IndexedDB database per collection (named `{bucket}/{collection}`), we now have only one database (called `KintoDB` by default) which stores records indexed by collection.
```
glasserc: I think it would be valuable to add a little bit more information here about the new database structure. Like Fred Brooks wrote, "Show me your flowcharts and conceal your tables, and I shall continue to be mystified. Show me your tables, and I won’t usually need your flowcharts; they’ll be obvious." My understanding from reading the code is that the database contains two stores, a `records` store and a `timestamp` store. The `records` store contains the records for all collections as plain JS objects with a `_cid` field corresponding to `${bucket}/${collection}`. Is that right? Out of curiosity, why do it this way rather than creating a separate store for `${bucket}/${collection}`? Is it because this way we can enumerate the collections, whereas we cannot enumerate the different stores in an IDB database?

leplatrem: > a little bit more information here about the new database structure

Will do!

> Is that right? 

Yes it's right. 

> Out of curiosity, why do it this way rather than creating a separate store for  ${bucket}/${collection}?

That was the approach we had before (https://github.com/Kinto/kinto.js/pull/823#issuecomment-400265375)

Basically, you cannot add new stores without upgrading the version of the database. So if we want to have an arbitrary list of stores (based on the collections that were synchronized) it's a nightmare.

src/adapters/IDB.js
```diff
@@ -39,70 +116,70 @@ const cursorHandlers = {
...
+
   if (!indexField) {
     // Get all records.
-    const request = store.openCursor();
+    const request = store.index("cid").openCursor(IDBKeyRange.only(cid));
     request.onsuccess = cursorHandlers.all(filters, done);
     return request;
   }
 
+  const value = filters[indexField];
```
peterbe: This seems dangerous. `indexField` could be `undefined` and who knows what should happen when you do `someMap[undefined]`? I'm pretty sure the output of `someMap[undefined]` is `undefined` but `value` is used in a bunch of other things later.

src/adapters/IDB.js
```diff
@@ -39,70 +116,70 @@ const cursorHandlers = {
...
+ * @param  {Object}           filters    Filter the records by field.
  * @param  {Function}         done       The operation completion handler.
  * @return {IDBRequest}
  */
-function createListRequest(store, indexField, value, filters, done) {
+function createListRequest(cid, store, filters, done) {
+  // Introspect filters and check if they leverage an indexed field.
+  const indexField = Object.keys(filters).filter(field => {
+    return INDEXED_FIELDS.includes(field);
+  })[0];
```
peterbe: You probably mean to use: `Object.keys(filters).find(field => INDEXED_FIELDS.includes(field))` to get the first (or `undefined`) from the array.

src/adapters/IDB.js
```diff
@@ -144,38 +219,54 @@ export default class IDB extends BaseAdapter {
...
+        const recordsStore = db.createObjectStore("records", {
+          keyPath: ["_cid", "id"],
         });
-        // Primary key (generated by IdSchema, UUID by default)
-        collStore.createIndex("id", "id", { unique: true });
+        // An index to obtain all the records in a collection.
+        recordsStore.createIndex("cid", "_cid");
+        // Here we create indices for every known field in records by collection.
+        // Record id (generated by IdSchema, UUID by default)
+        recordsStore.createIndex("id", ["_cid", "id"]);
```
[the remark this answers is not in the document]
leplatrem: It's not necessary indeed. But because of the code in `createListRequest()`, it is a lot easier to manage if there is an index with the same name as the indexed column, so I left the `id` one.

src/adapters/IDB.js
```diff
@@ -39,70 +116,70 @@ const cursorHandlers = {
...
   if (Array.isArray(value)) {
-    const request = store.index(indexField).openCursor();
-    request.onsuccess = cursorHandlers.in(value, done);
+    if (value.length === 0) {
+      return done([]);
+    }
+    const values = value.map(i => [cid, i]).sort();
+    const range = IDBKeyRange.bound(values[0], values[values.length - 1]);
+    const request = store.index(indexField).openCursor(range);
+    request.onsuccess = cursorHandlers.in(values, done);
```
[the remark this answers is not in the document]
leplatrem: Oh that's a good catch! Thanks!

src/adapters/IDB.js
```diff
@@ -277,42 +371,49 @@ export default class IDB extends BaseAdapter {
...
+            // The callback has thrown an error explicitly. Abort transaction cleanly.
+            abort(e);
           }
-          // XXX unsure if we should manually abort the transaction on error
-          transaction.onerror = event => reject(new Error(event.target.error));
-          transaction.oncomplete = event => resolve(result);
+        };
+
+        // No option to preload records, go straight to `callback`.
+        if (options.preload.length == 0) {
```
peterbe: Shouldn't this be `===` or just `if (!options.preload.length)`?

src/adapters/IDB.js
```diff
@@ -403,15 +489,15 @@ export default class IDB extends BaseAdapter {
...
-      transaction.oncomplete = event => {
-        resolve((request.result && request.result.value) || null);
-      };
-    });
+    try {
+      let entry;
+      await this.prepare("timestamps", store => {
+        store.get(this.cid).onsuccess = e => (entry = e.target.result);
+      });
+      return entry ? entry.value : null;
```
peterbe: This looks scary. if `entry` is a map (e.g. `{}`) the result of this line is `undefined`. Had it been Python, the output of `map.value if map else None` would be `None` if `map` was `{}`.

src/utils.js
```diff
@@ -127,3 +127,15 @@ export function omitKeys(obj, keys = []) {
     return acc;
   }, {});
 }
+
+export function arrayEqual(a, b) {
+  if (a.length != b.length) {
```
peterbe: `!==` instead?

src/utils.js
```diff
@@ -127,3 +127,15 @@ export function omitKeys(obj, keys = []) {
...
+    return false;
+  }
+  for (let i = 0, n = a.length; i < n; i++) {
+    if (a[i] != b[i]) {
```
peterbe: ditto

src/utils.js
```diff
@@ -127,3 +127,15 @@ export function omitKeys(obj, keys = []) {
     return acc;
   }, {});
 }
+
+export function arrayEqual(a, b) {
+  if (a.length != b.length) {
+    return false;
+  }
+  for (let i = 0, n = a.length; i < n; i++) {
```
peterbe: nit: You can use just one variable instead:
```javascript
for (let i = a.length; i--;) {
  if (a[i] !== b[i]) return false;
}
```

demo/demo.js
```diff
@@ -1,5 +1,10 @@
 function main() {
-  var db = new Kinto();
+  var db = new Kinto({
+    adapterOptions: {
+      // dbName: "DemoDB",
```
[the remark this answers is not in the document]
leplatrem: I was thinking to left it as a way to document it. What do you think?

src/adapters/IDB.js
```diff
@@ -1,22 +1,103 @@
...
+/**
+ * Helper to wrap the deletion of an IndexedDB database into a promise.
+ *
+ * @param dbName {String} the database to delete
+ * @return {Promise}
+ */
+async function deleteDatabase(dbName) {
+  return new Promise((resolve, reject) => {
+    const request = indexedDB.deleteDatabase(dbName);
+    request.onsuccess = resolve();
```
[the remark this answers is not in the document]
leplatrem: Oh. Very good catch.

demo/demo.js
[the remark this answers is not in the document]
peterbe: I like that. But perhaps mention it in a comment above so it doesn't *look* like a piece of leftover. E.g. `// Uncomment the next line to try different database names`

src/adapters/IDB.js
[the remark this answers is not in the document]
peterbe: Does that mean there wasn't correct automated testing to catch that?

src/adapters/IDB.js
leplatrem: we have `if (!indexField) {` above.

src/adapters/IDB.js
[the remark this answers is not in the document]
leplatrem: Well, the deletion actually works and is tested. But this code could have cause race conditions issues since the promise would potentially resolve before the deletion occurs.

peterbe:
@leplatrem I looked at the followup commits that were regarding my myriad of nits only on the code. Do you want me to take a broader/bigger review of the whole thing or is @glasserc 's review good enough on that aspect?

glasserc:
Looks great to me!

leplatrem:
I think I'm happy with the changes :) Thanks a lot for your comments!
One thing that could improve would be using an external library like mentioned in #832 

@peterbe could you try the migration with your project?

PR #839, opened 2018-08-16
Throw explicit error if data to migrate is found and adapter option is not set

leplatrem:
We realized that most developers will probably upgrade without reading the release notes. 

So to avoid that, we now raise an explicit error if an old database is found and the `migrateOldData` is not explicitly set.

### DIFF
--- a/src/adapters/IDB.js
+++ b/src/adapters/IDB.js
@@ -239,13 +239,16 @@ export default class IDB extends BaseAdapter {
     // Note: the built-in migrations from IndexedDB can only be used if the
     // database name does not change.
     const hasMigrateOption = this._options.hasOwnProperty("migrateOldData");
-    const toMigrate = this._options.migrateOldData || !hasMigrateOption
-      ? await migrationRequired(this.cid)
-      : null;
+    const toMigrate =
+      this._options.migrateOldData || !hasMigrateOption
+        ? await migrationRequired(this.cid)
+        : null;
 
     if (!!toMigrate && !hasMigrateOption) {
-      throw new Error("An old IndexedDB database was found, but the `migrateOldData` option was not set. " +
-                      "Check out ugprade notes https://kintojs.readthedocs.io/en/latest/upgrading/");
+      throw new Error(
+        "An old IndexedDB database was found, but the `migrateOldData` option was not set. " +
+          "Check out ugprade notes https://kintojs.readthedocs.io/en/latest/upgrading/"
+      );
     }
 
     this._db = await open(this.dbName, {
--- a/src/adapters/IDB.js
+++ b/src/adapters/IDB.js
@@ -238,19 +238,9 @@ export default class IDB extends BaseAdapter {
     // Check if it exists, and migrate data once new schema is in place.
     // Note: the built-in migrations from IndexedDB can only be used if the
     // database name does not change.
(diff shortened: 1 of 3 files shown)


src/adapters/IDB.js
```diff
@@ -587,7 +596,7 @@ async function migrationRequired(dbName) {
     return null;
   }
 
-  console.warn(`${dbName}: old IDB database found."`);
+  console.warn(`${dbName}: old IndexedB database found.`);
```
peterbe: IndexedDB

src/adapters/IDB.js
```diff
@@ -238,9 +238,18 @@ export default class IDB extends BaseAdapter {
...
-      : null;
+    const hasMigrateOption = this._options.hasOwnProperty("migrateOldData");
+    const toMigrate =
+      this._options.migrateOldData || !hasMigrateOption
+        ? await migrationRequired(this.cid)
+        : null;
+
+    if (!!toMigrate && !hasMigrateOption) {
+      throw new Error(
+        "An old IndexedDB database was found, but the `migrateOldData` option was not set. " +
```
peterbe: `was not explicitly set` instead of `was not set`

src/adapters/IDB.js
```diff
@@ -238,9 +238,18 @@ export default class IDB extends BaseAdapter {
...
+    const hasMigrateOption = this._options.hasOwnProperty("migrateOldData");
+    const dataToMigrate =
+      this._options.migrateOldData || !hasMigrateOption
+        ? await migrationRequired(this.cid)
+        : null;
+
+    if (dataToMigrate && !hasMigrateOption) {
+      throw new Error(
+        "An old IndexedDB database was found, but the `migrateOldData` option was not set. " +
+          "Check out ugprade notes https://kintojs.readthedocs.io/en/latest/upgrading/"
```
peterbe: `upgrade` not `ugprade`

src/adapters/IDB.js
```diff
@@ -238,9 +238,19 @@ export default class IDB extends BaseAdapter {
...
+    const dataToMigrate =
+      this._options.migrateOldData || !hasMigrateOption
+        ? await migrationRequired(this.cid)
+        : null;
+
+    if (dataToMigrate && !hasMigrateOption) {
+      throw new Error(
+        "An old IndexedDB database with data was found, but the `migrateOldData` option " +
+          "was not explicitly set. " +
+          "Check out upgrade notes https://kintojs.readthedocs.io/en/latest/upgrading/"
```
peterbe: Add `#11x-to-12x` to the URL

leplatrem:
Something is wrong. And I can't investigate it before tomorrow: https://travis-ci.org/Kinto/kinto.js/jobs/416897129

leplatrem:
With this PR, if you don't specify advanced adapter options, we will try to open the old database. And it will create it and delete it. Each time we perform an operation (each `db.open()`)

I'm really not sure I want this. Nor force everyone to add `{ adapterOptions: { migrateOldData: false } } ` to the main constructor.

If we had a solid and efficient way (and concurrency-proof) way to check if a database exists, I would be in favor, but everything I try results to be fragile.

I would be happy to hear your suggestions :)

Issue #849, opened 2018-09-18
Adding Kinto to create-react-app breaks yarn build


peterbe:
Steps to reproduce:

```bash
▶ cd /tmp
▶ create-react-app my-kinto-app
▶ cd my-kinto-app
▶ yarn add kinto
```
Then edit `src/App.js` to something like this:
```
...
import Kinto from "kinto";

class App extends Component {
   componentDidMount() {
     this.db = new Kinto()
   }
...
```
(not sure if this is necessary because I don't know if CRA's treeshaking would remove the dependency when building)

```bash
▶ yarn build
yarn run v1.9.4
$ react-scripts build
Creating an optimized production build...
Failed to compile.

Module not found: Error: Can't resolve '@babel/runtime/helpers/interopRequireDefault' in '/private/tmp/my-kinto-app/node_modules/kinto/lib'


error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

```

peterbe:
By the way, this works but is not a solution:

```bash
▶ yarn add kinto@11.2.2
▶ yarn build
...
Compiled successfully.
...
```

peterbe:
Actually, this works too:
```bash
▶ yarn add kinto@12.0.0
▶ yarn build
...
Compiled successfully.
...
```

So the regression is in [12.0.0...12.0.1](https://github.com/Kinto/kinto.js/compare/v12.0.0...v12.0.1).

peterbe:
I don't really know what I'm looking for but the diff between the `node_modules/kinto/lib` is very different:

```
▶ tree node_modules/kinto/lib
node_modules/kinto/lib
├── KintoBase.js
├── adapters
│   ├── IDB.js
│   ├── LocalStorage.js
│   └── base.js
├── api.js
├── collection.js
├── errors.js
├── http.js
├── index.js
├── kinto.js
├── kinto.min.js
├── kinto.noshim.js
├── moz-kinto-offline-client.js
└── utils.js

1 directory, 14 files

/tmp/my-kinto-app
▶ tree /tmp/node_modules_12.0.0/kinto/lib
/tmp/node_modules_12.0.0/kinto/lib
├── KintoBase.js
├── adapters
│   ├── IDB.js
│   ├── LocalStorage.js
│   └── base.js
├── api.js
├── collection.js
├── errors.js
├── http.js
├── index.js
└── utils.js

1 directory, 10 files
```

[...]

PR #850, opened 2018-09-18
Explicitly depend on @babel/runtime

glasserc:
This is meant to address #849.


leplatrem:
we can release a 12.0.2 ;)

PR #857, opened 2018-10-03
Updated some examples with await/async in api.md

bastienfp:
I rewrote some examples with await/async and try/catch for errors.

I also noted at line 388:
//Code needs review here before async/await example refactor
```js
articles.execute(txn => {
    let article1 = txn.get(id1);
    let article2 = txn.get(id2);
    return [article1, article2];
  }, {preloadIds: [id1, id2])
  .then(console.log.bind(console));
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -48,9 +48,11 @@ The collection object has the following (read-only) attribute:
 ## Creating a record
 
 ```js
-articles.create({title: "foo"})
-  .then(console.log.bind(console))
-  .catch(console.error.bind(console));
+try {
+  await articles.create({title: "foo"});
+} catch(e) {
+  console.error;
+}
 ```
 
 Result is:
@@ -76,9 +78,11 @@ but you can also define a [custom ID schema](#id-schemas));
 ## Retrieving a single record
 
 ```js
-articles.get("2dcd0e65-468c-4655-8015-30c8b3a1c8f8")
-  .then(console.log.bind(console))
-  .catch(console.error.bind(console));
+try {
+  await articles.get("2dcd0e65-468c-4655-8015-30c8b3a1c8f8");
+} catch(e) {
+  console.error;
+}
 ```
 
 Result:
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -76,9 +78,11 @@ but you can also define a [custom ID schema](#id-schemas));
 ## Retrieving a single record
 
 ```js
-articles.get("2dcd0e65-468c-4655-8015-30c8b3a1c8f8")
-  .then(console.log.bind(console))
+try {
+  await articles.get("2dcd0e65-468c-4655-8015-30c8b3a1c8f8");
+} catch(e) {
+  console.error;
```
glasserc: Shouldn't these be `console.error(e)`?

docs/api.md
```diff
@@ -338,14 +334,14 @@ actually starting the first sync with it.
...
+let records = await articles.loadDump([
   {
     id: "2dcd0e65-468c-4655-8015-30c8b3a1c8f8",
     title: "baz",
     last_modified: 1432222889337
   }
-])
-  .then(records => console.log(records));
+]);
+console.log(records));
```
glasserc: What's the reasoning for which places get `console.log` and which don't?

leplatrem:
OH ! I'm sorry, I merged #858 without paying attention to this one :(

docs/api.md
itaisteinherz: [`console.log`](https://developer.mozilla.org/en-US/docs/Web/API/Console/log) is used whenever you want to log a general value to the console, and [`console.error`](https://developer.mozilla.org/en-US/docs/Web/API/Console/error) is used whenever you want to log an error to the console.
There are also [a few more logging methods](https://developer.mozilla.org/en-US/docs/Web/API/console#Methods) you can check out, notably [`console.log`](https://developer.mozilla.org/en-US/docs/Web/API/Console/log), [`console.error`](https://developer.mozilla.org/en-US/docs/Web/API/Console/error), [`console.warn`](https://developer.mozilla.org/en-US/docs/Web/API/Console/warn) and [`console.info`](https://developer.mozilla.org/en-US/docs/Web/API/Console/info).

dstaley:
Fixed in https://github.com/Kinto/kinto.js/pull/1075

PR #860, opened 2018-10-08
Add support for expectedTimestamp to sync() and pullChanges()

glasserc:
sync() appears to have some support for a `filters` option, but that
doesn't really make sense since we don't support syncing partial
collections.

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -925,6 +925,8 @@ export default class Collection {
    * @param  {KintoClient.Collection} client           Kinto client Collection instance.
    * @param  {SyncResultObject}       syncResultObject The sync result object.
    * @param  {Object}                 options
+   * @param  {String}                 options.expectedTimestamp: A timestamp to use as a "cache
+   *   busting" query parameter.
    * @return {Promise}
    */
   async pullChanges(client, syncResultObject, options = {}) {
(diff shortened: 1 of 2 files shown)


leplatrem:
> sync() appears to have some support for a filters option, 

I don't think we should allow it then.

src/collection.js
```diff
@@ -925,6 +925,8 @@ export default class Collection {
    * @param  {KintoClient.Collection} client           Kinto client Collection instance.
    * @param  {SyncResultObject}       syncResultObject The sync result object.
    * @param  {Object}                 options
+   * @param  {String}                 options.expectedTimestamp: A timestamp to use as a "cache
+   *   busting" query parameter.
```
leplatrem: Is it the way we ident docs elsewhere?

src/collection.js
```diff
@@ -1184,6 +1192,8 @@ export default class Collection {
    *
    * Options:
    * - {Object} headers: HTTP headers to attach to outgoing requests.
+   * - {String} expectedTimestamp: A timestamp to use as a "cache
+   *   busting" query parameter.
```
leplatrem: Can you please add it to the docs please?

src/collection.js
```diff
@@ -1200,11 +1210,13 @@ export default class Collection {
     options = {
       strategy: Collection.strategy.MANUAL,
       headers: {},
+      filters: {},
```
leplatrem: where do we use this?

leplatrem:
I'll merge it manually and address my nits :)

glasserc:
Ugh, I guess there is no `filters()` in sync. Somehow I misread code that I had half-written and thought it was already present in master :disappointed:

PR #864, opened 2018-10-17
Allow dots in records IDs

leplatrem:
Counterpart of https://github.com/Kinto/kinto/pull/1822

If the server allows dots, then kinto.js has to accept them when loading
a dump.


leplatrem:
Not doing it.

PR #865, opened 2018-10-17
Fix #862: Use getAll() in IDB if no filter is set

### LINKED ISSUE #862: Use IDB getAll() when no filter is set
Instead of using cursors..
https://developer.mozilla.org/en-US/docs/Web/API/IDBObjectStore/getAll

leplatrem:
Fix #862 

I also want to investigate the performance impact depending on the collection size (eg. 100, 1K, 10K records)

### DIFF
--- a/src/adapters/IDB.js
+++ b/src/adapters/IDB.js
@@ -155,13 +155,22 @@ const cursorHandlers = {
  * @return {IDBRequest}
  */
 function createListRequest(cid, store, filters, done) {
+  const filterFields = Object.keys(filters);
+
+  // If no filters, get all results in one bulk.
+  if (filterFields.length == 0) {
+    const request = store.index("cid").getAll(IDBKeyRange.only(cid));
+    request.onsuccess = event => done(event.target.result);
+    return request;
+  }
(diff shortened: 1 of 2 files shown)


leplatrem:
Here is my test with 1K records:

https://bl.ocks.org/leplatrem/1c9a45ef92f2f56b369e87bf1db4548d/d500116b8062faef38717550f5c7d12df446ba12

## Before 

### with Firefox

sync: 113ms
**list all: 324ms**
filter by id: 1ms
filter by attribute: 209ms

### with Chrome

sync: 114.10498046875ms
list all: 84.248046875ms
filter by id: 1.52001953125ms
filter by attribute: 48.302978515625ms

## After

### with Firefox

sync: 154ms 
**list all: 21ms**
get by id: 2ms
get by attribute: 239ms

### with Chrome

sync: 77.898193359375ms
list all: 66.27490234375ms
get by id: 2.4287109375ms
get by attribute: 65.539794921875ms

## Conclusion

It makes no big difference with Chrome. But more than 10X on Firefox.

PR #866, opened 2018-10-17
Ref #862: use getAll() with indexed fields too

leplatrem:
I realized that in #865 I could also have gone further and use `.getAll()` when no filter is used on custom field.
For example, when we gather local changes we want all records whose is status is not "sync".


leplatrem:
> Feels a little non-DRY

There is enough variation between the two to justify the duplication IMO

PR #868, opened 2018-10-18
IDB: remove redundant index on `id` field

leplatrem:


glasserc:
Refs https://bugzilla.mozilla.org/show_bug.cgi?id=1486980#c37.

Issue #870, opened 2018-10-31
IDB insertions performance


leplatrem:
Iterating on list in order to perform massive IndexedDB insertions is not efficient.

We should chain insertions instead for the browser to schedule the insertions better.

leplatrem:
For the bulk insert in `Collection.importChanges`, maybe we should iterate in chunks instead. Modifying the storage adapter interface is less obvious. But maybe we could rely on `loadDump()` there too :) I'll investigate

PR #871, opened 2018-10-31
Ref #870: better bulk insert for IDB loadDump()

leplatrem:
Ref #870 

This doesn't entirely fix the #870 since we have another place (`Collection.importChanges()`) where the «chaining» could be leveraged. However, because we have an abstraction around storage adapters, we can't do it the same way as for loadDump.

In Gecko RemoteSettings we have a use-case where the dump contains around 1K entries. This PR should help greatly. However we have another use-case where the server contains 1K entries but no dump. For that one, the bottleneck will be in `importChanges`. I propose that we solve it in another PR after we've seen some improvements with this one.

### DIFF
--- a/src/adapters/IDB.js
+++ b/src/adapters/IDB.js
@@ -534,8 +534,10 @@ export default class IDB extends BaseAdapter {
   async loadDump(records) {
     try {
       await this.execute(transaction => {
+        // Since the put operations are asynchronous, we chain
+        // them together. The last one will be waited for the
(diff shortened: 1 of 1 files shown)


leplatrem:
Let's wait for confirmation from https://bugzilla.mozilla.org/show_bug.cgi?id=1502146 

> I have to say this change is a bit counterintuitive

Before we were inserting many records without waiting for the browser to finish. That probably means we kind of hijack its ability to schedule the operations.

Issue #872, opened 2018-11-01
Rename loadDump() to importBulk()? 


leplatrem:
As @peterbe commented, the wording `loadDump()` is not entirely intuitive. Load let us think it loads into memory or something.

I suggest to we rename the method to `importBulk()` (and add deprecation to the old one)

zakaluka:
I've put up a first attempt at this change - #875

maniac-tech:
I see that the solution @zakaluka had written has been accepted and merged into master. In that case, @leplatrem can you update the Issue, and close it?

leplatrem:
Oh yes! thank you for the heads up!

PR #875, opened 2018-11-07
First attempt at changing loadDump to importBulk

zakaluka:


zakaluka:
For Issue #872

docs/api.md
```diff
@@ -328,7 +328,7 @@ actually starting the first sync with it.
 The list of imported records is returned.
 
 ```js
-articles.loadDump([
+articles.importBulk([ // Previously, loadDump()
```
glasserc: I don't know if we need to call out specifically the older name of the method..

docs/api.md
```diff
@@ -340,6 +340,7 @@ articles.loadDump([
 
 > #### Notes
 >
+> - The use of `loadDump()` is deprecated.  Use `importBulk()` instead.
```
glasserc: Similarly, I'm not sure we need to add a note about a method that isn't documented here.

test/adapters/IDB_test.js
```diff
@@ -474,8 +493,11 @@ describe("adapter.IDB", () => {
     });
   });
 
-  /** @test {IDB#loadDump} */
-  describe("#loadDump", () => {
+  /** 
+   * @deprecated
+   * @test {IDB#loadDump} 
+   */
+  describe("Deprecated #loadDump", () => {
```
glasserc: I don't think we need the old tests for `loadDump`. Just add a test that it calls `importBulk`.

test/collection_test.js
```diff
@@ -1456,8 +1456,11 @@ describe("Collection", () => {
     });
   });
 
-  /** @test {Collection#loadDump} */
-  describe("#loadDump", () => {
+  /** 
+   * @deprecated
+   * @test {Collection#loadDump} 
+   */
+  describe("Deprecated #loadDump", () => {
```
glasserc: Same comment about how thoroughly we need/don't need to test `loadDump`.

test/adapters/IDB_test.js
zakaluka: I'm not exactly sure how to test that `loadDump` is calling `importBulk`.  Do you have any suggestions on how to do that?

test/collection_test.js
zakaluka: Same question as above.  Should I use something like [chai-spies](https://www.chaijs.com/plugins/chai-spies/)?

test/adapters/IDB_test.js
zakaluka: Ok, I figured out that I have to use `sinon.spy()` or one of its variations.  Now trying to figure out how to do that...

zakaluka:
Changing the check from `sandbox.spy()` to `should.be.a('promise')` appears to be working.  The failure seems to be related to `prettier`, but I'm not sure how to fix this.

beatrochee:
hey @zakaluka, played around with it for a bit and it looks like if you run "npm run cs-format" locally, it'll format your code in the required way for you! Running "npm run cs-check" after that confirms that it did fix those changed.

test/adapters/IDB_test.js
```diff
@@ -411,8 +411,26 @@ describe("adapter.IDB", () => {
...
+    it("should call importBulk", () => {
+      sandbox.stub(db, "prepare").callsFake(async (name, callback, options) => {
+        return callback({
+          put() {
+            throw new Error("transaction error");
+          },
+        });
+      });
+      return db.loadDump([{ foo: "bar" }]).should.be.a("promise");
+      //return ib.should.have.been.calledOnce;
```
glasserc: This test just checks that a promise was returned. I think `ib.should.have.been.calledOnce` seems more like a test that it called `importBulk`. There are a bunch of tests in `tests/collection_test.js` that do what I think you should do here -- just grep for `sinon.assert.calledOnce` or `sinon.assert.calledWithExactly`. I think you can stub `db.importBulk` instead of `db.prepare`. The return value of `stub()` is the stub/spy that you can use with `calledOnce`.

zakaluka: I have changed both tests to stub and confirm that `importBulk` is being called.

Issue #802, opened 2018-03-08
Expose last_modifieds in the "server was flushed" exception


glasserc:
Bugs like https://bugzilla.mozilla.org/show_bug.cgi?id=1416255 could be made a bit easier if we expose the last_modified we had, as well as the last_modified the server had, when we raise the "server was flushed" exception.

lexcorp16:
Hi @glasserc I would like to take a look at this. This is my first attempt to contribute to open source and I would be glad if you could point me in the right directions. Thanks

glasserc:
Hi @lexcorp16, thanks for your interest! kinto.js implements an algorithm to "sync" data to and from the server. It works by tracking what has changed on the client and sending that, and asking what has changed on the server and receiving that. The main entry point is the `sync()` method in collection.js (https://github.com/Kinto/kinto.js/blob/76246b3c9abfc6fb668d747c4ca2755767cbaa85/src/collection.js#L1193).

In the internals of this method, sometimes we encounter a situation where the server tells us that nothing has changed, but responds in a way that indicates that something *has* changed. This can happen on the "dev" server (described at http://kinto.readthedocs.io/en/stable/tutorials/install.html#run-kinto-mozilla-demo), which gets wiped every morning. When it gets wiped, it will tell the client that there are no records which have changed (because there are no records), but the server will still seem "updated" (because it was wiped).

We throw an exception when this happens. I'm not sure exactly where this behavior originated but I've heard people say it's to try to help them debug strange syncing behavior on the dev server (because once data gets wiped, it's hard to "sync" anything). It's also helped us spot data loss issues in the Kinto server (for instance Kinto/kinto#1209). The exact line is:

https://github.com/Kinto/kinto.js/blob/76246b3c9abfc6fb668d747c4ca2755767cbaa85/src/collection.js#L975

We sometimes see this error in the chrome.storage.sync API in Firefox, which is implemented using this library. I'm concerned that it represents another data loss bug, or something even stranger, in the Kinto server, but it's hard to debug because all it says is "Server was flushed". I was thinking that it would be good for the error message to include some of the facts that led kinto.js to throwing this exception. In particular, I guess the lastModified from the client as well as the one the server responded with would be good. At least in the chrome.storage.sync case, this might indicate that an error recovery path was working, or not working.

Hope this helps! Feel free to ask on IRC or Slack if you need more help! (I'm usually on during the week.)

neuroleptic:
[...]

PR #878, opened 2018-11-27
fixed issue 802 by adding lastModified in error statement

nobellium1997:
Super simple fix to a very simple issue listed. Thought this might be a good first try at contributing to open source.

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -991,12 +991,15 @@ export default class Collection {
     if (!options.exclude && localSynced && serverChanged && emptyCollection) {
       // throw Error("Server has been flushed. Client Side Timestamp: " + localSynced + " Server Side Timestamp: "
       // + unquoted);
-        let e = new Error;
-        e.message = "Server has been flushed. Client Side Timestamp: " + localSynced + " Server Side Timestamp: "
-        + unquoted;
-        e.clientTimestamp = localSynced;
-        e.serverTimestamp = unquoted;
-        throw e;
+      let e = new Error();
+      e.message =
+        "Server has been flushed. Client Side Timestamp: " +
+        localSynced +
+        " Server Side Timestamp: " +
+        unquoted;
+      e.clientTimestamp = localSynced;
+      e.serverTimestamp = unquoted;
+      throw e;
     }
 
     syncResultObject.lastModified = unquoted;
--- a/src/collection.js
+++ b/src/collection.js
@@ -989,7 +989,7 @@ export default class Collection {
     const serverChanged = unquoted > options.lastModified;
     const emptyCollection = data.length === 0;
     if (!options.exclude && localSynced && serverChanged && emptyCollection) {
-      throw Error("Server has been flushed." + since);
+      throw Error("Server has been flushed. Client Side Timestamp: " + localSynced + " Server Side Timestamp:  " + serverChanged);
     }
 
     syncResultObject.lastModified = unquoted;
--- a/src/collection.js
+++ b/src/collection.js
@@ -109,7 +109,7 @@ export class SyncResultObject {
   }
 }
 
-class ServerWasFlushedError extends Error {
+export class ServerWasFlushedError extends Error {
   constructor(clientTimestamp, serverTimestamp, message) {
     super(message);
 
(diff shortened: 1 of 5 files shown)


Natim:
refs #802

src/collection.js
```diff
@@ -989,7 +989,7 @@ export default class Collection {
     const serverChanged = unquoted > options.lastModified;
     const emptyCollection = data.length === 0;
     if (!options.exclude && localSynced && serverChanged && emptyCollection) {
-      throw Error("Server has been flushed.");
+      throw Error("Server has been flushed." + since);
```
Natim: Thank you for your proposal. That's a good start.

However this is a bit too simple, no space between the dot and the value, we don't actually know what the since value is while reading the message. Is it the server timestamp or the client timestamp?

> I guess the lastModified from the client as well as the one the server responded with would be good.

Can you enhance your PR by adding commits to it and display both the server side timestamp and the client side one?

I would also recommend adding a test for it.

Natim: The test you are looking to contribute to is here: https://github.com/Kinto/kinto.js/blob/571b0aeb2cc69f67d86523133d46b332cca5b3ae/test/integration_test.js#L2001

glasserc:
It also wouldn't be bad to expose the two values as properties of the error, although this is a bit more work!

src/collection.js
nobellium1997: Hi, so I fixed the issue with the client and the server timestamp, how would I go about matching the error to the error message in the integration_test file?

glasserc: You can use a regular expression instead of a literal string to match the error message.

nobellium1997:
Hey guys, so I added regex to the integration_test file so that it matches on the phrase and the timestamps. I also added the new error message with both the server and client timestamps.

Natim:
Thanks the test looks great. I agree with @glasserc that it would be great to also add the ``serverTimestamp`` and ``localTimestamp`` properties to the ``Error`` object so that we can use them  if we catch the Error in some code.

Natim:
Thanks to the test we can see that ``serverChanged`` is not the server timestamp. We need to find it somewhere else apparently.

nobellium1997:
Hey guys, so I just had a few questions. When you mean add the properties to the error object, where exactly should I be adding those properties? Looks like an error object only has a name, message, line number, and description so I'm not 100% sure where I would put the two properties. Also, I think the server timestamp is the variable named "unquoted" on line 983 in src/collection.js since it determines whether serverChanged is true or not.

nobellium1997:
I can definitely work on generating a subclass, I'm on break so I have time :smile:. One more question, I can't get the integration tests to run when I run npm run test, for some reason it keeps saying something like "before all hook" followed by ncaught Error: spawn pserve ENOENT and I was wondering if there was an obvious fix. I tried looking online but no luck so far.

glasserc:
The integration tests try to start `kinto`, so make sure you have it in your `PATH`. I usually get that error message when I forgot to activate my virtualenv. (Of course, if setting up `kinto` locally is too much work, it's fine to just rely on travis.)

nobellium1997:
Hey guys, I fixed the typo and made the client timestamp and server timestamp part of the error properties. I'll go ahead and start working on making a new error class now.

nobellium1997:
Hey everyone, is this pr still in review? I think it should be good to go

Natim:
It looks good to me. r? @glasserc @leplatrem ?

src/collection.js
```diff
@@ -989,7 +1002,17 @@ export default class Collection {
     const serverChanged = unquoted > options.lastModified;
     const emptyCollection = data.length === 0;
     if (!options.exclude && localSynced && serverChanged && emptyCollection) {
-      throw Error("Server has been flushed.");
+      // throw Error("Server has been flushed. Client Side Timestamp: " + localSynced + " Server Side Timestamp: "
+      // + unquoted);
```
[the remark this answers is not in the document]
nobellium1997: Hey all, just removed it. Totally forgot I left that there

nobellium1997:
Alrighty, I just removed the commented code so hopefully it should be good to go!

glasserc:
Thanks for your patience :)

PR #881, opened 2018-12-10
Fix dependencies vulnerabilities.

Natim:


package-lock.json
```diff
@@ -3734,7 +3738,7 @@
         },
         "util": {
           "version": "0.10.3",
-          "resolved": "https://registry.npmjs.org/util/-/util-0.10.3.tgz",
+          "resolved": "http://registry.npmjs.org/util/-/util-0.10.3.tgz",
```
Natim: I've no idea why on earth we would do that but since the file is automatically generated I have no clue.

Issue #558, opened 2016-09-21
Provide a «key-value store» ID schema


leplatrem:
The default id schema generates uuid. We could provide another one, that would be useful for when using kinto.js as a key value store:
- its `generate()` method throws error
- its `validate()` would always return true

d1ndra:
I'd like to work on this if anybody isn't. Any help regarding where I should start?

glasserc:
Go for it! We usually advise a somewhat TDD-based workflow where you write tests that demonstrate the API you would like to support, followed by implementations that make the tests pass. I think the best API would be one that exposed some ID schemas in the same module as whatever else most users import. There's already a UUID schema that could get exported at the same time.

Jonathansw:
Is this still open to be worked on?

glasserc:
Yes! I think the approach I wrote about above is still valid, although I should withdraw the remark that there's already a UUID schema, since the one that is there relies on Firefox internals.

AndresMurguido001:
If this is still open id like to give it a shot. But a little more info would be helpful. If im understanding this right, there should be an option to get and set keys like a hashtable and just use autoIncrement?

aqumus:
@glasserc Would be good to close this issue if the merged PR is working as expected

PR #944, opened 2019-03-26
Provide key-value store id schema(#558)

AndresMurguido001:
Id like to help with this issue. I followed the advice posted in the original issue just to start. Im wondering what exactly is needed to resolve this. I only added the tests to figure out the direction i should be heading. Any advice would be helpful. Thanks !

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -140,16 +140,14 @@ function createUUIDSchema() {
  * This will be the property used to retrieve this record.
  *
  * @example
- * const exampleCollection = db.collection("example", { idSchema: createKeyValueIdSchema() })
+ * const exampleCollection = db.collection("example", { idSchema: createKeyValueStoreIdSchema() })
  * await exampleCollection.create({ title: "How to tie a tie", favoriteColor: "blue", id: "user123" }, { useRecordId: true })
  * await exampleCollection.getAny("user123")
- *
  */
-
-export function createKeyValueIdSchema() {
+export function createKeyValueStoreIdSchema() {
   return {
     generate() {
-      throw new Error("createKeyValueIdSchema() does not generate an id");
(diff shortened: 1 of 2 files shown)


glasserc:
Refs #558. Great! Sorry I didn't respond to your question on the issue but it seems like you're on the right track. The goal is to export another idSchema similar to `createUUIDSchema`. So I think everything is correct about your patch except where you put the schema -- it shouldn't be in the tests, but maybe in the same place as `createUUIDSchema`. Also it would be good for there to be a docstring saying something about how this might be useful if you don't want IDs to be generated and giving a use case.

AndresMurguido001:
Thanks for the feedback! I removed it from the tests file and added everything you mentioned. I hope this helps!

src/collection.js
```diff
@@ -134,6 +134,29 @@ function createUUIDSchema() {
...
+ * Using this IDSchema requires you to set a property as the id.
+ * This will be the property used to retrieve this record.
+ *
+ * @example
+ * const exampleCollection = db.collection("example", { idSchema: createKeyValueIdSchema() })
+ * await exampleCollection.create({ title: "How to tie a tie", favoriteColor: "blue", id: "user123" }, { useRecordId: true })
+ * await exampleCollection.getAny("user123")
+ *
+ */
+
```
glasserc: I think we can get rid of this extra blank line.

```suggestion
```

src/collection.js
```diff
@@ -134,6 +134,29 @@ function createUUIDSchema() {
...
+export function createKeyValueIdSchema() {
```
glasserc: Nit: can we call it maybe `createKeyValueStoreIdSchema`? This is a term of art with a specific meaning, see https://en.wikipedia.org/wiki/Key-value_database.

Issue #816, opened 2018-05-03
Support sub-object filtering when listing records


leplatrem:
```js
const subset = await collection.list({ filters: { "person.age": 23 } })
```

:)

AndresMurguido001:
Is this issue still open?

leplatrem:
Yes!

Thanks for your interest!

PR #946, opened 2019-03-31
Sub-Object Filtering(#816)

AndresMurguido001:
I think this is a step in the right direction. What are your thoughts?

### DIFF
--- a/src/adapters/IDB.js
+++ b/src/adapters/IDB.js
@@ -139,7 +139,7 @@ const cursorHandlers = {
         cursor.continue(values[i]);
       }
     };
-  }
+  },
 };
 
 const transformSubObjectFilters = filtersObj => {
@@ -292,7 +292,7 @@ export default class IDB extends BaseAdapter {
         const db = event.target.result;
         // Records store
         const recordsStore = db.createObjectStore("records", {
-          keyPath: ["_cid", "id"]
+          keyPath: ["_cid", "id"],
         });
         // An index to obtain all the records in a collection.
         recordsStore.createIndex("cid", "_cid");
@@ -303,9 +303,9 @@ export default class IDB extends BaseAdapter {
         recordsStore.createIndex("last_modified", ["_cid", "last_modified"]);
         // Timestamps store
         db.createObjectStore("timestamps", {
-          keyPath: "cid"
+          keyPath: "cid",
         });
-      }
+      },
     });
 
     if (dataToMigrate) {
@@ -633,7 +633,7 @@ function transactionProxy(adapter, store, preloaded = []) {
 
     get(id) {
       return preloaded[id];
-    }
+    },
   };
 }
 
@@ -648,7 +648,7 @@ async function migrationRequired(dbName) {
     version: 1,
     onupgradeneeded: event => {
       exists = false;
-    }
+    },
   });
 
   // Check that the DB we're looking at is really a legacy one,
--- a/src/adapters/IDB.js
+++ b/src/adapters/IDB.js
@@ -155,7 +155,6 @@ const cursorHandlers = {
  * @return {IDBRequest}
  */
 function createListRequest(cid, store, filters, done) {
-  // console.log("FILTERS: ", filters);
   const filterFields = Object.keys(filters);
 
   // If no filters, get all results in one bulk.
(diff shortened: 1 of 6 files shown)


test/collection_test.js
```diff
@@ -1443,6 +1443,32 @@ describe("Collection", () => {
           .then(res => res.data.map(r => r.title))
           .should.eventually.become(["art1"]);
       });
+
+      it("should support sub-object filtering", () => {
+        return articles
+          .list({
+            filters: { "article.complete": true },
```
leplatrem: I don't think you can test the sub-object query with the entries of this test, since they don't have sub-objects in their fields.

You would require an objects like this one:
```js
{
  title: "art1",
  author: {
    name: "alice",
  }
}
```
And the query would be `{ filters: { "author.name": "alice" } }`

src/adapters/IDB.js
```diff
@@ -171,11 +171,28 @@ function createListRequest(cid, store, filters, done) {
...
+    if (isSubQuery) {
+      const newFilter = {};
+      Object.entries(filters).map(filter => {
+        filter.reduce((pv, cv) => {
+          if (pv.includes(".")) {
+            newFilter[pv.split(".")[1]] = cv;
+          } else {
+            newFilter[pv] = cv;
+          }
+        });
```
leplatrem: I don't think the `reduce()` is used correctly here. It requires [an initial value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce) and the function must return something.

Also, this implementation works with one level only, whereas ideally I think it should support filtering like `{ filters:  { "author.address.city": "barcelona" } }`

AndresMurguido001:
Thanks for the advice! It really helped a lot. I think this is closer to what you had in mind. I know im missing a couple test cases, which ill be happy to add. Do you have an example of:

- entries where the sub-object field is not a dict (eg. title.lang)

Thats the one test im missing from above.

src/adapters/IDB.js
```diff
@@ -142,6 +142,27 @@ const cursorHandlers = {
...
+        : acc.hasOwnProperty(cv)
+        ? acc[cv]
+        : (acc[cv] = {});
+    }, newFilter);
+    return newFilter;
+  };
+  for (const key in filtersObj) {
+    const keysArr = key.split(".");
+    const val = filtersObj[key];
+    makeNestedObjectFromArr(keysArr, val);
```
leplatrem: It's better if you don't mix scopes (for example here `newFilter` is modified outside this loop). You should take advantage of the fact the `reduce()` returns the result.

For example:
```js
for (const key in filtersObj) {
  const keysArr = key.split(".");
  const topLevel = keysArr[0];
  const val = filtersObj[key];
  filtersObj[topLevel] = makeNestedObjectFromArr(filtersObj[topLevel], keysArr, val);
}
```

src/adapters/IDB.js
```diff
@@ -142,6 +142,27 @@ const cursorHandlers = {
   },
 };
 
+const transformSubObjectFilters = filtersObj => {
+  const newFilter = {};
+  const makeNestedObjectFromArr = (arr, val) => {
+    const last = arr.length - 1;
+    arr.reduce((acc, cv, i) => {
+      return i === last
+        ? (acc[cv] = val)
```
leplatrem: It's not a recommended practice to put assignments in ternary operators.

```js
if (i === last) {
  acc[cv] = val;
} else if ...
```
is better

AndresMurguido001:
> ```js
> record = {
>   name: "bob",
>   age: 42,
> };
> ```
> 
> `filters = { "age.bad": 12 }`


Should something like this throw an error? As of right now, if a property doesnt exist, no results are returned and an error is printed to the console that includes the missing key.

leplatrem:
> Should something like this throw an error?

No I don't think so. Some `console.error()` is enough :)

AndresMurguido001:
Thanks a lot for the help. I think I fixed all the suggestions you mentioned above, but if i missed something or if i need to add anything just let me know. Happy to help👍

test/collection_test.js
```diff
@@ -1445,6 +1445,83 @@ describe("Collection", () => {
...
+          })
+          .then(res => {
+            return res.data.map(r => {
+              return r.title;
+            });
+          })
+          .should.eventually.become(["art3"]);
+      });
+
+      it("should throw on records missing subObject field", () => {
```
leplatrem: It does not throw, we should adjust the test title :)

AndresMurguido001:
No problem at all. I appreciate all the advice and guidance. Im just happy to contribute.

glasserc:
Thanks!

PR #972, opened 2019-05-06
Fix header in Gecko export

leplatrem:


glasserc:
This appears to have gotten changed in https://bugzilla.mozilla.org/show_bug.cgi?id=1514594.

PR #995, opened 2019-06-21
Optimize sync result code

leplatrem:
Before:
https://perfht.ml/2Xthz1c

After:
https://perfht.ml/2WVRfsq

--> Time spent in `importChanges()` becomes ridiculous.

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -140,6 +140,21 @@ export class SyncResultObject {
     delete this._cached[type];
     return this;
   }
+
+  toObject() {
+    return {
+      ok: this.ok,
+      lastModified: this.lastModified,
+      errors: this.errors,
+      created: this.created,
+      updated: this.updated,
+      deleted: this.deleted,
+      skipped: this.skipped,
+      published: this.published,
+      conflicts: this.conflicts,
+      resolved: this.resolved,
+    };
+  }
 }
 
 export class ServerWasFlushedError extends Error {
--- a/src/collection.js
+++ b/src/collection.js
@@ -143,6 +143,7 @@ export class SyncResultObject {
   }
 
   toObject() {
+    // Only used in tests.
     return {
       ok: this.ok,
       lastModified: this.lastModified,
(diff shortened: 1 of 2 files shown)


src/collection.js
```diff
@@ -35,8 +35,6 @@ export class SyncResultObject {
    */
   static get defaults() {
```
glasserc: Can we rename this to something that is clearly list-specific?

test/collection_test.js
```diff
@@ -2266,6 +2281,18 @@ describe("Collection", () => {
...
+            lastModified: result.lastModified,
+            errors: result.errors,
+            created: result.created,
+            updated: result.updated,
+            deleted: result.deleted,
+            skipped: result.skipped,
+            published: result.published,
+            conflicts: result.conflicts,
+            resolved: result.resolved,
+          }))
```
glasserc: Maybe we should extract this as a utility function, something like `reifySyncResult` ?

glasserc:
Are these profiles the right ones? I remember you showing me one that has lots of `addOneRecord` calls, but this one doesn't seem to have any..

leplatrem:
> Are these profiles the right ones? I remember you showing me one that has lots of addOneRecord calls, but this one doesn't seem to have any..

You're right... they don't show the right info.

Here is a new try:

Before https://perfht.ml/2ZN8M7E 
After https://perfht.ml/2FxY6SP

src/collection.js
```diff
@@ -69,44 +61,100 @@ export class SyncResultObject {
...
    */
   reset(type) {
-    this[type] = SyncResultObject.defaults[type];
-    this.ok = this.errors.length + this.conflicts.length === 0;
+    this._lists[type] = [];
+    delete this._cached[type];
     return this;
   }
+
+  toObject() {
```
glasserc: Maybe add a comment that this is just for testing?

leplatrem:
> I guess we no longer deduplicate errors or conflicts, although I imagine that isn't really important.

I don't think we did before:

https://github.com/Kinto/kinto.js/blob/ad789ca33256450c98e30dbff7fb6b3914117a0a/src/collection.js#L78-L79

PR #996, opened 2019-06-21
Import changes in batches

leplatrem:
Before
https://perfht.ml/2WVRfsq

After
https://perfht.ml/2WWcfz1

### DIFF
--- a/src/adapters/IDB.js
+++ b/src/adapters/IDB.js
@@ -10,6 +10,7 @@ import {
 } from "../utils";
 
 const INDEXED_FIELDS = ["id", "_status", "last_modified"];
+const MAX_PRELOAD = 25;
 
 /**
  * Small helper that wraps the opening of an IndexedDB into a Promise.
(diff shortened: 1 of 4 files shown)


glasserc:
The second profile seems to show that less time is being used overall. I think this could be due to blocking less on I/O?

leplatrem:
In this profile graph, we can see the situation with `.in()` before.
https://perfht.ml/2FvsBcc  (16s total)

By just *moving the i variable to a slightly higher scope so that we resume where we left off?*, here is it what we have:
https://perfht.ml/2ZMp5Se
A lot better (cf `self time` of `.in()`) and 3s total!

After adding the batch patch, it becomes this:
https://perfht.ml/2ZM3zgs
It seems better in terms of UI blocking (red line on top). But 6s total.

glasserc:
I sort of expected the "batches" thing to increase amount of work (more round trips) but decrease jank, so that doesn't surprise me. (I don't have an opinion on the amount of additional work that is acceptable.)

Using the "highlight" tool, it seems to me like the majority of kinto-offline-client work in the first profile is within [7.5s-9.2s](https://profiler.firefox.com/public/6ba2ca98e4a78d63a8feede4f0ad455ee60227d4/calltree/?globalTrackOrder=0-1-2-3-4&hiddenGlobalTracks=1-2-4&implementation=js&localTrackOrderByPid=5819-1-0~5847-0~&range=7.4946_9.2272&thread=0&v=3), and in the second within [5.3s-8.5s](https://profiler.firefox.com/public/30eb0ddd9c998bc93f1171efefcef31deaa46504/calltree/?globalTrackOrder=1-0-2-3-4&hiddenGlobalTracks=1-4&implementation=js&localTrackOrderByPid=7571-1-0~7599-0~&range=5.1711_8.5237&thread=0&v=3). On the other hand, that's still like a 2x slowdown. I thought the `slice` might be part of the cost, but it's hard to justify -- there's a little bit of extra GC time but not that much.

PR #997, opened 2019-06-21
Add a PULL_ONLY strategy

leplatrem:
- [x] POC
- [x] Update docs
- [x] See if some conflicts detection/resolution stuff can be bypassed

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -391,6 +391,7 @@ export default class Collection {
     return {
       CLIENT_WINS: "client_wins",
       SERVER_WINS: "server_wins",
+      PULL_ONLY: "pull_only",
       MANUAL: "manual",
     };
   }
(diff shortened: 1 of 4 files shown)


src/collection.js
```diff
@@ -248,7 +249,7 @@ function importChange(transaction, remote, localFields) {
     if (local._status === "deleted") {
       return { type: "skipped", data: local };
     }
-    if (isIdentical) {
+    if (isIdentical || strategy === Collection.strategy.PULL_ONLY) {
```
glasserc: I think we can even skip the check to see if records are identical if we're in `PULL_ONLY`. Not sure it's worth implementing though..

test/integration_test.js
```diff
@@ -354,6 +354,53 @@ describe("Integration tests", function() {
...
+          });
+
+          it("should have a valid lastModified value", () => {
+            expect(syncResult.lastModified).to.be.a("number");
+          });
+
+          it("should have no conflicts", () => {
+            expect(syncResult.conflicts).to.have.length.of(0);
+            expect(syncResult.resolved).to.have.length.of(0);
+            expect(syncResult.updated).to.have.length.of(1);
```
glasserc: `expect(syncResult.published).to.have.length.of(0)`?

src/collection.js
leplatrem: Well, we must now if the record was there or if it was updated though. Maybe just compare timestamps instead?

glasserc: I don't understand. According to the code you wrote, we don't actually need to know if the record was updated.

leplatrem: Actually, I was confused because the integration tests were implemented in a weird way (synced records were manually created on the server and the client, I replaced this by a sync). After rewroting the test setup I could skip the records comparison as expected :) Thanks for suggestion!

PR #1038, opened 2019-08-08
Replace Browserify and Babel with TypeScript and Rollup

dstaley:
This is largely a replication of the work done to migrate [kinto-http.js to TypeScript and Rollup](https://github.com/Kinto/kinto-http.js/pull/433). It makes the same modifications to `package.json` and uses essentially the exact same Rollup config.

There are, however, two small changes specific to this repo:
1. Fixed a bug in `test/adapters/IDB_test.js` that could cause an index out-of-bounds error.
1. Added a small Rollup plugin to exclude `uuid` from the Gecko build.
1. Updated the `cs-check` command to use prettier's `--check` option. I think this is why the builds have been green despite the test files not matching the prettier config.

Unfortunately, it looks like some of the test files didn't follow the Prettier style, so those changes also come along in this PR. The only other changes in the test files (besides the trailing commas) is the change of import for `uuid` to match how it's being imported in the main library (`import uuid from 'uuid/v4'`

**Note:** As of right now the `5.0.0-alpha.0` version of `kinto-http` has a small error. If this note is still here and you want to test the new build process, please replace the following line in `node_modules/kinto-http/package.json`:
> `"module": "lib/es/index.js",`

with

> `"module": "lib/esm/index.js",`


.gitignore
```diff
@@ -6,3 +6,5 @@
 /esdoc
 /node_modules
 /demo/kinto*
+.DS_Store
+.nyc_output
```
Natim:
```suggestion
.nyc_output

```

dstaley:
Closing this PR since I've managed to anger the Git gods. New PR incoming!

PR #1044, opened 2019-08-14
Fix #1043: Reject when transaction is aborted

### LINKED ISSUE #1043: IDB.js library code doesn't handle transaction abort events
Similar to how IndexedDB.jsm in the firefox tree needed to be updated to properly translate "abort" events to reject promises in https://bugzilla.mozilla.org/show_bug.cgi?id=1542181 in https://phabricator.services.mozilla.com/D36822, the code around https://github.com/Kinto/kinto.js/blob/add721282c69418e3af1651695efc8d249a55116/src/adapters/IDB.js#L76 needs to handle the abort event.

The general context in which this is a problem is when Firefox/Gecko is shutting down and QuotaManager shuts down but code like kinto and RemoteSettings are still active.  QuotaManager will abort the outstanding transaction without necessarily firing an error event.

leplatrem:
Fix #1043

@asutherland is it required on other requests as well? like database open/delete etc?

### DIFF
--- a/src/adapters/IDB.js
+++ b/src/adapters/IDB.js
@@ -74,7 +74,13 @@ export async function execute(db, name, callback, options = {}) {
     }
     transaction.onerror = event => reject(event.target.error);
     transaction.oncomplete = event => resolve(result);
-    transaction.onabort = () => reject(transaction.error);
+    transaction.onabort = event => {
+      const error =
+        event.target.error ||
+        transaction.error ||
+        new DOMException("The operation has been aborted", "AbortError");
+      reject(error);
+    };
   });
 }
 
--- a/src/adapters/IDB.js
+++ b/src/adapters/IDB.js
@@ -26,6 +26,16 @@ export async function open(dbname, { version, onupgradeneeded }) {
     request.onupgradeneeded = event => {
       const db = event.target.result;
       db.onerror = event => reject(event.target.error);
+      // When an upgrade is needed, a transaction is started.
+      const transaction = event.target.transaction;
+      transaction.onabort = event => {
+        const error =
+          event.target.error ||
+          transaction.error ||
+          new DOMException("The operation has been aborted", "AbortError");
+        reject(error);
+      };
+      // Callback for store creation etc.
       return onupgradeneeded(event);
     };
     request.onerror = event => {
--- a/src/adapters/IDB.js
+++ b/src/adapters/IDB.js
@@ -74,6 +74,7 @@ export async function execute(db, name, callback, options = {}) {
     }
     transaction.onerror = event => reject(event.target.error);
     transaction.oncomplete = event => resolve(result);
+    transaction.onabort = () => reject(transaction.error);
   });
 }
 


src/adapters/IDB.js
```diff
@@ -74,6 +74,7 @@ export async function execute(db, name, callback, options = {}) {
     }
     transaction.onerror = event => reject(event.target.error);
     transaction.oncomplete = event => resolve(result);
+    transaction.onabort = () => reject(transaction.error);
```
Natim: This property is null if the transactionwas aborted with the IDBTransaction.abort method.

[source](https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction/error)

Natim: > All pending IDBRequest objects created during this transaction have their IDBRequest.error attribute set to AbortError.

[source](https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction/abort)

leplatrem: Oh, interesting! That's how they did it in https://phabricator.services.mozilla.com/D36822  I didn't even check the docs...

Natim: However I have no idea how to grab the IDBRequest at this point.
Maybe the error is linked to the event:

```suggestion
    transaction.onabort = () => reject(event.target.error);
```

leplatrem: Well, like the error then `transaction.onabort = event => reject(event.target.error);`

Natim: We shall try both and see what happens.

rpl: @leplatrem @Natim I just took another look to what is the error rejected (if any) on the test case that was getting stuck without this fix applied in IndexedDB.jsm and in that case `transaction.error` was set to `QuotaExceededError`.

Then I tried to see what is happening when I explicitly reject the transaction using transaction abort and for me both `transaction.error` and `event.target.error` were not defined.

I haven't yet submitted a patch for IndexedDB.jsm on phabricator, but I'm thinking to resort to something like the following to be sure to always reject the transaction with an error:

```
transaction.onabort = (evt) => {
  let error =
    transaction.error ||
    evt.target.error ||
    new DOMException("The operation has been aborted", "AbortError");
  reject(error);
};
```

I'll keep an eye on this PR (and let you know once I have further updates about how we are going to handle this in the wrapper defined by IndexedDB.jsm).

rpl: (I just filed [Bug 1573828](https://bugzilla.mozilla.org/1573828) for tracking the issue for  IndexedDB.jsm)

asutherland:
@leplatrem Yes, it looks like IDB.js needs extra logic around the initial open request's upgrade transaction.  Specifically:
- `onabort` EventHandlers exist on the `IDBDatabase` and `IDBTransaction` interfaces.  When a transaction is aborted, including an upgrade transaction, the "abort" event is only ever fired at the transaction.  But the "abort" event is bubbling, and so if `stopPropagation()` (or equivalent) is not called on the event, it will bubble up to the database.
- In the event that the dispatch of the "upgradeneeded" event throws, abort a transaction will be run against the version upgrade transaction (per step 9.7 of https://w3c.github.io/IndexedDB/#run-an-upgrade-transaction).
- It doesn't look like IDB.js has wrapped the `transaction` found on the `IDBOpenDBRequest` interface (specifically found on the superclass `IDBRequest`) in order to process its abort event.  Note that the `transaction` will only become available during the dispatch of the "upgradeneeded" event, so that abort listener needs to be added inside there.  Alternately, one could add an "abort" event listener to the database, but it also only becomes available at that time as well, so it's probably a better idea to just add it to the transaction since that transaction will be short-lived whereas the database "abort" handler would hang around longer.

PR #1061, opened 2019-09-11
Replace Browserify and Babel with TypeScript and Rollup

dstaley:
_The content of this PR is the same as the previously closed PR. I must have done something to anger the Git gods because I could no longer merge upstream changes due to unrelated histories, so I just manually recreated this PR._

This is largely a replication of the work done to migrate [kinto-http.js to TypeScript and Rollup](https://github.com/Kinto/kinto-http.js/pull/433). It makes the same modifications to `package.json` and uses essentially the exact same Rollup config.

There are, however, two small changes specific to this repo:
1. Fixed a bug in `test/adapters/IDB_test.js` that could cause an index out-of-bounds error.
1. Added a small Rollup plugin to exclude `uuid` from the Gecko build.
1. Updated the `cs-check` command to use prettier's `--check` option. I think this is why the builds have been green despite the test files not matching the prettier config.

Unfortunately, it looks like some of the test files didn't follow the Prettier style, so those changes also come along in this PR. The only other changes in the test files (besides the trailing commas) is the change of import for `uuid` to match how it's being imported in the main library (`import uuid from 'uuid/v4'`


leplatrem:
@dstaley I'm sorry, this completely slipped through my inbox... 

Is it up-to-date?

dstaley:
@leplatrem It is now! Once this is merged can you do an alpha release on npm? I'd love to give that a try in my TypeScript project using Kinto.

leplatrem:
Just published `kinto@13.0.0-alpha.0` :)

PR #1120, opened 2019-11-19
Update build scripts to support Windows

dstaley:
This is an exact copy of [the PR made on `kinto-http.js`](https://github.com/Kinto/kinto-http.js/pull/520) to support building on Windows along with [the PR adding Windows CI tests](https://github.com/Kinto/kinto-http.js/pull/568). It also runs Prettier on tests since that wasn't being run before.


dstaley:
@leplatrem I figured it would be fine for me to merge this, but since this library is a bit more important than the HTTP one, I requested your review :)

Issue #832, opened 2018-08-14
Checkout idb library


leplatrem:
Some of our IndexedDB code could be replace by https://github.com/jakearchibald/idb

PR #1172, opened 2020-01-13
Convert to TypeScript

dstaley:
This PR converts the library to TypeScript. :tada:

For clarity, you can view a diff of the generated JavaScript [here](https://www.diffchecker.com/04dX8KGO). This reflects all changes that affect the runtime behavior (as opposed to changes that only impact TypeScript users). This is the Gecko build, so it doesn't contain `kinto-http.js`.

The vast majority of changes are just specifying the argument and return types for all the functions in the library. You can see that's the case by looking at the emitted JavaScript above. The only changes are switching from the abstract equality operator (`==`) to the more type-safe strict equality operator. There's also a few minor changes to reflect variables that are possibly undefined.

As far as the types go, I think I've got them mostly correct. I compared the types to the return values from running the library along with documentation. All the tests are passing, so I'm hoping that means they're correct.

That being said, I do have a few areas of concern:
1. The `waterfall` function is impossible to type in TypeScript due to the fact that it accepts an arbitrary list of functions. In lieu of typing the function directly, I've typed the functions that consume the output of `waterfall`. However, I don't think it's possible to type the `incoming-changes` hook since that's an arbitrary function. This is *probably* okay since TypeScript will allow anything by default.
1. `Collection` takes in a generic type `B` that defaults to `any`. This allows a consumer to specify the shape of their records, and TypeScript will then provide type-checking on all relevant operations. So when you see `B`, know that it stands for "the type of record, or `any` if not specified".
1. There's several places where you'll see `IDB` in lieu of the expected `BaseAdapter`. This is because a few places in `Collection` depended on properties defined on `IDB` but not `BaseAdapter`. I wanted this translation to reflect the current state of the library, so this is one area that I want to improve once this PR is merged.
1. The typings in `IDB.ts` are fairly loose since it's difficult to type an arbitrary data store. Also, the TypeScript definitions for IndexedDB aren't great. I think it might make sense to implement #832 since that might give us better types. Something to explore after this is merged.
1. There's still a good bit of `any` in `Collection`, which I want to continue to refine. However I don't feel that it's so much that it's not ready to be merged in.
1. The tests also have a good bit of `any`, but I figured that's okay since we're actually testing the code anyway. I didn't want to add a bunch of type declaration code to the tests when it's not useful.

TODO:
- [x] Provide a summary of the runtime changes
- [x] Point out areas of concern
- [x] Publish version of `kinto-http.js` with TypeScript declarations


dstaley:
@leplatrem okay, I think this is ready for a proper review now! I still want to refine the types for `SyncResult`, but I don't think that's big enough to block merging this in.

dstaley:
The Windows build is failing due to https://github.com/ultrajson/ultrajson/issues/369.

dstaley:
@leplatrem I'd love a quick look over this if you have the time! I have some availability coming up, and would love to get started switching this over to Intern so that we can test all the IndexedDB code in an actual browser.

src/types.ts
```diff
@@ -0,0 +1,132 @@
...
+  [key in AvailableHook]?: ((
+    record: any,
+    collection: Collection<T>
+  ) => Promise<{ changes: T[] }>)[];
+};
+
+export interface KintoRepresentation<T = unknown> {
+  data: T;
+  permissions: { [key in Permission]?: string[] };
+}
```
leplatrem, line 35: Don't we want to reuse kinto-http.js types for this?

dstaley: I'll take another look at where we're using `KintoRepresentation`, but technically the equivalent in `kinto-http.js` for this type has some additional required fields on the `data` property that I think we were missing in this library.

PR #1279, opened 2020-04-25
Run tests in browser using Intern

dstaley:
This PR builds on the work done in https://github.com/Kinto/kinto-http.js/pull/715 to support running the tests in real browsers!

Thankfully, we didn't need any major modifications to get this to work (apart from switching from `chai-as-promised` to `async`/`await`).


dstaley:
@leplatrem I'm good to merge this, but I did have a question about one test:

```ts
it("should delete unsynced virtually deleted local records", async () => {
  const locallyDeletedId = records[0].id;
  sandbox.stub(KintoClientCollection.prototype, "batch").returns(
    Promise.resolve({
      published: [{ data: { id: locallyDeletedId, deleted: true } }],
      errors: [],
      conflicts: [],
      skipped: [],
    })
  );
  // TODO: Is the following line important?
  // await articles.delete(locallyDeletedId);
  await articles.pushChanges(client, records, result);
  await expectAsyncError(
    () => articles.get(locallyDeletedId, { includeDeleted: true }),
    /not found/
  );
});
```

Calling `await articles.delete(locallyDeletedId)` results in a `not found` error, which was being picked up by Mocha as an expected failure. However, since we're now testing the _specific_ call that throws the error, we can see that we never reached the `articles.get(locallyDeletedId, { includeDeleted: true })` call. Since `articles.get(locallyDeletedId, { includeDeleted: true })` is throwing a `not found` error, is it safe to omit the call to `articles.delete(locallyDeletedId)`?

dstaley:
@leplatrem After playing with both the original test and the new one, I think this is a better expression of the ideas in this particular test:

```ts
it("should delete unsynced virtually deleted local records", async () => {
  const record = await articles.create({ title: "record to be deleted" });
  sandbox.stub(KintoClientCollection.prototype, "batch").returns(
    Promise.resolve({
      published: [{ data: { id: record.data.id, deleted: true } }],
      errors: [],
      conflicts: [],
      skipped: [],
    })
  );
  await articles.delete(record.data.id);
  await articles.pushChanges(client, records, result);
  await expectAsyncError(
    () => articles.get(record.data.id, { includeDeleted: true }),
    /not found/
  );
});
```

So, the logic is now as follows:

1. Create a record
2. Virtually delete the record.
3. Push the changes to the server. The server acknowledges the deletion of the record.
4. Confirm that the record has been _actually_ deleted.

Does this still cover the original intent of the test?

leplatrem:
Sorry, that it took me so many days again to catch up :/

I agree, your version of the test covers the idea !

dstaley:
@leplatrem No need to apologize! I appreciate you taking the time to confirm!

PR #1304, opened 2020-05-09
[semver major] Improve storage adapter support

dstaley:
This PR improves the typing support for adapters, decoupling the dependence on the IDB adapter.

This contains a **breaking change** in the way adapters are provided to the client. Instead of passing the class as the adapter property, callers will now need to pass a function that returns an instantiated instance of the adapter.

```ts
// Old
const client = new Kinto({ adapter: IDB });

// New
const client = new Kinto({ adapter: (cid, options) => new IDB(cid, options) }
```

This allows us to support arbitrary constructors for storage adapters.

This PR also includes a new in-memory storage adapter, intended only for documentation purposes. It's not exported in the bundle, so users won't have access to it. It serves as an easier-to-understand implementation so that people who want to write their own adapter don't need to understand the black magic voodoo that is IndexedDB. Just to ensure that the implementation is valid, it's included in our tests (although it doesn't currently have tests of its own, instead relying on the tests for `Collection`).

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -13,7 +13,7 @@ const db = new Kinto(options);
 - `remote`: The remote Kinto server endpoint root URL (eg. `"https://server/v1"`). Not that you *must* define a URL matching the version of the protocol the client supports, otherwise you'll get an error;
 - `headers`: The default headers to pass for every HTTP request performed to the Kinto server (eg. `{"Authorization": "Basic bWF0Og=="}`);
 - `retry`: Number of retries when the server fails to process the request (default: `1`)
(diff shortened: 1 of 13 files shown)


dstaley:
@leplatrem since this is a breaking change requiring a semver major bump, I'd really love it if you took a look! You don't need to review the actual code changes (unless you want to!); I mainly just want your sign-off on the new way to provide an adapter.

leplatrem:
I'm fine with the changes! 

I think I have to accept that I can't review thoroughly the changes in this PR :/ 

I let you manage the merging/releasing ;) I also opened https://github.com/Kinto/kinto.js/issues/1306 to officialize that

PR #1330, opened 2020-05-24
Remove Gecko build

### LINKED ISSUE #1306: Remove Gecko build
This library was used in Gecko (~ Firefox internals). In [one component](https://searchfox.org/mozilla-esr68/source/services/settings/RemoteSettingsClient.jsm), only a very tiny part was used, since we didn't need both-way sync. In [another component](https://searchfox.org/mozilla-esr68/source/toolkit/components/extensions/ExtensionStorageSync.jsm), we used both-way sync, but it is being modernized and migrated to another tech that matches the rest of the Firefox sync.

This means mainly two things:
* We don't need the Gecko dist anymore
* We won't provide much maintenance efforts from the Mozilla standpoint (ie. me). Note that it's been pretty stable and quiet in the last years.

@dstaley has been doing an amazing amount of work on the repo, and I would be happy to give him the full privileges for the maintenance and releasing if that allows to guarantee the continuity of this library :)

dstaley:
Per the discussion in #1306, this build removes the Gecko build. It also removes all the other related stuff such as build scripts and dependencies that were only used for the Gecko build.

Fixes #1306


dstaley:
> **Size Change:** -20 kB (114%) 🏆

Reduce your bundle size with this one weird trick!

PR #1629, opened 2021-10-28
Use the official demo URL

leplatrem:


dstaley:
@leplatrem Looks like the TLS certificate has expired. Would you mind checking on that?

leplatrem:
> @leplatrem Looks like the TLS certificate has expired. Would you mind checking on that?

Seems ok now, thanks!

Issue #1640, opened 2021-11-24
Upgrade deps to keep this and kinto-http.js in sync


matt-boris:
As suggested by @dstaley https://github.com/Kinto/kinto-http.js/pull/1183#issuecomment-966614295.

PR #1641, opened 2021-11-24
For #1640: Upgrade deps to match kinto-http.js

matt-boris:
For https://github.com/Kinto/kinto.js/issues/1640
Changes inspired by https://github.com/Kinto/kinto-http.js/pull/1183

## Changes

- Set node version to the [current active LTS (16)](https://nodejs.org/en/about/releases/)
- Removed Travis CI config
- Upgraded nearly every dev dependency in order to make it easier to keep package vulns at a minimum and this should make future package upgrades more straightforward, as there should be less breaking changes
    - `@rollup/plugin-typescript` proved to be the most difficult to fully upgrade, as there were many breaking changes
        - I took the suggestion from https://github.com/rollup/plugins/issues/287#issuecomment-611368317 to move `declaration` and `outDir` out of the `tsconfig.json` file
- Updated target from `es2017` to `es2021`


dstaley (changes requested):
@matt-boris Everything here looks great! Would you mind updating the body of the PR to contain some notes on what you've changed and why? I just want to make sure that we're recording the reasoning behind some of the less obvious changes (like using `npx` in package.json) so that we understand why those changes were made, and can also determine when they're no longer needed.

(Additionally please copy/paste that description into the original `kinto-http.js` PR so we have a record there.)

Once again, thanks for all these updates!

grahamalama:
Looks like Chrome/ Chromedriver version 96 has landed in `ubuntu-latest` and `windows-latest`, so you might be able to remove that from this PR and close https://github.com/Kinto/kinto.js/issues/1642.

Also, it looks like Node 16 is due to hit these runners [any day now](https://github.com/actions/virtual-environments/issues/4446), so depending on timing, you may be able to remove those steps from this PR as well. I opened https://github.com/Kinto/kinto-http.js/pull/1196 as a draft while I wait on that to happen.

dstaley (approved):
@matt-boris thank you!

leplatrem (approved):
LGTM!  Thanks for tackling this!

As Graham noted, maybe better to wait for rollout of new base images and align with  https://github.com/Kinto/kinto-http.js/pull/1196

matt-boris:
Will hold off on merging until https://github.com/actions/virtual-environments/issues/4446 is completed. This PR also addresses https://github.com/Kinto/kinto.js/issues/1642.

PR #1935, opened 2022-12-06
Update `ChromeUtils` module imports

grahamalama:
There've been several patches landed in Moz Central that have directly updated `kinto-http-client.js`. These changes all have to do with the way `ChromeUtils` modules are imported.

This commit makes it so our generated files contain these changes.

See also:
https://bugzilla.mozilla.org/show_bug.cgi?id=1777486
https://bugzilla.mozilla.org/show_bug.cgi?id=1792341


leplatrem (approved):
LGTM

PR #2050, opened 2023-06-14
Generate ES6/ES2015 Javascript for Gecko mozilla-central

leplatrem:
- [x] Get rid of `umd`
- [x] ~~Replace `__decorate()` polyfill by `Reflect`~~ ESlint will remain disabled, no need
- [x] ~~Figure out why we still missing a few object shorthands~~ same
- [x] Rename as `.sys.mjs` for linting to detect it as a module in mozilla-central
- [x] Run prettier with mozilla-central config to obtain a consistent output


leplatrem:
The file resulting of this PR was pushed here https://phabricator.services.mozilla.com/D181066

grahamalama (approved):

PR #2074, opened 2023-08-04
fix: explicitly set path to chrome driver

dstaley:


dstaley:
closing in favor of #2075

